PR #1, opened 2020-11-20
OCLC Classify Updates and Bugfixes

mwbenowitz:
This updates several components of the OCLC Classify process to improve its output and resolve several small bugs:

- Add a handler for unexpected status codes from the Classify service. The service will always return a `200` response, with custom response codes in the response XML which must be checked for. The current structure accounts for all documented responses but it seems that this is not guaranteed
- Improves parsing of authors/contributors to separate authors from contributors (who have specific roles assigned to them in the creation of a work). As these distinctions are made in a free text field this must be done with a regex
- Fix a bug where the process was repeatedly persisting ALL records to the database after the 10,000th record
- Improve the means by which records are marked as "FRBRized" (aka looked up in the Classify service). This simply marks all results as FRBRized once processed here, rather than looking at a specific outcome as this more accurately reflects that status.

### DIFF
--- a/dcdw/managers/oclcClassify.py
+++ b/dcdw/managers/oclcClassify.py
@@ -175,7 +175,7 @@ class ClassifyManager:
             
             return outRecords
         else:
-            print(responesXML)
+            print(responseXML)
             raise ClassifyError(message='Got unexpected response code')
     
     def checkTitle(self, oclcTitle):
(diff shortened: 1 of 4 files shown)


mwbenowitz:
Closing this to merge in new test suite and refactor, will reopen as a new PR

PR #79, opened 2021-04-15
Add Schedule for Tests.yaml

jcsmit17:
Adding scheduled run for tests.yaml.

### DIFF
--- a/.github/workflows/tests-regression.yaml
+++ b/.github/workflows/tests-regression.yaml
@@ -39,8 +39,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r dev-requirements.txt
           pip install -r requirements.txt
-          pip install allure-pytest
-      - name: Run test suite
+      - name: Run test suite with Allure
         run: |
           make allure-test
       - name: Get Allure history
--- a/.github/workflows/tests-regression.yaml
+++ b/.github/workflows/tests-regression.yaml
@@ -1,4 +1,4 @@
-name: ETL_Pipeline_Tests
+name: ETL Pipeline Tests (Regression)
 
 on: 
   push:
--- a/.github/workflows/tests-regression.yaml
+++ b/.github/workflows/tests-regression.yaml
@@ -3,8 +3,6 @@ name: ETL Pipeline Tests (Regression)
 on: 
   push:
     branches: [ main ]
-  pull_request:
-    branches: [ main ]
   schedule:
     - cron:  '0 0 * * 1'
 
--- a/.github/workflows/tests-regression.yaml
+++ b/.github/workflows/tests-regression.yaml
@@ -3,7 +3,7 @@ name: ETL Pipeline Tests (Regression)
 on: 
   push:
     branches: [ main ]
-  pull-request:
+  pull_request:
     branches: [ main ]
   schedule:
     - cron:  '0 0 * * 1'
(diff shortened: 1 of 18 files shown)


mwbenowitz:
@jcsmit17 I'm not familiar with Allure but I like the idea of it! if we generate the report via an action, how do we get that to some place it's reviewable? We could probably add some scripting to upload/serve them from an s3 bucket but is there a simpler way?

jcsmit17:
@mwbenowitz I am attempting to add the generated report to a GitHub page. I can add this in the about section of the drb-etl-pipeline so it is viewable.

jcsmit17:
@mwbenowitz To set this page up, I would need one of the following:

1. Gain admin access & set _gh-pages_ branch as main GitHub page for project.
2. Admin needs to set _gh-pages_ branch as main GitHub page for project.

This would be after the code review is approved.

jcsmit17:
@antlong @mwbenowitz Allure workflow is ready for review. Let me know if anything needs changed!

.github/workflows/tests-regression.yaml
```diff
@@ -0,0 +1,74 @@
...
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r dev-requirements.txt
+          pip install -r requirements.txt
+          pip install allure-pytest
```
mwbenowitz, line 40: Since you've added the option to run this through the Makefile (which, thanks!) can you add this to `dev-requirements.txt`? Just so anyone cloning this in the future doesn't end up wondering why they can't get it to work locally

.github/workflows/tests-unit.yaml
```diff
@@ -35,6 +35,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r dev-requirements.txt
           pip install -r requirements.txt
+          pip install allure-pytest
```
mwbenowitz, line 38: Same as above, would just like to add this to the dev requirements so it doesn't get forgotten about in the future.

CHANGELOG.md
```diff
@@ -9,14 +9,16 @@
 ## 2021-04-15 -- v0.5.6
 ### Added
 - Added `date_created` and `date_modified` indexes to the `records` table
-### Fixed
+- Scheduled tests-regression.yaml to run every Monday 12 AM EST
```
mwbenowitz, line 12: There's been a release since this was updated. Could you merge in the latest changes from `main` and make this part of the `v0.5.7` release?

mwbenowitz (commented):
I really like this, it will be a very useful tool for ensuring good coverage and testing practices. I just have a few housekeeping requests for the PR and then will be very happy to approve!

(P.S. I'm making you an admin on the repo so that you can configure Pages to display the reports)

.github/workflows/tests-regression.yaml
jcsmit17: I can do that. I also need to refactor my GitHub workflow to include the make command that runs allure-test. I'll do that in the next few commits.

CHANGELOG.md
jcsmit17: I can do that. Good callout!

mwbenowitz (approved):
🎉

PR #82, opened 2021-04-21
Remove Changelog Requirement for Regression Tests

jcsmit17:
Since tests run by allure-pytest will not have changes associated with each scheduled run, there is no need for Changelog requirement in Github workflow tests-regression.yaml.


mwbenowitz (approved):
Good catch, I didn't even think about that!

PR #150, opened 2022-01-19
Sfr 1405 oic webscraper

mitri-slory:


BeautifulSoupScript_edits.py
```diff
@@ -0,0 +1,213 @@
...
+            catContainer = soup.find(class_='catalog')
+
+            pubList = catContainer.find('tbody')
+
+            pubEntry = pubList.find_all('tr')
+
+            for pub in pubEntry:
+                print(url)
+
+                print(parse_Pub_Dict(pub))
```
mwbenowitz, line 62: This is a good separation of concerns! Maintaining these parsing methods helps keep this code easier to read/follow

BeautifulSoupScript_edits.py
```diff
@@ -0,0 +1,213 @@
...
+
+                metadata['url'] = downloadLink.get('href', None)
+                print(metadata)
+        '''
+        #Serializing download_dict to create a JSON file
+        with open("data_file.json", "w") as write_file:
+            json.dump(download_dict, write_file)
+        '''
+#Parsing the journal and miscelleanous publications for metadata at the bottom of webpage
+def parse_Pub_Jour(pub, journal_key):
```
mwbenowitz, line 95: minor thing: I'd pick either camelCase or snake_case for consistency (so either `parsePubJour` or `parse_pub_jour`) Python doesn't make a strong recommendation, but the DRB codebase uses camelCase so I'd ask that you follow that too! (happy to talk about this if you have a strong preference)

BeautifulSoupScript_edits.py
```diff
@@ -0,0 +1,213 @@
...
+    #Link should only be the link to the publication webpage
+    detailLink = pubLinks[0].get('href', None)
+
+    detailURL = '{}{}'.format(OIC_ROOT, detailLink)
+
+    print(detailURL)
+
+    try:
+        pubElem = fetchPageHTML(detailURL)
+    except ConnectionError or HTTPError:
```
mwbenowitz, line 106: 👍 I think it's really important to only explicitly handle exceptions we expect to get.

BeautifulSoupScript_edits.py
```diff
@@ -0,0 +1,213 @@
...
+
+    if pubContainer is None:
+        return None
+    #Metadata list is the last unordered list on the webpage
+    dataList = pubContainer.find_all('ul')[-1]
+
+    metadata = dataList.find_all('li')
+
+    #CHDS metadata dictionary
+    if journal_key == 2:
```
mwbenowitz, line 121: This is why I think using a `dict` with keys like `chds` above helps, this section would be more straightforward (both for us and people in the future!)

BeautifulSoupScript_edits.py
```diff
@@ -0,0 +1,213 @@
...
+                'extent': metadata[1]
+            }
+        elif len(metadata) == 4: #OIP 117 has metadata in middle of page instead of at the bottom
+            return None
+        else:
+            return {
+                'series': metadata[0].text,
+                'publicationInfo': metadata[1].text,
+                'isbnNumber(hardbook)': metadata[2].text,
+                'isbnNumber(ebook)': metadata[3].text,
```
mwbenowitz, line 157: I'd make this an array in the `isbnNumber` field for two reasons:

1. It keeps this object consistent with all of the other objects we're producing, which is important for data parsing/ingest
2. We actually treat those identifiers as essentially equivalent when it comes to processing, so it's easier down the road for us

BeautifulSoupScript_edits.py
```diff
@@ -0,0 +1,213 @@
...
+            'isbnNumber': metadata[1].text,
+            'extent': metadata[2].text
+            }
+    #Miscellaneous metadata dictionary
+    elif journal_key == 13:
+        return {
+            'title': metadata[0].text,
+            'publicationInfo': metadata[1].text,
+            'isbnNumber-13': metadata[3].text,
+            'isbnNumber-10': metadata[4].text,
```
mwbenowitz, line 175: Same comment as above, I'd merge these two things! (Thanks for the attention to detail in capturing this though!)

BeautifulSoupScript_edits.py
```diff
@@ -0,0 +1,213 @@
...
+            }
+
+#Parsing Dictionary catalogs
+def parse_Pub_Dict(pub):
+    pub_tds = pub.find_all('td')
+    metadata = {
+        'Volume': pub_tds[0].text,
+        'publicationInfo': pub_tds[2].text,
+        'isbnNumber': pub_tds[3].text,
+        'Extent': pub_tds[1].text
```
mwbenowitz, line 203: minor thing: `Extent` should be lower-case here to align with everything else

mwbenowitz (commented):
This looks really good, I left some comments on specific things that I think could be changed but overall this is really easy to follow and is well commented/structured.

I think the only element that is missing is adding the titles for each publication to the metadata block. Once that's been added I think we can call this complete!

BeautifulSoupScript_edits.py
```diff
@@ -0,0 +1,213 @@
...
+from requests.exceptions import ConnectionError, HTTPError
+
+
+def fetchPageHTML(url):
+    elemResponse = requests.get(url)
+    elemResponse.raise_for_status()
+    return elemResponse
+
+#Array of each catalog publications
+request_urls = [
```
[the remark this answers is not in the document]
mitri-slory: I agree. I just changed the request_urls array into a dictionary with the catalog acronyms as the keys. Makes the code and conditionals clearer.

BeautifulSoupScript_edits.py
mitri-slory: I don't have a preference, but I changed the method names to be camelCase to match the DRB codebase.

BeautifulSoupScript_edits.py
```diff
@@ -0,0 +1,213 @@
...
+    'https://oi.uchicago.edu/research/publications/lost-egypt',
+    'https://oi.uchicago.edu/research/publications/miscellaneous-publications'
+]
+
+
+OIC_ROOT = 'https://oi.uchicago.edu'
+
+def main(journal_key=0):
+    #Looping through each catalog to add the download links and metadata to a dictionary
+    for i, url in enumerate(request_urls):
```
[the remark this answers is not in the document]
mitri-slory: I actually changed my loop to what you suggested before reading this comment.

mwbenowitz (approved):
This looks good! I'm not going to merge it in because it's essentially a proof-of-concept for an OIC ingest process but this is very useful as a starting point for this.

I think we can consider SFR-1405 done now!

PR #151, opened 2022-01-31
Sfr 1415 999 data issue

mitri-slory:
Created new branch to resolve Edition Date Issues. I added a 9999date module to print out editions in the database for testing and created a publicationDateCheck method in the sfrRecord module to exclude edition dates in the future or in the past before 1488.


mitri-slory:
Making new branch because deleted 9999Date script

PR #152, opened 2022-01-31
Sfr 1415 9999 date issue

mitri-slory:
Created new branch to resolve Edition Date Issues. I created a publicationDateCheck method in the sfrRecord module to exclude edition dates in the future or in the past before 1488.


managers/sfrRecord.py
```diff
@@ -354,11 +354,20 @@ def saveEdition(self, edition):
             newEd.sub_title = edition['sub_title'].most_common(1)[0][0]
         newEd.alt_titles = [t[0] for t in edition['alt_titles'].most_common()]
 
-        # Set Publication Data
-        pubYearGroup = re.search(r'([0-9]{4})', str(edition['publication_date']))
-        if pubYearGroup:
-            newEd.publication_date = date(year=int(pubYearGroup.group(1)), month=1, day=1)
+        # Set Publication Date
+        def publicationDateCheck(edition):
+            #Exclude dates in the future and dates before the oldest book publisher was founded
```
mwbenowitz, line 359: 👍 Thanks for checking this out and adding this comment! Seems like the right approach to picking that date

managers/sfrRecord.py
```diff
@@ -354,11 +354,20 @@ def saveEdition(self, edition):
             newEd.sub_title = edition['sub_title'].most_common(1)[0][0]
         newEd.alt_titles = [t[0] for t in edition['alt_titles'].most_common()]
 
-        # Set Publication Data
-        pubYearGroup = re.search(r'([0-9]{4})', str(edition['publication_date']))
-        if pubYearGroup:
-            newEd.publication_date = date(year=int(pubYearGroup.group(1)), month=1, day=1)
+        # Set Publication Date
+        def publicationDateCheck(edition):
```
mwbenowitz, line 358: This method is a great start! A couple of requests:
- I think it makes more sense for it to be a method as part of the `SFRRecord` class rather than within this function. I think this is clearer and makes it more testable as well. (can make it a `@staticmethod` for simplicity).
- We should account for UTC time in the date check line, so I think `datetime.utcnow()` instead of `date.today()`. Will this ever make a difference? probably not, but we do use UTC time elsewhere and you never now what could happen

mwbenowitz (changes requested):
This is a great start, a simple improvement and adding the method gives us a place that we can start extending this functionality.

In addition to the comments I made in-line I have 2 more requests:

1. Please update the `CHANGELOG.md` following this guide https://github.com/olivierlacan/keep-a-changelog/blob/main/CHANGELOG.md. This helps us track what's in each release (this is why the PR is failing right now, there is a script that checks for this)
2. Please add a few tests for the new method in [this file](https://github.com/NYPL/drb-etl-pipeline/blob/10257aa9b0843b4606be9a8c944f076863bb5b36/tests/unit/test_sfrRecord_manager.py). They should just verify that we get the right year when we expect to and `None` when it falls outside of our range

tests/unit/test_sfrRecord_manager.py
```diff
@@ -249,6 +252,37 @@ def test_buildItems(self, testInstance, testDCDWRecord):
...
+        testPubDateCheck2 = SFRRecordManager.publicationDateCheck(testEdition2)
+        testPubDateCheck3 = SFRRecordManager.publicationDateCheck(testEdition3)
+        testPubDateCheck4 = SFRRecordManager.publicationDateCheck(testEdition4)
+        testPubDateCheck5 = SFRRecordManager.publicationDateCheck(testEdition5)
+
+        assert testPubDateCheck.year == 1900
+        assert testPubDateCheck2.year == date.today().year
+        assert testPubDateCheck3.year == 1488
+        assert testPubDateCheck4 == None
+        assert testPubDateCheck5 == None
```
mwbenowitz, line 284: These tests are great and cover everything I can think of (The tests for `.today()` and `.timedelta(days=1)` are especially useful I think)

I would just separate them out into 5 separate tests so that if one starts to fail we can have a good idea of the specific problem.

mwbenowitz (commented):
This looks great, please see my in-line comment for one recommended formatting change, but otherwise I think this done and ready to go.

mwbenowitz (approved):
🎉

PR #153, opened 2022-02-08
NOREF Clustering Stability and Performance Improvements

mwbenowitz:
This contains a range of improvements to the DRB ETL pipeline. Included in this are:

- Stability improvements for the clustering process by implementing an ES indexing solution that avoids the persistent `409 Conflict Error` issues that have been encountered. These also have the side effect of improving the speed of indexing by batching ES writes
- Performance improvements to the OCLC Classification process. This also contains a batch writing update, in this case to the RabbitMQ service
- Update/refresh of the development setup script
- Other improvements
- Testing updates

### DIFF
--- a/managers/sfrRecord.py
+++ b/managers/sfrRecord.py
@@ -26,6 +26,7 @@ class SFRRecordManager:
 
     def mergeRecords(self):
         dcdwUUIDs = set()
+
         for edition in self.work.editions:
             dcdwUUIDs.update(edition.dcdw_uuids)
 
--- a/managers/sfrRecord.py
+++ b/managers/sfrRecord.py
@@ -25,8 +25,6 @@ class SFRRecordManager:
         self.work = Work(uuid=uuid4(), editions=[])
 
     def mergeRecords(self):
-        existingIDs = {}
-
         dcdwUUIDs = set()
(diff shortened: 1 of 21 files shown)


managers/sfrRecord.py
```diff
@@ -442,15 +459,26 @@ def saveItem(self, item):
         return newItem
 
     @staticmethod
-    #Exclude dates in the future and dates before the oldest book publisher company was founded
+    # Exclude dates in the future and dates before the oldest book publisher company was founded
     def publicationDateCheck(edition):
```
mwbenowitz, line 463: @mitri-slory In testing this in QA I realized there are actually a few ways in which the `publication_date` can be set when it reaches this method. I've update the function to account for those, which are essentially differently formatted strings.

The core logic that you added is still the core of this, I simply extended the class by handling more cases up front. Would welcome any thoughts that you have on this!

mwbenowitz:
@mitri-slory This PR should bring the ETL pipeline into line with a system that you should be able to spin up locally. See the `Dependencies and Installation` section of the README, that should give you what you need to know to get this running locally. Would appreciate it if you gave it a try and let me know how it goes!

PR #154, opened 2022-02-08
Srf 852 citation plugin

mitri-slory:
Added initial citationPlugin Script and a drbCitation test file(still working on the tests). Also, added comments about writing code to have the initialization of the local DevelopmentSetUp file reset the database to its initial setup every time.


tests/unit/test_api_citation_blueprint.py
```diff
@@ -0,0 +1,66 @@
...
+            mockDB = mocker.MagicMock()
+            mockDBClient = mocker.patch('api.blueprints.drbCitations.DBClient')
+            mockDBClient.return_value = mockDB
+
+            mockDB.fetchSingleWork.return_value = 'dbWorkRecord'
+
+            mockUtils['formatResponseObject'].return_value\
+                = 'citationResponse'
+
+            with testApp.test_request_context('/'):
```
mwbenowitz, line 34: Right now these tests are failing because the don't have a value for the `formats=` param. You can fix that by giving the `test_request_context` method '/?formats=mla' (or whatever makes sense). Same for the second test below

tests/unit/test_api_citation_blueprint.py
```diff
@@ -0,0 +1,66 @@
+from flask import Flask
+import pytest
+
+from api.blueprints.drbCitation import citationFetch
+from api.utils import APIUtils
+
+
+class TestCitationBlueprint:
```
mwbenowitz, line 8: These tests look great (see comment below for why I think they're failing). I'd add one more to test that we get the expected response if `formats=` is not set.

mwbenowitz (changes requested):
A few minor changes but overall this looks really good! Should be in a good place to merge this in soon. We'll have to add some documentation to the Swagger docs file, but we can discuss that later (it's not hard but finicky and annoying).

mwbenowitz (approved):
🎉 looks good to me! one minor housekeeping thing and then should be good to go

PR #156, opened 2022-03-01
Created MLA Citation methods

mitri-slory:
For the edge cases when I test for translated works and works prepared by editors, the mlaGenerator method returns null instead of a string. I only tested for the edge case at lines 541-576 with the uuid "cdb43a57-44b8-4cc7-bcef-8dec1500137e" which only has one author, so I'm not sure if it's this specific work that's the problem or the code itself. I would appreciate feedback to resolve this issue for these edge cases.


api/blueprints/drbCitation.py
```diff
@@ -59,6 +60,777 @@ def citationFetch(uuid):
...
+
+def mlaBibleGenerator(citationWork):
+    #Bible editions before 1900
+        #Specific version info of Bible currently not displayed in database
+    if citationWork.editions[0].publication_date < date(1900, 1, 1):
+        citationResponse = f"{citationWork.title}. Unknown version, \
+                            {citationWork.editions[0].publication_place}, \
+                            {citationWork.editions[0].publication_date}."
+            
+        return ' '.join(citationResponse.split())
```
mwbenowitz, line 95: We can eliminate this `return` and just have the single one at the end of the method, which will make maintenance easier in the future.

(side note: 👍 on the split/join implementation here to eliminate any extra spaces, appreciate that!)

api/blueprints/drbCitation.py
```diff
@@ -59,6 +60,777 @@ def citationFetch(uuid):
...
+        #Specific version info of Bible currently not displayed in database
+    else:      
+        citationResponse = f"{citationWork.title}. Unknown version, \
+                            {citationWork.editions[0].publishers[0]['name']}, \
+                            {citationWork.editions[0].publication_date}."
+            
+        return ' '.join(citationResponse.split())
+
+def mlaGovGenerator(citationWork):
+    #Government Publications before 1900
```
mwbenowitz, line 107: Could you add a comment at the top of this describing why 1900 is the dividing line for these records? Very useful to know, would be good to know what's the difference! (looks like Publisher name vs place?)

api/blueprints/drbCitation.py
```diff
@@ -59,6 +60,777 @@ def citationFetch(uuid):
...
+    if citationWork.editions[0].publication_date < date(1900, 1, 1):
+
+        govName = citationWork.authors[0]['name']
+        govNameAndAgency = govName.replace('.', ',', 1)
```
mwbenowitz, line 111: Comment on this? Looks fine functionally, just curious why!

api/blueprints/drbCitation.py
```diff
@@ -59,6 +60,777 @@ def citationFetch(uuid):
...
+            return mlaNoAuthorGenerator(citationWork)
+
+        elif len(citationWork.authors) == 1:
+            return mlaOneAuthorGenerator(citationWork)
+
+        elif len(citationWork.authors) == 2:
+            return mlaTwoAuthorsGenerator(citationWork)
+
+        elif len(citationWork.authors) == 3:
+            return mlaThreeAuthorsGenerator(citationWork)
```
mwbenowitz, line 80: I think we can de-duplicate a lot of code in these methods! The logic as implemented looks good, but I think we can abstract it away into a smaller class so that you don't have to repeat yourself so much. This is generally a good practice because it makes maintenance easier and helps make the code more readable.

mwbenowitz (changes requested):
I'm requesting changes at the organizational level -- the low-level implementation of the MLA citation rules looks great, really appreciate the detail and thought that went into that.

What I'd like to do is go through the methods that provide specific formatting and abstract that logic into a shared class (or something else). Happy to discuss this, do it in a pair programming session, or anything else.

api/blueprints/drbCitation.py
```diff
@@ -59,6 +60,777 @@ def citationFetch(uuid):
...
+        if editorName:
+            listEditorName = editorName.split(', ')
+            editorLastName = listEditorName[0]
+            editorFirstName = listEditorName[1]
+            citationResponse = f"{authName}. {citationWork.title}, \
+                                edited by {editorFirstName} {editorLastName}, \
+                                {citationWork.editions[0].publishers[0]['name']}, \
+                                {citationWork.editions[0].publication_date}."
+
+            return ' '.join(citationResponse.split())
```
mwbenowitz, line 381: The issue mentioned in the PR comment stems from here, I think. There is no case for handling a book with no translator or editor, which results in a return of `None`.

I think this issue is one that will get handled with a refactor so that we can implement a base case for these conditions that can be shared across the formatting rules!

mitri-slory:
I'm not sure why the actual results from the test "def test_citationFetch_three_authors_work2 "keeps returning None when I run pytest. I would appreciate some feedback as to why that is.

api/blueprints/drbCitation.py
```diff
@@ -57,8 +58,710 @@ def citationFetch(uuid):
...
+                citationResponse = f"{authName}, et al. {citationWork.title}, \
+                                    edited by {editorFirstName} {editorLastName}, \
+                                    {citationWork.editions[0].publishers[0]['name']}, \
+                                    {citationWork.editions[0].publication_date.year}."
+
+            return ' '.join(citationResponse.split())
+        else:
+            mlaThreeAuthorsExtraRulesGenerator(citationWork)
+    else:
+        mlaThreeAuthorsExtraRulesGenerator(citationWork)
```
mwbenowitz, line 714: @mitri-slory The test is failing because these two method calls are missing their `return`s!

mwbenowitz (commented):
This is a solid implementation of this logic! I'd like to use this as a jumping off point for how we can make this more concise and maintainable. Let's talk about that tomorrow.

I'd say this is merge-able as is, but I'd like to discuss how we can make it better and more robust.

mwbenowitz (approved):
🎉 🎉 🎉

PR #159, opened 2022-04-01
Added script to generate updated muse manifests

mitri-slory:


SFR-1410_ManifestScript.py
```diff
@@ -0,0 +1,49 @@
...
+import json
+import logging
+
+from urllib.parse import urlparse, parse_qsl, urlencode, urlunparse
+
+logging.basicConfig(filename='manifest.log', encoding='utf-8', level=logging.INFO)
+
+s3_client = boto3.client("s3")
+
+listObjects = s3_client.list_objects_v2(Bucket='drb-files-qa', Prefix='manifests/muse/')
```
mwbenowitz, line 11: By default this will only load the first 1000 objects it finds from s3 (see the `MaxKeys` parameter in the boto3 documentation). So I think we need to add some paging capability here to load everything.

It might make sense to break this up into a few methods: 1 to load a batch and another to update all of the records in the batch.

For the behavior we can use to do this see the `IsTruncated` object in the `list_objects_v2` response documentation.

SFR-1410_ManifestScript.py
```diff
@@ -0,0 +1,49 @@
...
+contentListObjects = listObjects['Contents']
+
+#Iterating through all the Content objects to access the JSON muse objects
+for c in contentListObjects:
+    currKey = c['Key']
+    museObject = s3_client.get_object(Bucket='drb-files-qa', Key=f'{currKey}')
```
mwbenowitz, line 17: If this is something we are committing to the repo we should also have the `Bucket` name be configurable as a variable or something so that we can run this script on the production/qa buckets without having to change anything in the code

mwbenowitz (commented):
Have one comment to make sure we page through all of the objects in s3. The specific implementation of the update stuff looks great, I really like using `parse_qsl` which is something I should be more in the habit of!

Beyond that I'm wondering how we can best add this to the repo. In the past I've held off committing these kinds of scripts to the repo because I'm the only one who used them. Now I do think it makes sense but we should put them somewhere where we can organize them. Maybe we should put this in a `scripts/` directory? Open to ideas/suggestions.

Would also be great to have a little general commenting about what this does and what it should accomplish!

SFR-1410_ManifestScript.py
mitri-slory: I decided to use a Paginator to iterate through all the objects instead.

SFR-1410_ManifestScript.py
```diff
@@ -0,0 +1,68 @@
...
+        for c in batch['Contents']:
+            currKey = c['Key']
+            museObject = s3_client.get_object(Bucket= bucketName, Key= f'{currKey}')
+            update_batch(museObject, bucketName, currKey)
+
+# Loading batches of 1000 ProjectMuse JSON records using a paginator until there are no more batches
+@staticmethod
+def load_batch():
+    paginator = s3_client.get_paginator('list_objects_v2')
+    page_iterator = paginator.paginate(Bucket= bucketName, Prefix= 'manifests/muse/')
```
mwbenowitz, line 27: I didn't know about the paginator! This is really useful

SFR-1410_ManifestScript.py
```diff
@@ -0,0 +1,68 @@
+import boto3
+import json
+import logging
+
+from urllib.parse import urlparse, parse_qsl, urlencode, urlunparse
+
+logging.basicConfig(filename='manifest.log', encoding='utf-8', level=logging.INFO)
```
mwbenowitz, line 7: I didn't focus on this before but this is a good logging implementation. There is a shared DRB logging implementation, but we might want to update this to output to a (rotated) log file as you do here. A potential improvement for the future we should remember.

mwbenowitz (commented):
This looks great! I'm approving, before you merge it would be great if you could update the CHANGELOG because I just formally cut a release and this should be the first feature in the next `v0.10.1` version.

Also if this could be updated to move this to the new `scripts/` directory we discussed but isn't necessary.

mwbenowitz (approved):
Sorry meant for that last comment to be a formal approval!

PR #161, opened 2022-04-07
Modified constructWebpubManifest script

mitri-slory:


managers/muse.py
```diff
@@ -121,12 +121,24 @@ def constructWebpubManifest(self):
...
 
             if pdfItem:
-                pdfManifest.addChapter(
-                    '{}{}'.format(
-                        self.MUSE_ROOT, pdfItem.parent.get('href')
-                    ),
-                    titleItem.span.a.string,
-                )
+                #skipCover indicates whether to skip the intersitital cover page by adding a query parameter
+                skipCover = False
```
mwbenowitz, line 125: I think this needs to be set outside of the `for` loop otherwise it will always evaluate to `False` here.

managers/muse.py
```diff
@@ -121,12 +121,24 @@ def constructWebpubManifest(self):
...
+                if skipCover == False:
+                    pdfManifest.addChapter(
+                        '{}{}'.format(
+                            self.MUSE_ROOT, pdfItem.parent.get('href')
+                        ),
+                        titleItem.span.a.string,
+                    )
+                if skipCover == True:
+                    queryParam = '?start=2'
```
mwbenowitz, line 134: I think this could be simplified a bit by just using `skipCover` to set the `queryparam` and have only one call to `addChapter`. But that's a preference thing!

mwbenowitz (changes requested):
Structure looks good! just need to resolve what looks like a bug in setting the `skipCover` variable.

Also two other requests:
- Update the changelog
- Add/update at least one test to validate that this is working as anticipated

managers/muse.py
mitri-slory: Agreed. I moved it outside the for loop.

mwbenowitz (approved):
Looks great!

To resolve the merge conflict you should merge `main` into this branch and then move your updates to the new `v0.10.1` release there. That should resolve the conflict and allow all the checks to pass.

PR #162, opened 2022-04-21
Sfr 1451 nr instrumentation

mitri-slory:
When running the drbCitation script after running the APIProcess, the diagnostics shows up as a web transaction. For MuseProcess and HathiTrustProcess, I see both of them as non web transactions on the NR One account, however for CatalogProcess I don't see any metrics for it right now.


Dockerfile
```diff
@@ -2,6 +2,12 @@ FROM python:3.9
 ADD . /src
 WORKDIR /src
 
+RUN pip install --upgrade pip
+
+RUN pip install --no-cache-dir newrelic
+
+RUN pip install futures==3.1.1
```
mwbenowitz, line 9: This line is an artifact of some dependency issues that got resolved last sprint, we can get rid of it! (my bad here)

newrelic.ini
```diff
@@ -0,0 +1,219 @@
...
+# license_key = 
+
+# The application name. Set this to be the name of your
+# application as you would like it to show up in New Relic UI.
+# The UI will then auto-map instances of your application into a
+# entry on your home dashboard page. You can also specify multiple
+# app names to group your aggregated data. For further details,
+# please see:
+# https://docs.newrelic.com/docs/apm/agents/manage-apm-agents/app-naming/use-multiple-names-app/
+app_name = DRB ETL Pipeline
```
mwbenowitz, line 41: Leonard wrote up some best practices on NR configuration: https://docs.google.com/document/d/1qp3YqAnn_casSuMgg-gBbecoo6Kp71RzhBTs-tBPZy8 including naming which I think we should adopt. This can be done on a per-environment basis, so we should have a default name here and then override it in the env-specific section at the end of this file.

processes/hathiTrust.py
```diff
@@ -10,6 +10,14 @@
 from mappings.hathitrust import HathiMapping
 from logger import createLog
 
+import newrelic.agent
+
+if os.environ.get('NEW_RELIC_LICENSE_KEY', None):
```
mwbenowitz, line 15: I _think_ that this is necessary because the `newrelic.agent` needs to be told about it's configuration the first time we create a `background_task`.

The idea that I have is that if we instrument a background task in `main.py` (see my comment there). That we could delete this and instrument these methods _within_ that task and they would get captured.

main.py
```diff
@@ -1,6 +1,17 @@
...
+        config_file='newrelic.ini',
+        environment=os.environ.get('ENVIRONMENT', 'local')
+        )
+
 import argparse
 import inspect
-import os
 import yaml
 
 from logger import createLog
```
mwbenowitz, line 17: Related to my other comment, I think if we can instrument Line 44 below (`runProcess()`) with a background task we can capture everything. The tricky part of this is that we only want that to happen when the process is NOT `APIProcess`. I think we can figure that out though and hopefully that would simplify our instrumentation process.

mwbenowitz (commented):
In my testing this seems to work well! I left a comment about being able to initialize the agent once and then rely on that context. Not 100% sure if it will work but in my local testing it seemed to be successful? Just need to add a strategy for not running it in the `APIProcess` context

newrelic.ini
```diff
@@ -0,0 +1,224 @@
...
+#
+# The application environments. These are specific settings which
+# override the common environment settings. The settings related to a
+# specific environment will be used when the environment argument to the
+# newrelic.agent.initialize() function has been defined to be either
+# "development", "test", "staging" or "production".
+#
+
+[newrelic:local]
+monitor_mode = true
```
mwbenowitz, line 205: Lets set `monitor_mode` to `false` for everything but `qa` and `production` for now.

processes/developmentSetup.py
```diff
@@ -4,6 +4,7 @@
 import os
 from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
 import requests
+import newrelic.agent
```
mwbenowitz, line 7: I don't think we should instrument this process. It should only be run by people setting up test environments so it doesn't really tell us anything about how the app is running!

processes/migration.py
```diff
@@ -1,5 +1,6 @@
-from alembic import config
+import newrelic.agent
```
mwbenowitz, line 1: Would also skip this process! This is maintenance which also generally doesn't tell us much about the app. Anytime this gets run I expect you/I/another dev to be watching it and be able to report on what happens.

mwbenowitz (commented):
A few requests for tweaks to the instrumentation!

Also it looks like the installation line in the `Dockerfile` for the `newrelic` package is missing there. I think we can probably just add it to `requirements.txt` but it needs to be somewhere.

It can either go in requirements, or it needs to be added in three places so it's available to our build processes:
- `Dockerfile`
- `test_unit.yml`
- `.tugboat/config.yml`

PR #163, opened 2022-04-28
Added total_track_hits to ES parameter

mitri-slory:


mwbenowitz (commented):
Tested this in the temporary deployment and the issue is fixed! 👍

Just two requests before merging this:
1. Update a variable name for consistency
2. Fix the failing test

mwbenowitz (approved):
👍 👍 👍

PR #164, opened 2022-05-03
Added extractEdition script

mitri-slory:
For right now, I'm focusing on editions that have English as the language tied to them since some edition statements are in a different language. I'll attempt to find a way to translate those statements into English to extract the edition number. However, if we can't find a way to do that  we should talk to Sean to see how much of a priority it is to have the edition column filled for those foreign works.


scripts/extractEdition.py
```diff
@@ -0,0 +1,46 @@
...
+
+#Extracting the edition number from the edition_statement column of the DRB database to fill out the edition column
+def main():
+
+    dbManager = DBManager(
+        user='',
+        pswd='',
+        host='',
+        port='',
+        db=''
```
mwbenowitz, line 14: just an improvement suggestion for making scripts like this easier to run: set it up so that these are loaded from the CLI, or borrow the environment loading script from `main.py` and use environment variables

(that way you don't have to remove the secrets before committing)

scripts/extractEdition.py
```diff
@@ -0,0 +1,46 @@
...
+
+    dbManager.generateEngine()
+
+    dbManager.createSession()
+
+    for edit in dbManager.session.query(Edition) \
+        .filter(Edition.edit_statement != None and Edition.languages[0]['language'] == 'English'):
+            editStatement = edit.edition_statement
+            editDict = {'first': 1, 'second': 2, 'third': 3, 'fourth': 4, 'fifth': 5, \
+                        'sixth': 6, 'seventh': 7, 'eigth': 8, 'ninth': 9, 'tenth': 10}
```
mwbenowitz, line 25: re: question in the PR, we _could_ create mappings like this for common terms in French, German, Japanese, and other languages that commonly show up with edition statements. We could probably get the stats on these languages with some creative SQL and then write up the crosswalks.

(this is if we can't figure out an easy translation tool)

mwbenowitz (commented):
This looks like a good script! I think exploring how we can capture this data from other languages is a good next step.

We also need to apply this to newly processed records so that we can use it in our clustering process. There is a spot for it in the `Record` model: https://github.com/NYPL/drb-etl-pipeline/blob/c013fd6dff2282a0e2503bc35994927b4b500468/model/postgres/record.py#L35, we just need a way to run a script like this when ingesting new records. Once we have the core logic here I think configuring it to run in different places shouldn't be too hard.

scripts/extractEdition.py
```diff
@@ -0,0 +1,50 @@
+import os
+
+from model import Edition
+from main import loadEnvFile
+from managers import DBManager
```
mitri-slory, line 5: For some reason, when I run this script in the scripts folder, I get a ModuleNotFound Error for the model, main, and managers. It works fine when it's outside the scripts folder. Do you have a suggestion as to how to fix this and why it's happening?

mwbenowitz: I think if you leave it in the scripts file and run it from root (like `python main.py scripts/extractEdition.py`) that should work.

scripts/extractEdition.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+    dbManager.generateEngine()
+
+    dbManager.createSession()
+
+    for edit in dbManager.session.query(Edition) \
+        .filter(Edition.edition_statement != None and Edition.languages[0]['language'] == 'English'):
+            editStatement = edit.edition_statement
+            editDict = {'first': 1, 'second': 2, 'third': 3, 'fourth': 4, 'fifth': 5, \
```
mwbenowitz, line 28: This should probably use the `extract` method you used for the model! Just so we only have to maintain it in one place.

tests/unit/test_model_record.py
```diff
@@ -0,0 +1,61 @@
+import pytest
+
+
+class TestModelRecord:
+    @pytest.fixture
+    def testRecord(self, mocker):
+        mockHasVersion = mocker.MagicMock()
+        mockHasVersion.hex = 'first edition'
+        return mocker.MagicMock(
```
mwbenowitz, line 9: I don't think we need to use `MagicMock` here. We should just create a plain `Record` object and then run the tests on it. I _think_ this should trigger the getter/setter functionality.

Right now I don't think this is testing the actual model, it's only ever reaching mocked objects/methods. The tests themselves look good, so I think this is just about making sure they reach the actual code we need to test!

model/postgres/record.py
```diff
@@ -57,3 +59,16 @@ def __dir__(self):
...
+        return self.has_version
+
+    @hybrid_property
+    def getLanguage(self):
+        return self.langauges[0]['language']
+
+    @hasVersion.setter
+    def updateHasVersion(self, langugage):
+        editionNo = extract(self.has_version, langugage)
+        self.has_version = f'self.has_version|{editionNo}'
```
mwbenowitz, line 74: Missing brackets around `self.has_version` here

tests/unit/test_model_record.py
```diff
@@ -0,0 +1,61 @@
...
+    #Test for getter method of record model langauge
+    def test_language(self, testRecord):
+        testRecord.getLanguage.return_value = 'English'
+        testlanguage = testRecord.getLanguage()
+        assert testlanguage == 'English'
+
+    #First test for setter method of record model
+    def test_updateHasVersion(self, testRecord):
+        testRecord.updateHasVersion.return_value = 'first edition|1'
+        newVersion = testRecord.updateHasVersion()
```
mwbenowitz, line 51: We shouldn't have to call the setter method explicitly, we should be able to just do:
`testRecord.has_version = 'first edition'` and expect it to set it to `first edition|1`

mwbenowitz (changes requested):
The factored-out method for `extract` looks great! Just a few requested changes around testing and code organization.

1. I'm not sure where the `extract` function should live. I think the `scripts` folder is for maintenance scripts we run manually, so I'm not sure it should be there. Perhaps we should move it to a new `utilities` module in the `model` directory or elsewhere? Wherever you think makes sense and we would logically be able to find it
2. We should make sure the test suite is actually testing the new methods in the code. I don't think it's reaching the actual model class right now.

scripts/extractFutureEdition.py
```diff
@@ -0,0 +1,23 @@
+#Extracting the edition number from the has_version field of records for any future records
+def extract(version : str, language : str):
```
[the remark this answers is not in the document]
mitri-slory: I think for offhand scripts such as this one, it's a good idea since these are easier to forget over time unlike the other pieces of code we will interact with more often

model/postgres/record.py
```diff
@@ -57,3 +59,20 @@ def __dir__(self):
...
         for attr in dir(self):
             yield attr, getattr(self, attr)
+
+    @hybrid_property
+    def has_version(self):
+        return self._has_version
+
+    @hybrid_property
+    def getLanguage(self):
+        return self.languages[0]['language']
```
mwbenowitz, line 69: I think this is unused right now so we could probably eliminate it

scripts/extractFutureEdition.py
```diff
@@ -0,0 +1,23 @@
...
+    editStatement = version
+    editDict = {'first': 1, 'second': 2, 'third': 3, 'fourth': 4, 'fifth': 5, \
+                'sixth': 6, 'seventh': 7, 'eigth': 8, 'ninth': 9, 'tenth': 10}
+    editNumber = None
+
+    if language.lower() == 'english' or language.lower() == 'en' or language.lower() == 'eng':
```
mwbenowitz, line 8: 👍 to handling the full language and codes

mwbenowitz (commented):
Looks great! Only reason I'm not formally approving is that I think you said you'd be working on the foreign languages. So once that test is passing (or if we need more tests for other languages) should be good!

mwbenowitz (commented):
Just one comment really about simplifying some of the code in the `extract` method. Otherwise this looks good!

model/postgres/record.py
```diff
@@ -57,3 +59,20 @@ def __dir__(self):
...
         for attr in dir(self):
             yield attr, getattr(self, attr)
+
+    @hybrid_property
+    def has_version(self):
+        return self._has_version
+
+    @hybrid_property
+    def getLanguage(self):
+        return self.languages[0]
```
mwbenowitz, line 69: a minor nit, if we aren't using this method we should remove it!

mwbenowitz (approved):
Update looks really good, thanks!

model/postgres/record.py
mitri-slory: Sure thing.

PR #167, opened 2022-05-13
Sfr 1469 date in api response

mitri-slory:
I'm not sure if I should insert a space between the d and T, so it's clearer to see the date properly in the API Response or if it should stay this way since it needs to be there no space for the regression tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Getter and setter method for record model 
 - Script to extract edition number from edition statement for the edition field
 - Script to update future records with edition number in the has_version field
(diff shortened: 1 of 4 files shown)


mwbenowitz (commented):
👍 Looks great! Just needs a changelog entry and a fix to the failing test(s)

mwbenowitz:
@mitri-slory Also there shouldn't be a space in the output string. The format is the ISO 8601 standard for date/time: https://en.wikipedia.org/wiki/ISO_8601, so sticking with that means it should be parse-able by any application that consumes the API.

mwbenowitz (approved):
LGTM!

PR #168, opened 2022-05-18
HOTFIX Add has_version None handler

mwbenowitz:
This is a minor fix to an edge case where no edition statement (`has_version`) in the `Record` table model is present

### DIFF
--- a/model/postgres/record.py
+++ b/model/postgres/record.py
@@ -66,7 +66,9 @@ class Record(Base, Core):
 
     @has_version.setter
     def has_version(self, versionNum):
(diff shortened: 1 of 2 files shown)


mwbenowitz:
@mitri-slory if this looks good to you please feel free to merge at cut the release tomorrow morning! This is just a minor thing I noticed when checking out the code and wanted to fix before we hit production.

mitri-slory (approved):
This all looks good to me. That was a good conditional to add  and I appreciate the additional test added as well.

PR #169, opened 2022-05-19
Updated format dictionary in APIUtils class

mitri-slory:
The links I attached to the readable, downloadable, and requestable keys are just random for right now since I'm not sure which links should be associated with each key. I think we should ask Jackie and Crystal which media types and links should be considered readable, downloadable, and requestable.

### DIFF
--- a/api/utils.py
+++ b/api/utils.py
@@ -20,10 +20,9 @@ class APIUtils():
         'pdf': ['application/pdf'],
         'webpub_json': ['application/webpub+json'],
 
-        'readable': ['application/pdf', 'application/epub+zip', 'application/epub+xml'],
-        'downloadable': ['application/pdf', 'application/epub+zip', 'application/epub+xml', 'text/html'],
-        'requestable': ['application/pdf', 'application/epub+zip', 'application/epub+xml', 'text/html', \
-                        'application/html+edd', 'application/x.html+edd', 'application/webpub+json']
+        'readable': ['application/epub+xml', 'text/html', 'application/webpub+json'],
+        'downloadable': ['application/pdf', 'application/epub+zip'],
+        'requestable': ['application/html+edd', 'application/x.html+edd']
     }
 
     SOURCE_PRIORITY = {
--- a/api/utils.py
+++ b/api/utils.py
@@ -18,7 +18,12 @@ class APIUtils():
         'html': ['text/html'],
         'html_edd': ['application/html+edd', 'application/x.html+edd'],
         'pdf': ['application/pdf'],
-        'webpub_json': ['application/webpub+json']
+        'webpub_json': ['application/webpub+json'],
+
+        'readable': ['application/pdf', 'application/epub+zip', 'application/epub+xml'],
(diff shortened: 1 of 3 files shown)


api/utils.py
```diff
@@ -18,7 +18,12 @@ class APIUtils():
         'html': ['text/html'],
         'html_edd': ['application/html+edd', 'application/x.html+edd'],
         'pdf': ['application/pdf'],
-        'webpub_json': ['application/webpub+json']
+        'webpub_json': ['application/webpub+json'],
+
+        'readable': ['application/pdf', 'application/epub+zip', 'application/epub+xml'],
+        'downloadable': ['application/pdf', 'application/epub+zip', 'application/epub+xml', 'text/html'],
+        'requestable': ['application/pdf', 'application/epub+zip', 'application/epub+xml', 'text/html', \
+                        'application/html+edd', 'application/x.html+edd', 'application/webpub+json']
```
mwbenowitz, line 26: Should clarify a few things here about what role each of these media_types plays! And each format should only be present in one of these categories I think, so the front end can accurately serve a file that performs as the user expects.

- `application/epub+zip` is a downloadable format only. The `/epub+xml` is the un-zipped version of this and can only be read online
- `text/html` is a readable format only. This is how we embed 3rd-party readers/websites
- `application/html+edd` should be the only `requestable` things (Which is what we call NYPL's "Scan & Deliver" service: https://www.nypl.org/research/scan-and-deliver)
- `application/webpub+json` is _technically_ the only thing that is actually "readable" in the webreader, though I'm not sure it makes sense to only provide that.

mwbenowitz (changes requested):
Just need to update the mappings to the `media_types`. Really appreciate the thorough tests on this!

Also needs a quick addition to the changelog

api/utils.py
mitri-slory: Thanks for the clarification. Do you think we should also label application/pdf as the one exception to being in two formats: readable and downloadable format?

mitri-slory: Also, is `application/x.html+edd` also considered a requestable thing and what is the difference between that format and the format `application/html+edd.`

mwbenowitz: I'm actually a bit torn here. From the context of the DRB front-end a PDF is download-only. Another _hypothetical_ application might want to treat that differently, but will another such application ever exist? I can't decide what our model should be...

We could include it in `readable` for now and see if that causes any issues (which I don't think it will)

On your second question the `x.html+edd` version is a legacy thing that keeps cropping up as a bug because I can't seem to completely remove it from the database. So we _should_ be able to only have `html+edd` but I think we still need that other version for compatibility reasons.

mitri-slory: From what you've said, I'll keep PDF as a downloadable since it fits the context of the front end and I'll keep x.html+edd as a requestable alongside html+edd.

mwbenowitz (approved):
👍 just resolve that merge conflict then should be good to go!

PR #170, opened 2022-05-24
Added work title and author to edition endpoint

mitri-slory:


api/utils.py
```diff
@@ -199,6 +199,14 @@ def formatEditionOutput(
...
+    # Return a list of authors to formatEdition method
+    @staticmethod
+    def workAuthorInfo(edition):
+        workAuthors = []
+        for i in edition.work.authors:
+            workAuthors.append(i['name'])
+        return workAuthors
+
     @classmethod
     def formatEdition(
```
mwbenowitz, line 211: This looks really good, the only issue is that this will insert `work_title` and `work_authors` into all edition responses (search, work detail and edition detail), when I think we only want this to appear in the edition detail page.

mwbenowitz (changes requested):
This implementation is technically great, but I think it actually needs to be narrowed a bit so this only happens in edition detail requests.

I _think_ that could be done by adjusting the `APIUtils.formatEditionOutput` method, either by adding a new parameter to the call to `formatEdition` or making another change.

api/blueprints/drbEdition.py
```diff
@@ -21,12 +21,14 @@ def editionFetch(editionID):
...
-
+            
     if edition:
         statusCode = 200
         records = dbClient.fetchRecordsByUUID(edition.dcdw_uuids)
+        editionWorkTitle = edition.work.title
+        editionWorkAuthors = workAuthorInfo(edition)
         responseBody = APIUtils.formatEditionOutput(
-            edition, records=records, showAll=showAll, reader=readerVersion
+            edition, editionWorkTitle, editionWorkAuthors, records=records, showAll=showAll, reader=readerVersion
```
mwbenowitz, line 31: Should we move this into the `APIUtils` module next to `formatEditionOutput`? My gut tells me yes so we can keep the blueprint relatively simple, but not sure if that is moving the functionality to place where its less visible.

mitri-slory: I'll follow your suggestion and once I commit it, we can see how it ultimately looks.

mwbenowitz (approved):
👍 Looks great!

PR #172, opened 2022-05-31
Created new parser for inTechOpen website

mitri-slory:
I'm not sure how to most efficiently access the inTechOpen book identifier number, so I can add it to the HTML url link. Do you have any suggestions how to do so and how to test to make sure those works have the updated links?


managers/parsers/inTechOpenParser.py
```diff
@@ -0,0 +1,60 @@
+from managers.parsers.abstractParser import AbstractParser
+from managers.webpubManifest import WebpubManifest
+
+
+class inTechOpenParser(AbstractParser):
```
mwbenowitz, line 5: nit: Class names should be capitalized!

managers/parsers/inTechOpenParser.py
```diff
@@ -0,0 +1,60 @@
...
+        self.source = self.record.source
+        self.identifier = list(self.record.identifiers[0].split('|'))[0]
+
+    def validateURI(self):
+        return super().validateURI()
+
+    def createLinks(self):
+        s3Root = self.generateS3Root()
+
+        if self.mediaType == 'application/pdf':
```
mwbenowitz, line 20: re: your question. The identifier that the website uses seems to only be consistently available as part of the URL we use to construct a manifest and download the book from the `application/pdf` link.

Right now it seems like the best solution would be to extract that identifier from here using a regex and then make sure that it is in the `text/html` URL later on.

The only issue is that this relies on those links being sent in the right order... perhaps that is something we could improve in the `DOABLinkManager` class?

managers/parsers/inTechOpenParser.py
```diff
@@ -0,0 +1,60 @@
...
+
+        elif self.mediaType == 'text/html':
+            htmlPath = 'text/{}/{}.html'.format(self.source, self.identifier)
+            htmlURI = '{}{}'.format(s3Root, htmlPath)
+
+            return [
+                (htmlURI, {'reader': False}, 'text/html', None, None)
+            ]
+
+        return super().createLinks()
```
mwbenowitz, line 54: This might be WIP artifact, but just a heads up: the way this is implemented in the base/abstract class, this will clobber any changes that you make to the method above and just use the default, inherited, implementation

mwbenowitz (commented):
Left a comment about approach, but I'm not 100% sure it's the final answer.

As for testing we can run the DOAB ingest process for a single record. So to test for the example in the ticket, you can run the process like this:
`python main.py -p DOABProcess -i single -r 20.500.12854/66288`

managers/parsers/inTechOpenParser.py
```diff
@@ -23,35 +23,39 @@ def createLinks(self):
...
+            # QUESTIONS
+            # 1. Do all Intechopen books have PDF links?
+            # 2. How do we handle potential errors/add a fallback condition
+            # 3. Are PDF links formatted consistently?
+
             return [
                 (manifestURI, {'reader': True}, 'application/webpub+json', (manifestPath, manifestJSON), None),
-                (self.uri, {'reader': False, 'download': True}, self.mediaType, None, None)
+                (self.uri, {'reader': False, 'download': True}, self.mediaType, None, None),
+                (htmlURI, {'reader': False}, 'text/html', self.mediaType, None, None)
```
mwbenowitz, line 41: `self.mediaType` is duplicative of the explicit media type set here so we should delete it!

managers/parsers/inTechOpenParser.py
```diff
@@ -23,35 +23,39 @@ def createLinks(self):
...
-            ]
 
-        elif self.mediaType == 'text/html':
-            htmlPath = 'text/{}/{}.html'.format(self.source, self.identifier)
-            htmlURI = '{}{}'.format(s3Root, htmlPath)
+    def createHTMLURI(self):
+        '''Using regular expressions to search for identifier in PDF and parse it into HTML URI'''
+        
+        identRegex = r'intechopen.com\/storage\/books\/([\d]+)'
+        match = re.search(identRegex, self.uri)
```
mwbenowitz, line 53: Not sure if you looked up the difference between `match` and `search` but `search` is 100% the right method to use here 👍

managers/parsers/inTechOpenParser.py
```diff
@@ -23,35 +23,39 @@ def createLinks(self):
...
+        identifier = match.group(1)
```
mwbenowitz, line 54: This might raise an error if the regex search fails. We should have a handler for that, and then also in the method above make sure we don't add a link object if  `htmlURI` is `None`

mwbenowitz (changes requested):
This is looking really good! The `createLinks` method is just about there, just need to add some error handling around the regex.

We also need to implement a general check in the `validateURI` property. The `super()` method always returns `True` so that will break the intended flow of these parsers. It should do some kind of check (either substring or regex) to make sure that the URI we have is from intechopen.

managers/parsers/inTechOpenParser.py
```diff
@@ -35,25 +41,34 @@ def createLinks(self):
...
         match = re.search(identRegex, self.uri)
-        identifier = match.group(1) 
 
-        htmlURI = f'www.intechopen.com/books/{identifier}'
+        if match != None:
+            identifier = match.group(1) 
+
+            htmlURI = f'www.intechopen.com/books/{identifier}'
 
         return htmlURI
```
mwbenowitz, line 73: Possible that this will be undefined here if `match` is `None`!

mwbenowitz (changes requested):
`validateURI` looks great!

For the PDF link parsing I think we should be handling that array of link tuples a bit differently. Mainly want to be sure that we give patrons all of the valid options when we have them!

mwbenowitz (approved):
Looks great! We might want to remove the `TODO` comments, but other than that let's get it out in the next prod release!

PR #173, opened 2022-06-01
Updated script for toc object in Muse manifests

mitri-slory:


scripts/Interstitial_Pages_SFR-1410.py
```diff
@@ -54,9 +54,33 @@ def update_batch(museObject, bucketName, currKey):
...
+        query = dict(parse_qsl(url_parts[4]))
+        logging.info(query)
+
+        # Updating query parameter of url
+        params = {'start': 2}
+        query.update(params)
+        url_parts[4] = urlencode(query)
+
+        # Updating old url link with new link
+        museDictObject['toc'][r]['href'] = urlunparse(url_parts)
```
mwbenowitz, line 79: A general practice comment: this could all be refactored into a separate method that contains this logic for `toc` and `readingOrder` because it should be the same. This would make it a bit more maintainable and give us confidence that the same thing is happening in both arrays.

Obviously this works! This is just with an eye towards best practice.

mwbenowitz (approved):
Looks great, just see my comment about best practice for DRY!

scripts/Interstitial_Pages_SFR-1410.py
mitri-slory: Thanks for the advice. I'll keep that in mind for future scripts I write.

PR #175, opened 2022-06-03
NOREF Fix DOAB single record import

mwbenowitz:
The method to import individual DOAB records (used for testing and dev purposes) had broken due to a change in DOAB's OAI-PMH feed. This updates the method to accurately extract the correct XML object that can be understood by the DOAB mapping.


mitri-slory (approved):
The change looks good and the only change I suggest is to fix the conflicting Changelog files before merging this branch into main. Actually, you just fixed it while I wrote this review. Good job on getting on that.

PR #176, opened 2022-06-09
Made new script to update current intechopen links

mitri-slory:


scripts/updateOldInTechOpenLinks.py
```diff
@@ -0,0 +1,46 @@
...
+
+    doabProcess = DOABProcess('single', None, None, None)
+
+    identRegex = r'intechopen.com\/storage\/books\/([\d]+)'
+
+    for record in dbManager.session.query(Record) \
+        .filter(Record.source != 'doab') \
+        .filter(Record.publisher == '{IntechOpen||,IntechOpen||}').all():
+            
+            match = re.search(identRegex, record.has_part)
```
mwbenowitz, line 37: `record.has_part` comes back as an array I think, so you might need to iterate through this and do the check on each value. The regex test looks good though!

mwbenowitz (commented):
Looks great, just want to make sure we're handling the `record.has_part` field correctly. If we are then happy to approve this!

mwbenowitz (approved):
Looks great!

scripts/updateOldInTechOpenLinks.py
```diff
@@ -42,8 +41,10 @@ def main():
 
             if match == None:
                 doabProcess.importSingleOAIRecord(record.source_id)
-                
-            
+                doabProcess.saveRecords()
+                #Make session record a set after updating record in database to reduce memory space
+                record = set()
```
mwbenowitz, line 46: should be `doabProcess.records = set()` here! That will make sure that the process class doesn't build up a huge object containing all the records we've processed.

mitri-slory: Appreciate the correction!

mwbenowitz (approved):
👍

PR #178, opened 2022-06-14
Sfr 921v2 work author object

mitri-slory:
I decided to expose the entire author object instead of just the name and viaf properties.


mwbenowitz (approved):
I think exposing the whole object is the right option! They're pretty small so it shouldn't have any performance impact and it allows any application that consumes our API to choose what to do with those fields.

PR #179, opened 2022-06-21
Sfr 1497 catalogs in link table

mitri-slory:


scripts/updateRecordsCatalogLink.py
```diff
@@ -0,0 +1,43 @@
...
+    catalogRegex = r'catalog: true'
+    linkRegex = r'text/html'
+
+    for record in dbManager.session.query(Record) \
+        .filter(Record.source == 'nypl').all():
+            
+            for i in record.has_part:
+                if re.search(catalogRegex, i) != None:
+                    if re.search(linkRegex, i) != None:
+                        i.replace('text/html', 'application/html/catalog')
```
mwbenowitz, line 37: I'm not 100% sure this will persist the change in the database. We might need to update the full `has_part` array, should try to evaluate that in QA with a limited test before merging this

mwbenowitz (commented):
Scripts look good, just a few minor things!

scripts/updateLinksCatalog.py
```diff
@@ -0,0 +1,40 @@
...
+
+    dbManager.createSession()
+
+    for link in dbManager.session.query(Link) \
+        .filter(Link.media_type == 'text/html').all():
+            
+            if link.flags != None and len(link.flags) > 0:
+                for i in link.flags:
+                    if i['catalog'] == True:
+                        link.media_type = 'application/html/catalog'
```
mwbenowitz, line 34: This totally works, but it might be more efficient to do an `update` query as described here: https://docs.sqlalchemy.org/en/14/core/dml.html#sqlalchemy.sql.expression.update

Alternatively, should be able to add the catalog `True` check as a filter, which will simplify this a bit.

mwbenowitz:
@mitri-slory This looks like it includes the changes for SFR-1495 and SFR-1496, is that correct? If so I think this is small enough to review and test all of these changes here and once that's done we can close the other PRs

scripts/updateRecordsCatalogLink.py
```diff
@@ -0,0 +1,43 @@
...
+    catalogRegex = r'catalog\": true' 
+    linkRegex = r'text\/html'
+
+    for record in dbManager.session.query(Record) \
+        .filter(Record.source == 'nypl').all():
+            
+            for i, elem in enumerate(record.has_part):
+                if re.search(linkRegex, elem) != None:
+                    if re.search(catalogRegex, elem) != None:
+                        record.has_part = [record.has_part[i].replace('text/html', 'application/html+catalog')]
```
mwbenowitz, line 37: This will have a side effect of resetting the `has_part` array to a single value, deleting any other values originally in the array!

I think the best way to deal with this is to create a local array that all of the `has_part` strings get loaded into, and then set the column value to that after the loop is executed.

scripts/updateLinksCatalog.py
```diff
@@ -0,0 +1,37 @@
...
+
+    dbManager.generateEngine()
+
+    dbManager.createSession()
+
+    for link in dbManager.session.query(Link) \
+        .filter(Link.media_type == 'text/html') \
+        .filter(Link.flags != {}) \
+        .filter(Link.flags['catalog'] == 'true').all():
+            link.media_type = 'application/html+catalog'
```
mwbenowitz, line 31: 👍

scripts/updateRecordsCatalogLink.py
mitri-slory: I believe putting record.has_part[i].replace('text/html', 'application/html+catalog') into an array fixed this issue since when I tested out my script on the Record with the id 4924, the has_part field updated with the new catalog media type.

`{"1|https://www.nypl.org/research/collections/shared-collection-catalog/bib/b22313364|nypl|application/html+catalog|{\"catalog\": true, \"download\": false, \"reader\": false}"}`

The has_part field for that record was originally this:

`{"1|https://www.nypl.org/research/collections/shared-collection-catalog/bib/b22313364|nypl|text/html|{\"catalog\": true, \"download\": false, \"reader\": false}"}`

scripts/updateRecordsCatalogLink.py
```diff
@@ -0,0 +1,52 @@
...
+    dbManager.generateEngine()
+
+    dbManager.createSession()
+
+    catalogRegex = r'catalog\": true' 
+    linkRegex = r'text\/html'
+
+    for record in dbManager.session.query(Record) \
+        .filter(Record.source == 'nypl') \
+        .filter(Record.id == 4925).all():
```
mwbenowitz, line 33: `filter` on specific id leftover from development here!

scripts/updateRecordsCatalogLink.py
```diff
@@ -0,0 +1,52 @@
...
+
+            recordArray = []
+            
+            for i, elem in enumerate(record.has_part):
+                if re.search(linkRegex, elem) != None:
+                    if re.search(catalogRegex, elem) != None:
+                        recordArray.append(record.has_part[i].replace('text/html', 'application/html+catalog'))
+                    else:
+                        recordArray.append(record.has_part[i])
```
mwbenowitz, line 42: technically don't need this, but that's just a nit!

mwbenowitz (approved):
Looks good! Just need to take out one bit leftover from development then this is good to merge!

PR #183, opened 2022-06-29
Sfr 1498 update es catalog links

mwbenowitz:
This script finds all work/edition records that have a resolvable link pointing to an NYPL resource. If that resource is a link to a catalog record (not to an EDD option), the `mediaType` of that link is updated to `application/html+catalog` which prevents these links from being picked up as readable.

### DIFF
--- a/scripts/updateESEditionFormats.py
+++ b/scripts/updateESEditionFormats.py
@@ -1,0 +1,87 @@
+import os
+import re
+
+from model import Work, Item, Edition, ESWork
+from main import loadEnvFile
+from managers import DBManager, ElasticsearchManager
+
+
+def main():
+    '''
+    Updating NYPL catalog records with new link to replace text/html link
+    '''
+
+    loadEnvFile('local-qa', fileString='config/{}.yaml')
+
+    dbManager = DBManager(
+        user=os.environ.get('POSTGRES_USER', None),
+        pswd=os.environ.get('POSTGRES_PSWD', None),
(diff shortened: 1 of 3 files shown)


scripts/updateESEditionFormats.py
```diff
@@ -0,0 +1,87 @@
...
+            else:
+                print('SKIP')
+        
+        if saveWork:
+            workRec.save()
+
+    dbManager.closeConnection()
+
+
+def getEditionFormats(items):
```
mitri-slory, line 59: I suggest adding a docstring to explain this method

mitri-slory, line 59: I suggest adding a docstring to explain this method

scripts/updateESEditionFormats.py
```diff
@@ -0,0 +1,87 @@
...
+            formats.add('application/html+catalog')
+
+        for link in links:
+            if link[0] not in ['text/html', 'application/html+catalog']:
+                formats.add(link[0])
+
+    return list(formats) if len(formats) > 0 else None
+
+
+def linkCriteriaChecker(links, flag, value):
```
mitri-slory, line 82: Same as my previous comment about adding a docstring to this method.

mitri-slory (commented):
This script looks really good and looks to cover all the possible items with links besides text/html as well which is a good catch. I just have suggestions to add docstrings to the two methods you implemented to explain them. Otherwise, the script is good to go.

PR #184, opened 2022-07-06
SFR-1498 Update script to be more robust

mwbenowitz:
This adds some error handling and edge-case guarding to make the script more robust. Specifically this:
1. Handles an ES `NotFoundError` in the case a work cannot be located
2. Skips any incorrectly identified editions (generally occurs in the case of duplicate works).


mitri-slory (approved):
The try and except blocks with the additional conditional look good for catching these duplicate works and non-indexed works.

PR #185, opened 2022-07-11
Added two filters for govDoc and noGovDoc

mitri-slory:
To add the government document filters to the backend, I decided to first add two new QUERY_TERMS to represent the "onlyGovDoc" and "noGovDoc" filters. Then, I'm currently attempting to update the "createFilterClausesAndAggregations()" method in the Elastic API to add these two filters, but I'm struggling to understand how to format my queries for these filters especially since the measurements field of edition is a jsonb object. This object has two terms I need to search for which are the type and value that should be "government_document" as the type and "0" or "1" as the value. The two conditionals I made in this method are commented out for right now since I'm still thinking how the queries should look for both filters.

### DIFF
--- a/api/elastic.py
+++ b/api/elastic.py
@@ -343,13 +343,23 @@ class ElasticClient():
         languageFilters = list(filter(
             lambda x: x[0] == 'language', filterParams)
         )
-        formatFilters = list(filter(lambda x: x[0] == 'format', filterParams))
+        formatFilters = list(filter(
+            lambda x: x[0] == 'format', filterParams)
+        )
         displayFilters = list(filter(
             lambda x: x[0] == 'showAll', filterParams)
+        )
+        govDocFilters = list(filter(
+            lambda x: x[0] == 'govDoc', filterParams)
+        )
+        noGovDocFilters = list(filter(
+            lambda x: x[0] == 'noGovDoc', filterParams)
         )
 
         dateFilter, dateAggregation = (None, None)
         formatFilter, formatAggregation = (None, None)
+        govDocFilter, govDocAggregation = (None, None)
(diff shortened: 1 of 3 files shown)


api/elastic.py
```diff
@@ -343,13 +343,23 @@ def createFilterClausesAndAggregations(self, filterParams):
...
+        )
         displayFilters = list(filter(
             lambda x: x[0] == 'showAll', filterParams)
         )
+        govDocFilters = list(filter(
+            lambda x: x[0] == 'govDoc', filterParams)
+        )
+        noGovDocFilters = list(filter(
+            lambda x: x[0] == 'noGovDoc', filterParams)
+        )
```
mwbenowitz, line 357: I would make this a single `govDoc` filter and have that have 3 valid values:
1. `onlyGov` returns only government documents
2. `nonGov` Returns only NON government documents
3. `all` Returns everything regardless (the default if this filter isn't set, but still need an explicit value IMO)

api/elastic.py
```diff
@@ -386,6 +396,10 @@ def createFilterClausesAndAggregations(self, filterParams):
             formatAggregation = A(
                 'filter', **{'terms': {'editions.formats': formats}}
             )
+        
+        #if len(govDocFilters) > 0:
+
+        #if len(noGovDocFilters) > 0:
```
mwbenowitz, line 402: There is a field in the ES index `is_government_document` but at present it does not appear to be populating correctly. I think this _would_ be the easiest way to implement the feature if we can fix it!

mwbenowitz (commented):
We might need to implement a bug fix to how some metadata is generated, should discuss the details of this, but off to a good start here! I have a better idea of how we can implement this now

PR #186, opened 2022-07-12
Sfr 1506 gov doc in es

mitri-slory:
I updated the sfrElasticRecord module to utilize the measurements field of editions instead of works when calling the addGovDocStatus() method and made a new script to update old government document works in ES with a True boolean value in their is_government_document fields.

### DIFF
--- a/managers/sfrElasticRecord.py
+++ b/managers/sfrElasticRecord.py
@@ -89,13 +89,10 @@ class SFRElasticRecordManager:
     
     @staticmethod
     def addGovDocStatus(measurements):
-        govDocMeasurement = list(filter(
-            lambda x: x.quantity == 'government_document', measurements
-        ))
-        if len(govDocMeasurement) > 0:
-            if int(govDocMeasurement[0].value) == 1:
-                return True
-        
+        for dict in measurements:
+            if dict['type'] == "government_document":
+                if dict['value'] == "1":
+                    return True
         return False
 
     def createEdition(self, edition):
--- a/managers/sfrElasticRecord.py
+++ b/managers/sfrElasticRecord.py
@@ -76,7 +76,7 @@ class SFRElasticRecordManager:
         ]
         
         self.work.is_government_document = SFRElasticRecordManager.addGovDocStatus(
-            self.dbWork.measurements
+            self.work.editions.measurements
         )
 
         self.work.editions = [self.createEdition(e) for e in self.dbWork.editions]
--- a/managers/sfrElasticRecord.py
+++ b/managers/sfrElasticRecord.py
@@ -76,7 +76,7 @@ class SFRElasticRecordManager:
         ]
         
         self.work.is_government_document = SFRElasticRecordManager.addGovDocStatus(
-            self.work.editions.measurements
+            self.work.editions
         )
 
         self.work.editions = [self.createEdition(e) for e in self.dbWork.editions]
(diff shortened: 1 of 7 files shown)


scripts/govDocES.py
```diff
@@ -0,0 +1,51 @@
...
+    dbManager.generateEngine()
+
+    dbManager.createSession()
+
+    for work in dbManager.session.query(Work) \
+        .join(Edition) \
+        .filter(Edition.measurements != None) \
+        .filter(Edition.measurements != []) \
+        .filter(Edition.measurements != [{}]).all():
+            for dict in work.edit.measurements:
```
mwbenowitz, line 33: needs to be `work.edition.measurements` I think

scripts/govDocES.py
```diff
@@ -0,0 +1,51 @@
...
+                            workRec = ESWork.get(work.uuid, index=esManager.index)
+                            workRec.is_government_document = True
+                            workRec.save()
+                        except NotFoundError:
+                            print('Work not indexed, skipping')
+                            continue
+                    else:
+                        continue
+                else:
+                    continue
```
mwbenowitz, line 46: a nit, but can probably omit these, unless you like the explicitness of having them

scripts/govDocES.py
```diff
@@ -0,0 +1,51 @@
...
+        .filter(Edition.measurements != None) \
+        .filter(Edition.measurements != []) \
+        .filter(Edition.measurements != [{}]).all():
+            for dict in work.edit.measurements:
+                if dict['type'] == "government_document":
+                    if dict['value'] == "1":
+                        try:
+                            workRec = ESWork.get(work.uuid, index=esManager.index)
+                            workRec.is_government_document = True
+                            workRec.save()
```
mwbenowitz, line 39: for speed/performance I think you can add a `break` here so you're not saving the same work multiple times in case it has multiple editions with that measurement.

managers/sfrElasticRecord.py
```diff
@@ -76,7 +76,7 @@ def enhanceWork(self):
         ]
         
         self.work.is_government_document = SFRElasticRecordManager.addGovDocStatus(
-            self.dbWork.measurements
+            self.work.editions.measurements
```
mwbenowitz, line 79: `self.work.editions` is an array, so this has to become a bit more complicated unfortunately, either here or in the `addGovDocStatus` method.

mitri-slory: Ok. I decided to change the addGovDocStatus method.

mwbenowitz: sorry if I was unclear! `self.work.editions` is an array of `edition` objects, each of which has a `measurements` array, so that's the complexity that the method has to account for. Should be very similar logic-wise to the script that you wrote.

Once this is fixed up the PR should be good to go!

scripts/govDocES.py
```diff
@@ -0,0 +1,54 @@
...
+                                workRec = ESWork.get(work.uuid, index=esManager.index)
+                                workRec.is_government_document = True
+                                workRec.save()
+                                break_out_flag = True
+                                break
+                            except NotFoundError:
+                                print('Work not indexed, skipping')
+                                continue
+                if break_out_flag:
+                    break
```
mwbenowitz, line 49: 👍

mwbenowitz (approved):
Looks great! Let's run this in QA but I think this should un-block us on the filter ticket.

PR #187, opened 2022-07-22
WIP SFR-1510 Make query for edition measurements more effecient

mwbenowitz:
This improves the query in two ways:
* Eliminates unecessary data returns by limiting the query to returning uuid and measurement fields
* Reduces number of rows returned by aggregating all edition measurements with a single uuid for a work

### DIFF
--- a/model/postgres/edition.py
+++ b/model/postgres/edition.py
@@ -48,7 +48,10 @@ class Edition(Base, Core):
     links = relationship('Link', secondary=EDITION_LINKS, backref='editions', cascade='all, delete')
     rights = relationship('Rights', secondary=EDITION_RIGHTS, backref='editions', cascade='all, delete')
 
-    __tableargs__ = (Index('ix_editions_dcdw_uuids', dcdw_uuids, postgresql_using="gin"))
+    __tableargs__ = (
+        Index('ix_editions_dcdw_uuids', dcdw_uuids, postgresql_using="gin"),
+        Index('ix_editions_measurements', measurements, postgresql_using='gin')
+    )
 
     def __repr__(self):
         return '<Edition(place={}, date={}, publisher={})>'.format(
--- a/scripts/govDocES.py
+++ b/scripts/govDocES.py
@@ -36,14 +36,15 @@ def main():
         .filter(Edition.measurements != [{}]) \
         .group_by(Work.uuid) \
         .yield_per(batchSize):
+            uuid, editionMeasurementArray = work
 
             break_out_flag = False
-            for editionMeasurements in work[0]:
+            for editionMeasurements in editionMeasurementArray:
                 for measurement in editionMeasurements:
                     if measurement['type'] == "government_document":
                         if measurement['value'] == "1":
                             try:
-                                workRec = ESWork.get(work[0], index=esManager.index)
+                                workRec = ESWork.get(uuid, index=esManager.index)
                                 workRec.is_government_document = True
                                 workRec.save()
                                 break_out_flag = True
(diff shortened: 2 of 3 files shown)


scripts/govDocES.py
```diff
@@ -27,20 +28,22 @@ def main():
...
+        .filter(Edition.measurements != [{}]) \
+        .group_by(Work.uuid) \
+        .yield_per(batchSize):
 
             break_out_flag = False
-            
-            for edition in work.editions:
-                for measurement in edition.measurements:
+            for editionMeasurements in work[0]:
+                for measurement in editionMeasurements:
```
mitri-slory, line 42: Why does work[0] equal to the editionMeasurements and not the uuid or any other attribute?

scripts/govDocES.py
```diff
@@ -27,20 +28,22 @@ def main():
...
                     if measurement['type'] == "government_document":
                         if measurement['value'] == "1":
                             try:
-                                workRec = ESWork.get(work.uuid, index=esManager.index)
+                                workRec = ESWork.get(work[0], index=esManager.index)
```
mitri-slory, line 46: Why is work.uuid replaced with work[0] here? Does it work better to find the exact work we're looking for?

mitri-slory (commented):
This script looks really nice Mike. Great to see that it's sped up the process. I only have questions about what work[0] represents and how the query you made allowed work[0] to be editionMeasurements based on your code. I also found some other exceptions like ValueError and ConflictError when I ran the original non multi-processed script in production, so I'll commit those to this script soon.

scripts/govDocES.py
mwbenowitz: When you change the query from just `Work` to `Work.uuid, ...` it changes the behavior. Instead of getting an ORM object you get a tuple of the fields you requested. I can update this to make that clearer (also this should be `[1]` here!

mitri-slory: Appreciate the explanation.

scripts/govDocES.py
mitri-slory: You already answered this in my previous question.

scripts/govDocES.py
```diff
@@ -36,14 +36,15 @@ def main():
         .filter(Edition.measurements != [{}]) \
         .group_by(Work.uuid) \
         .yield_per(batchSize):
+            uuid, editionMeasurementArray = work
```
mitri-slory, line 39: This change makes the work query way clearer than before.

model/postgres/edition.py
```diff
@@ -48,7 +48,10 @@ class Edition(Base, Core):
     links = relationship('Link', secondary=EDITION_LINKS, backref='editions', cascade='all, delete')
     rights = relationship('Rights', secondary=EDITION_RIGHTS, backref='editions', cascade='all, delete')
 
-    __tableargs__ = (Index('ix_editions_dcdw_uuids', dcdw_uuids, postgresql_using="gin"))
+    __tableargs__ = (
+        Index('ix_editions_dcdw_uuids', dcdw_uuids, postgresql_using="gin"),
+        Index('ix_editions_measurements', measurements, postgresql_using='gin')
+    )
```
mitri-slory, line 54: Great use of indexes to speed up the process of the govDocES script.

mwbenowitz:
@mitri-slory Are there further changes to this script we should make? Otherwise I'll move it out of draft and we can work on getting it merged in.

mitri-slory:
> @mitri-slory Are there further changes to this script we should make? Otherwise I'll move it out of draft and we can work on getting it merged in.

I believe the script is fine the way it is now. I think we should focus on getting it merged now.

mwbenowitz:
Sounds good, will tackle that on Monday!

mitri-slory (approved):
Everything looks good in this script and the migration. I believe we're ready to merge this in now.

PR #188, opened 2022-07-29
Added govDocFilter and aggregation

mitri-slory:


api/elastic.py
```diff
@@ -389,16 +393,32 @@ def createFilterClausesAndAggregations(self, filterParams):
                 'filter', **{'terms': {'editions.formats': formats}}
             )
 
+        if len(govDocFilters) > 0:
+            # onlyGovDoc|noGovDoc|all
+            # if govDocFilter is all: skip
+            # elif govDocFilter is no: is_government_document: false
+            # elif govdocFilter is only: is_gov_doc: true
+            if govFilter is 'noGovDoc':
```
mwbenowitz, line 401: `govFilter` is still `None` here, should look at `govDocFilters` which is an array of 2-tuples in the form of `(filterName, filterValue)`. There should only ever be one tuple in the array though, which makes this easier.

api/elastic.py
```diff
@@ -389,16 +393,32 @@ def createFilterClausesAndAggregations(self, filterParams):
...
+                dateFilter = Q('term', **{'is_government_document': False})
+                dateAggregation = A(
```
mwbenowitz, line 403: should be `govFilter` and `govAggregation` here and below

mwbenowitz (changes requested):
Unfortunately this turns out to be a bit harder than I had hoped. This looks good, just a few tweaks, but there are some bad assumptions elsewhere in the code that block this from working. Specifically this is in the `applyFilters` method, which I think needs to be refactored a bit to accommodate this.

We should also add tests and expand the Swagger docs slightly to include this change.

mitri-slory:
Should we also add on to the appliedAggregation method? That's why I left a comment in the method about adding one for govDocFilters.

api/elastic.py
```diff
@@ -389,6 +393,24 @@ def createFilterClausesAndAggregations(self, filterParams):
...
 
+        if len(govDocFilters) > 0:
+            # onlyGovDoc|noGovDoc|all
+            # if govDocFilter is all: skip
+            # elif govDocFilter is no: is_government_document: false
+            # elif govdocFilter is only: is_gov_doc: true
+            if govDocFilters[0][1] == 'noGovDoc':
+                self.govDocFilters = [
+                    Q('term', **{'is_government_document': False})
+                ]
```
mwbenowitz, line 404: I don't think this needs to be an array, because it's a boolean filter it only makes sense for it to only be set once. So this could be `self.govDocFilter`, which should simplify things below!

mwbenowitz (commented):
Heading in the right direction here! A few things should get shifted around but I think we're pretty close to a workable solution.

api/elastic.py
```diff
@@ -479,12 +501,15 @@ def applyAggregations(self):
...
             .bucket('editions_per', 'reverse_nested')
 
+
         lastAgg.bucket(
             'formats', 'terms',
             **{'field': 'editions.formats', 'size': 10}
         )\
             .bucket('editions_per', 'reverse_nested')
 
+        self.query.aggs.bucket('govDoc', 'terms', **{'field': 'is_government_document', 'size': 10})
```
mwbenowitz, line 511: `size` can be 2 here since it's a boolean value

tests/unit/test_api_es.py
```diff
@@ -869,6 +913,50 @@ def test_applyFilters_language_filter_only(self, testInstance, mocker):
...
+
+        testInstance.query = Search().query(Q())
+        testInstance.dateSort = None
+        testInstance.languageFilters = []
+        testInstance.appliedFilters = []
+        testInstance.govDocFilter = Q('term', **{'is_government_document': True})
+
+        testInstance.applyFilters()
+
+        pprint.pprint(testInstance.query.to_dict(), indent=2)
```
mwbenowitz, line 926: nit: drop any `print` statements left over from development!

api/elastic.py
```diff
@@ -462,15 +480,19 @@ def applyFilters(self, size=100):
...
     def applyAggregations(self):
         rootAgg = self.query.aggs.bucket(
             'editions', A('nested', path='editions')
         )
 
         lastAgg = rootAgg
         for i, agg in enumerate(self.appliedAggregations):
-            currentAgg = 'edition_filter_{}'.format(i)
-            lastAgg = lastAgg.bucket(currentAgg, agg)
+            if agg != A('filter', **{'term': {'is_government_document': False}}) and agg != A('filter', **{'term': {'is_government_document': True}}):
```
mwbenowitz, line 493: A few thoughts, doesn't look like the aggregation object is being constructed above, so this isn't being used right now.

Also I'm not sure that testing for equivalence of `A()` objects will work, depending on the implementation.

I'd get rid of this, and add the aggregation similarly as we did with the filter, as an attribute and then setting down where we create the bucket.

api/elastic.py
```diff
@@ -462,15 +478,19 @@ def applyFilters(self, size=100):
...
             'editions', A('nested', path='editions')
         )
 
         lastAgg = rootAgg
         for i, agg in enumerate(self.appliedAggregations):
-            currentAgg = 'edition_filter_{}'.format(i)
-            lastAgg = lastAgg.bucket(currentAgg, agg)
+                currentAgg = 'edition_filter_{}'.format(i)
+                lastAgg = lastAgg.bucket(currentAgg, agg)
+
```
mwbenowitz, line 493: nit: fix over-indentation here

mitri-slory: Just committed this fix.

tests/unit/test_api_es.py
```diff
@@ -909,28 +997,58 @@ def test_applyFilters_no_filters_date_ort(self, testInstance, mocker):
...
+                                        {'aggs': 
+                                            {'languages': 
+                                                {'aggs': 
+                                                    {'editions_per': {'reverse_nested': {}}},
+                                                'terms': {'field': 'editions.languages.language', 'size': 200}}},
+                                            'nested': {'path': 'editions.languages'}}},
+                                        'filter': {'exists': {'field': 'editions.formats'}}}},
+                                    'filter': {'terms': {'editions.formats': 'testFormat'}}}},
+                                'filter': {'range': {'editions.publication_date': 'testDate'}}}},
+                            'nested': {'path': 'editions'}}
```
mwbenowitz, line 1049: I'm torn on this (and for the record, don't change it).

I love the detail but it's tough to have this live in this file (just due to length). We have a `fixtures` directory that is used for a few tests but not widely. It might be good in the future to take blocks like this, store them as a JSON file and then load them as a test fixture that we can reference. Feels more maintainable to me.

mwbenowitz (approved):
This looks great!! 👍

tests/unit/test_api_es.py
mitri-slory: That makes sense. We can revisit this topic when we update and create future unit tests.

PR #189, opened 2022-08-08
New script to delete ESC works not in Postgresql

mitri-slory:
I used the `govDocES` script as a basis for this script and the only question I have is how to iterate through the ESC works because I just put `esManager` instead of `dbManager` in front of the `session.query` code line for now.

### DIFF
--- a/scripts/deleteElasticRecords.py
+++ b/scripts/deleteElasticRecords.py
@@ -28,9 +28,9 @@ def main():
 
     dbManager.createSession()
 
-    batchSize = 10
+    batchSize = 100
     esWorkUUIDS = []
-    workUUIDTuples = []
+    psqlWorkUUIDTuples = []
     searchES = Search(index=os.environ['ELASTICSEARCH_INDEX'])
 
     for work in searchES.query('match_all').scan():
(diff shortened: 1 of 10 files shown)


scripts/deleteElasticRecords.py
```diff
@@ -0,0 +1,42 @@
...
+    esManager = ElasticsearchManager()
+    esManager.createElasticConnection()
+
+    dbManager.generateEngine()
+
+    dbManager.createSession()
+
+    batchSize = 1000
+    for work in esManager.session.query(ESWork) \
+        .yield_per(batchSize):
```
mwbenowitz, line 31: You'll need to construct a Search object the way that the API does, but there is an added complexity that ElasticSearch doesn't do "deep" paging well by default (not really it's intended use case).

The recommended solution is to use scan: see here: https://elasticsearch-dsl.readthedocs.io/en/latest/api.html#elasticsearch_dsl.Search.scan

The implementation should be straightforward, but I'm not sure about performance. If that's a problem there could be other approaches that we take.

scripts/deleteElasticRecords.py
```diff
@@ -0,0 +1,42 @@
...
+
+    batchSize = 1000
+    for work in esManager.session.query(ESWork) \
+        .yield_per(batchSize):       
+
+        if dbManager.session.get(Work, {'uuid': work.uuid}) == None:
+            try:
+                ESWork.delete(work.uuid)
+            except ValueError:
+                print('Empty value')
```
mwbenowitz, line 37: This totally works! But it might be more efficient if we batch the database queries together, say by searching for 100 UUIDs and then comparing the result. Doing 2.7m queries will be slow, so if we can make that more efficient then we can speed this up.

mwbenowitz (commented):
Good start! Left comments about tackling the search and perhaps improving the performance.

scripts/deleteElasticRecords.py
```diff
@@ -0,0 +1,53 @@
...
+    dbManager.createSession()
+
+    batchSize = 200
+    searchES = Search(index=os.environ['ELASTICSEARCH_INDEX'])
+
+    for work in searchES.query('match_all').scan():    
+        if dbManager.session.query(Work) \
+            .group_by(Work.uuid) \
+            .filter(Work.uuid == work.uuid) \
+            .yield_per(batchSize) == None:
```
mwbenowitz, line 38: Right approach! but since `work.uuid` is only going to yield one value this will only ever return one work.

You need to add an accumulator above this. Something like:
```
workUUIDs.append(work.uuid) # Initialize the array outside of the loop here
if len(workUUIDs) >= batchSize:
    # run the database query with
    uuidTuples = dbManager.session.query(Work.uuid).filter(Work.uuid == work.uuid).filter(Work.uuid._in(workUUIDs).all()
    # Upack the tuples to a flat array
    # Convert the array of ElasticSearch UUIDs and the array of postgresql UUIDs to sets()
    # Take the difference of of the ES set and the postgresql set (https://docs.python.org/3/library/stdtypes.html#frozenset.difference)
    # what you have left should be a set of UUIDs that are in ES but not in psql
    # that's what you should delete!

**Review by mwbenowitz — COMMENTED** [review]
Getting closer but the batching approach needs to be updated.

The general idea of the comment I left is you want to create collections of UUIDs from ElasticSearch and postgresql for each batch, compare them, and then delete any UUIDs from ElasticSearch that are missing from postgres.

A bit more complicated than the straight approach but it should be a lot faster!

File: scripts/deleteElasticRecords.py [hunk 4]
```diff
@@ -28,26 +29,51 @@ def main():
 
     dbManager.createSession()
 
-    batchSize = 200
+    batchSize = 1000
+    esWorkUUIDS = deque()
+    psqlWorkUUIDTuples = deque()
```
mwbenowitz, line 34: What's the advantage of `deque` here over starting with `set`?

scripts/deleteElasticRecords.py
```diff
@@ -28,26 +29,51 @@ def main():
...
+    
     dbManager.closeConnection()
 
+def findESOnlyWorks(esWorkUUIDS, psqlWorkUUIDTuples, dbManager, esManager):
+        for uuid in esWorkUUIDS:
+            uuidTuple = dbManager.session.query(Work.uuid) \
+                .filter(Work.uuid == uuid) \
+                .filter(Work.uuid.in_(esWorkUUIDS)).first()
+            if uuidTuple != None:
+                psqlWorkUUIDTuples.append(str(uuidTuple[0])) #PSQL work uuids are of type uuid unlike ES work uuids
```
mwbenowitz, line 60: I think we should be able to avoid the need for a loop here and just use `extend` with a list comprehension or something similar.

Or main thing is executing only one database query, how the UUIDs make their way into `psqlWorkUUIDTuples` is less concerning from a performance standpoint

scripts/deleteElasticRecords.py
```diff
@@ -28,26 +29,51 @@ def main():
...
+
+        setESWorkUUIDS = set(esWorkUUIDS) 
+        setPSQLWorkUUIDS = set(psqlWorkUUIDTuples)
+
+        onlyESWorkUUIDS = setESWorkUUIDS.difference(setPSQLWorkUUIDS) #Set should be empty
```
mwbenowitz, line 65: What does the comment mean here? I would assume the set _shouldn't_ be empty!

mwbenowitz (commented):
Looks great, just mainly have a question about iterating through the database results.

Will be interesting to validate that `deleteWorkRecords` actually works since that is one of the possible culprits as the source of the issue (would be actually good if it was since we could then confidently diagnose the issue).

scripts/deleteElasticRecords.py
mwbenowitz: I didn't know that about `deque` thanks for the info!

scripts/deleteElasticRecords.py
mitri-slory: Are you referring to do something like this: ```psqlWorkUUIDTuples.update({uuid for uuid in esWorkUUIDS if dbManager.session.query(text(uuid)) != None})```

scripts/deleteElasticRecords.py
mitri-slory: I was thinking of the cases where the ES records are also in PSQL which is usually most of the time when I wrote this comment. Most results I've seen in the terminal when printing onlyESWorkUUIDS is `set()`, but there are a few like
`{'084e9ab3-dfa4-4819-8536-0c31d5da5c46'}` that are the records we are looking for. I can see how you can be confused by my comment there.

mwbenowitz: ah I see okay thanks! So we expect the set to usually be empty, that makes sense since we're looking for a relatively small number of records

mwbenowitz (approved):
Looks good to me! curious to see what kind of performance we get out of this.

PR #190, opened 2022-08-08
HOTFIX Remove trailing slashes from API Endpoints

mwbenowitz:
This removes trailing slashes from two API endpoints -- `search` and `collection`. This misconfiguration was causing issues with the Swagger documentation, resulting in false errors when interacting with the API.

The only effect of this change should be to resolve those errors, any other interaction with the API should result in a `308` redirect, which is invisible in other use cases.


mitri-slory (approved):
This hotfix looks good and ready to go.

PR #191, opened 2022-08-22
SFR-153 Ingest Report Fix

mwbenowitz:
This addresses an issue raised by the project/product teams to fix broken reporting of daily ingest volume. There are two related fixes for this issue in this PR:
1. The immediate issue is resolved by updating the analytics queries. an ineffecient strategy was being used to create date-based queries. This has been removed and replaced with a more standard query structure.
2. This is supported by the addition of a database maintenance process. Without this standard process database table can grow slow and harder to query. This provides a tool that can be used to ensure that the database is regularly pruned of unused rows.

Overall this should restore intended functionality to the database and ensure that similar situations do not occur in the future. It should also provide a safeguard against degradation and improve performance, if in a small way.

### DIFF
--- a/processes/ingestReport.py
+++ b/processes/ingestReport.py
@@ -1,7 +1,6 @@
-from datetime import date, timedelta
-from sqlalchemy import func
-from sqlalchemy.dialects.postgresql import DATE
+from datetime import datetime, timedelta
 import newrelic.agent
+from sqlalchemy import func
 
 from .core import CoreProcess
 from model import Work, Edition, Item, Record
@@ -9,6 +8,7 @@ from managers import SmartSheetManager
 from logger import createLog
 
 logger = createLog(__name__)
+
 
 class IngestReportProcess(CoreProcess):
     def __init__(self, *args):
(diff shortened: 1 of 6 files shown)


processes/ingestReport.py
```diff
@@ -27,28 +27,32 @@ def runProcess(self):
 
     @newrelic.agent.background_task()
     def generateReport(self):
-        ingestDate = date.today() - timedelta(days=1)
+        ingestDate = datetime.today()
+        ingestDateStart = ingestDate.replace(hour=0, minute=0, second=0)
+        ingestDateEnd = ingestDateStart + timedelta(days=1)
```
mitri-slory, line 32: What's the purpose of specifically replacing the hours, minutes, and seconds for the ingestDateStart variable?

mwbenowitz: I'm pretty sure `datetime.today()` populates those fields with the values for when the `datetime` object is created. We need `datetime` for the start and end of each day, and this was the easiest way I could think to do that.

If you know a better way of doing that I'm open to it!

mitri-slory: Oh okay. That makes sense since the start of each day would naturally be all 0s in the 24 hour time format. I can't think of another way to implement this so it's fine with me.

PR #195, opened 2022-09-23
Updated ES API generateSearchQuery method

mitri-slory:
With this PR I added a new conditional for the `generateSearchQuery` method to include identifier as a field and made a `identifierQuery` method to account for that. In addition, instead of adding another conditional for an authority field, I decided to update the existing viaf and lcnaf conditional since they seemingly utilize an authorityQuery method which is similar to what we want for when one searches for a specific authority and the identifier afterwards. What are your thoughts on doing this rather than creating a separate conditional and new authorityQuery method? Also, I only included isbn and lccn as new authorities for right now, since I'm not aware and don't have access to the ES QA database to check all the types of authorities that exist for each record. Could you assist me on finding all the types of authorities that exist in the ES database?

### DIFF
--- a/api/elastic.py
+++ b/api/elastic.py
@@ -73,9 +73,7 @@ class ElasticClient():
             elif field == 'viaf' or field == 'lcnaf':
                 searchClauses.append(self.authorityQuery(field, escapedQuery))
             elif field == 'identifier':
-                searchClauses.append(self.identifierQuery(escapedQuery))
-            elif field == 'authority: identifier':
-                searchClauses.append(self.authIdentQuery(authorityList, escapedQuery))
+                searchClauses.append(self.identifierQuery(authorityList, escapedQuery))
             else:
                 searchClauses.append(Q('match', **{field: escapedQuery}))
 
@@ -233,15 +231,20 @@ class ElasticClient():
                 )
             )
         ])
-    
-    #Query for the identifier value
-    def identifierQuery(self, identifierText):
+
+    #Query for the identifier authority followed up with the identifier value
+    def identifierQuery(self, authorityList, authIdentText):
         self.searchedFields.extend(['identifiers', 'editions.identifiers'])
 
-        return Q('bool', should=[
+        authIdentList = authIdentText.split(': ')
+
+        #When user only types in the identifier in the search bar
+        if len(authIdentList) < 2:
+
+            return Q('bool', should=[
             Q(
                 'query_string',
-                query=identifierText,
+                query=authIdentText,
                 fields=['identifiers.*'],
                 default_operator='and'
             ),
@@ -250,22 +253,19 @@ class ElasticClient():
                 path='editions',
                 query=Q(
                     'query_string',
-                    query=identifierText,
+                    query=authIdentText,
                     fields=['editions.identifiers.*'],
                     default_operator='and'
                 )
             )
-        ])
+            ])
 
-    #Query for the identifier authority followed up with the identifier value
-    def authIdentQuery(self, authorityList, authIdentText):
-        self.searchedFields.extend(['identifiers', 'editions.identifiers'])
+        #When user types in the authority and identifier in the search bar
+        else:
 
-        authIdentList = authIdentText.split(': ')
-        authority = authIdentList[0]
-        identifier = authIdentList[1]
+            authority = authIdentList[0]
+            identifier = authIdentList[1]
 
-        if len(authIdentList) < 3:
             if authority in authorityList:
 
                 return Q('bool', should=[
--- a/api/elastic.py
+++ b/api/elastic.py
@@ -49,7 +49,8 @@ class ElasticClient():
         return self.executeSearchQuery(params, page, perPage)
 
     def generateSearchQuery(self, params):
-        authorityList =['isbn', 'lccn']
+        authorityList =['isbn', 'issn', 'lcc', 'lccn', 'oclc', 'nypl', 'hathi', 'gutenberg', 'doab']
+
         search = self.createSearch()
         search.source(['uuid', 'editions'])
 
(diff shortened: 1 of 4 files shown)


api/elastic.py
```diff
@@ -60,16 +61,20 @@ def generateSearchQuery(self, params):
                 searchClauses.append(Q('bool', should=[
                     self.titleQuery(escapedQuery),
                     self.authorQuery(escapedQuery),
-                    self.subjectQuery(escapedQuery)
+                    self.subjectQuery(escapedQuery),
+                    self.identifierQuery(escapedQuery),
+                    self.authorityQuery(escapedQuery)
```
mwbenowitz, line 66: I think we should leave this out of this general query. The authority query is more specific and is mostly used internally.

api/elastic.py
```diff
@@ -60,16 +61,20 @@ def generateSearchQuery(self, params):
...
                 ]))
             elif field == 'title':
                 searchClauses.append(self.titleQuery(escapedQuery))
             elif field == 'author':
                 searchClauses.append(self.authorQuery(escapedQuery))
             elif field == 'subject':
                 searchClauses.append(self.subjectQuery(escapedQuery))
-            elif field == 'viaf' or field == 'lcnaf':
+            elif field == 'viaf' or field == 'lcnaf' or field in authorityList:
```
mwbenowitz, line 74: If we want to do this we need to extend the `authorityQuery` method, which I'm not opposed to, but might be unnecessary depending on how we implement the identifier query (see comment below)

api/elastic.py
```diff
@@ -256,6 +283,29 @@ def authorQuery(self, authorText):
...
+                query=Q(
+                    'query_string',
+                    query=identifierText,
+                    fields=['editions.identifiers.*'],
+                    default_operator='and'
+                )
+            )
+        ])
+
+
```
mwbenowitz, line 308: duplicate code here!

api/elastic.py
```diff
@@ -227,6 +232,28 @@ def titleQuery(self, titleText):
...
         ])
+    
+    def identifierQuery(self, identifierText):
+        self.searchedFields.extend(['identifiers', 'editions.identifiers'])
+
+        return Q('bool', should=[
+            Q(
+                'query_string',
+                query=identifierText,
+                fields=['identifiers.*'],
```
mwbenowitz, line 243: This will definitely work, but I wonder if we should have it work along the lines of the `authorityQuery` where we break out the identifier authority and value and look up that way. It's a bit more complicated but I think would be more accurate and precise.

Unless we already discussed and decided to take a different approach, I'm not sure

mwbenowitz: It would mean splitting the string and handling those two values separately, but I don't think that's a crazy amount of complexity to add

api/utils.py
```diff
@@ -8,7 +8,7 @@
 
 class APIUtils():
     QUERY_TERMS = [
-        'keyword', 'title', 'author', 'subject', 'viaf', 'lcnaf',
+        'keyword', 'title', 'author', 'subject', 'lccn', 'isbn', 'viaf', 'lcnaf',
```
mwbenowitz, line 11: If we don't extend `authorityQuery` we can drop this update

mwbenowitz (commented):
Overall really good (and something that would work, I think), but have a few questions about the desired functionality regarding how we query for identifiers. I think allowing for a specific query of the authority and identifier fields was the original intention.

I think extending the authority query is an interesting idea but would require some major updates to how that works since they are querying separate documents, might be clearer to keep the separate.

As for finding all of the identifier authorities, you should be able to do a SELECT DISTINCT query in the `identifiers` table in postgresql!

api/elastic.py
mitri-slory: Ok. I'll delete lines 65 and 66.

api/elastic.py
mitri-slory: Ah. Thanks for catching that.

api/elastic.py
mitri-slory: From my TAD, we decided to have one query for the user to search for a work just based on the identifier value and another query for the user to search based on the authority and value. Based on this, I believe the identifierQuery method can stay the same, but we can add an authIdentQuery method that does what you suggest with splitting up a string and handling the two values.

api/utils.py
mitri-slory: I agree in not extending the authorityQuery and instead making a new method.

mitri-slory: So I'll delete `lccn` and `isbn`

api/elastic.py
mwbenowitz: I think that works! Another approach would be to have a method that checks for the presence of an authority and if present adds that as part of the query and if not just does the search by identifier value, but I leave that up to you!

api/elastic.py
```diff
@@ -70,6 +72,10 @@ def generateSearchQuery(self, params):
                 searchClauses.append(self.subjectQuery(escapedQuery))
             elif field == 'viaf' or field == 'lcnaf':
                 searchClauses.append(self.authorityQuery(field, escapedQuery))
+            elif field == 'identifier':
+                searchClauses.append(self.identifierQuery(escapedQuery))
+            elif field == 'authority: identifier':
+                searchClauses.append(self.authIdentQuery(authorityList, escapedQuery))
```
mwbenowitz, line 78: I think we can/should roll this into one method. The `authority: identifier` will be hard to pass in as part of a parameter string and probably less user friendly. I think it's a relatively simple update that I'll outline below.

api/elastic.py
```diff
@@ -227,6 +233,65 @@ def titleQuery(self, titleText):
...
+        ])
+
+    #Query for the identifier authority followed up with the identifier value
+    def authIdentQuery(self, authorityList, authIdentText):
+        self.searchedFields.extend(['identifiers', 'editions.identifiers'])
+
+        authIdentList = authIdentText.split(': ')
+        authority = authIdentList[0]
+        identifier = authIdentList[1]
+
```
mwbenowitz, line 267: For updating this I think we can look at the length of this array. If it's less than 2 we run the query from `identifierQuery` if it's equal to 2 then we run the query below. That's it!

mwbenowitz (commented):
I really just have one thought about simplifying this and ensuring it's easy to use for people/applications calling the API.

This should ensure that we have a flexible identifier search endpoint that accomodates the use cases we expect to see.

mwbenowitz (approved):
Looks great! Let's get this into QA and see how it does. It's a very nice added bit of functionality.

PR #196, opened 2022-09-26
Sfr 1539 multi host es

mitri-slory:
I added a if/else conditional for the cases where there is a single or multiple hosts for ES.

### DIFF
--- a/managers/elasticsearch.py
+++ b/managers/elasticsearch.py
@@ -29,32 +29,27 @@ class ElasticsearchManager:
 
         creds = '{}:{}@'.format(user, pswd) if user and pswd else ''
 
+        #Allowing multple hosts for a ES connection
         multHosts = []
 
-        if isinstance(host, list) or \
-            isinstance(host, dict) or \
-            isinstance(host, set) or \
-            isinstance(host, tuple):
+        if ',' not in host:
+            host = '{}://{}{}:{}'.format(scheme, creds, host, port)
+
+            multHosts.append(host)
+
+        else:
+            hostList = host.split(', ')
             
-            for i in host:
+            for i in hostList:
                 multHosts.append('{}://{}{}:{}'.format(scheme, creds, i, port))
             
-            connectionConfig = {
+        connectionConfig = {
                 'hosts': multHosts,
                 'timeout': timeout,
                 'retry_on_timeout': True,
                 'max_retries': 3
             }
 
-        else:
-            host = '{}://{}{}:{}'.format(scheme, creds, host, port)
-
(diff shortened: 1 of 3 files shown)


managers/elasticsearch.py
```diff
@@ -29,14 +29,32 @@ def createElasticConnection(self, scheme=None, host=None, port=None, user=None,
 
         creds = '{}:{}@'.format(user, pswd) if user and pswd else ''
 
-        host = '{}://{}{}:{}'.format(scheme, creds, host, port)
+        multHosts = []
+
+        if isinstance(host, list) or \
+            isinstance(host, dict) or \
+            isinstance(host, set) or \
+            isinstance(host, tuple):
```
mwbenowitz, line 37: minor improvement here `isinstance` can take a tuple of types to check against!

managers/elasticsearch.py
```diff
@@ -29,14 +29,32 @@ def createElasticConnection(self, scheme=None, host=None, port=None, user=None,
...
+            
+            for i in host:
```
mwbenowitz, line 39: I think it might be necessary for `host` to be a string because of how it's stored in AWS, not sure we can get it out of there as an iterable. Would just need to pick a delimiter and then split/iterate on that.

managers/elasticsearch.py
```diff
@@ -29,14 +29,32 @@ def createElasticConnection(self, scheme=None, host=None, port=None, user=None,
...
-        }
+        else:
+            host = '{}://{}{}:{}'.format(scheme, creds, host, port)
+
+            connectionConfig = {
+                'hosts': [host],
+                'timeout': timeout,
+                'retry_on_timeout': True,
+                'max_retries': 3
+            }
```
mwbenowitz, line 57: opportunity to make this DRY-er, just add this to the existing `multHosts` array and then have only one `connectionConfig` block, makes it easier to maintain in the future if we want to update this config!

mwbenowitz (commented):
Looks great, only concern is about being able to provide a string to this configuration in our AWS hosted environments, might need changing for that reason only (though also worth looking into if that's true)

managers/elasticsearch.py
mitri-slory: I'll do just that.

managers/elasticsearch.py
mitri-slory: That makes sense so I'll pick a comma with a space as the delimiter and split then iterate on that.

mwbenowitz (approved):
This looks great!

PR #197, opened 2022-09-28
SFR-1540: Added script to reindex ES works

mitri-slory:
This script includes connecting to the ES manager and creating a ES index with the portion commented out to be code that I will type into the console of the Elastic.co app with the current ES QA configurations filling in the blanks. Ashok from DevOps told me he would add the proper integrations and basic authentication to the new ES cluster and he will also have a look at the old one from QA as well. Once that integration is finished, then it should be a good time to run the commented out code into the Elastic console.


mwbenowitz (commented):
The plan for the reindex sounds great, just left a comment about using a reindex helper method here, otherwise sounds good!

scripts/esReIndex.py
```diff
@@ -0,0 +1,35 @@
...
+def main():
+
+    '''Re-indexing works into new ES cluster'''
+
+    loadEnvFile('local-qa', fileString='config/{}.yaml')
+
+    esManager = ElasticsearchManager()
+    esManager.createElasticConnection()
+    esManager.createElasticSearchIndex()
+
```
mwbenowitz, line 15: Import `ESWork` and run `ESWork.init()` here

mitri-slory: When I put ESWork.init() at line 15 and run the script I get this ValueError.

raise ValueError("You cannot perform API calls on the default index.")
ValueError: You cannot perform API calls on the default index.

mwbenowitz: oh, I know why this is failing, the `scripts/__init__.py` file is causing the import to happen first still. That traces back to `import scripts` from `scriptRunner.py`. To get this to work correctly we'd have to set the environment variables _before_ that import line, which is possible but a bigger refactor job. Would actually be a general improvement to prevent this kind of thing in the future.

mitri-slory: Thanks for the suggestion. I just did that and now the `drb_dcdw_qa_new_cluster` mapping is now filled out when I run the script.

mwbenowitz (approved):

PR #198, opened 2022-10-05
Updated search score in agent parser method

mitri-slory:
I updated the search score to 0.85 instead of 0.9 and specified the version number of the `python-levenshtein` dependency in the requirements file to be 0.20.4.


mwbenowitz (approved):
👍 👍 👍

PR #199, opened 2022-10-18
SFR-1509v2: Updated identifier query for nested search

mitri-slory:
I changed the identifier queries to resemble the author queries since both of them utilize nested fields when searching. Due to this change, I adjusted the unit tests as well.

### DIFF
--- a/api/elastic.py
+++ b/api/elastic.py
@@ -245,22 +245,12 @@ class ElasticClient():
             identifier = authIdentText
 
             workIdentQuery = Q('bool', must=[
-                Q(
-                    'term',
-                    query=identifier,
-                    fields=['identifiers.identifier'],
-                    default_operator='and'
-                )
-            ])
+                Q('term', **{'identifiers.identifier': identifier})
+                ])
 
             editionIdentQuery = Q('bool', must=[
-                Q(
-                    'term',
-                    query=identifier,
-                    fields=['editions.identifiers.identifier'],
-                    default_operator='and'
-                )
-            ])
+                Q('term', **{'editions.identifiers.identifier': identifier})
+                ])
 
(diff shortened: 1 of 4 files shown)


api/elastic.py
```diff
@@ -234,30 +234,37 @@ def titleQuery(self, titleText):
...
                 )
-            )
+            ])
+
+            editionIdentQuery = Q('bool', must=[
+                Q(
+                    'term',
+                    query=identifier,
+                    fields=['editions.identifiers.identifier'],
+                    default_operator='and'
```
mwbenowitz, line 261: with the `term` query we can simplify this because we don't need these specific keys, actually just need field and query: `'term', **{'editions.identifiers.identifier': identifier} (need the dictionary unpacking because we need the string as the key)

api/elastic.py
```diff
@@ -268,29 +275,47 @@ def identifierQuery(self, authorityList, authIdentText):
...
-                            query=identifier,
-                            fields=['editions.identifiers.identifier'],
-                            default_operator='and'
-                        )
-                    )
+                ])
+
+                return Q('bool', should=[
+                    Q('nested', path='identifiers', query=workAuthQuery),
+                    Q('nested', path='editions.identifiers', query=editionAuthQuery),
```
mwbenowitz, line 316: These should be clustered in a nested `bool` query. That will give us only documents that match both authority and identifier. Right now this will give us _either_ authority or identifier matches.

mwbenowitz (commented):
A few more improvements to the query construction and this is there.

Appreciate the approach to constructing the nested queries, the syntax here is really clear.

mwbenowitz (approved):
👍 👍 👍

PR #200, opened 2022-11-08
SFR-1543: Adding Read Online Link to Collections

mitri-slory:
This PR focuses on updating the collectionFetch endpoint with the Read Online link that was previously not visible such as this [https://digital-research-books-beta.nypl.org/read/543649](url). I'm not sure if the rel, type, and id in the url I typed are the correct ones, but the format is similar to the other links in the array. Sean gave an example of a collection with no Read Online link on Jira under the comments of the SFR-1543 ticket as well. Also, I removed the `loadEnvFile` from all the modules in the `scripts` directory since the `scriptrunner.py` module has it already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Multi-host configuration for ES connections
 - Script to reindex ES works into new ES cluster
 - Logstash file for reindexing ES works for new cluster
+- Updated links array for collectionFetch endpoint with Read Online link
 
 ### Fixed
 - Updated unit tests for developmentSetup process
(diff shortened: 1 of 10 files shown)


mwbenowitz (commented):
Your approach is the right instinct but actually I _tried_ to address this elsewhere and I think we should look there for the solution. In the `parseEditionToPublication` method we're already adding these links (see here: https://github.com/NYPL/drb-etl-pipeline/blob/ade4e66641b8e52e8e80e4c4477c07db53c232d0/api/opds2/publication.py#L168) but I think that block needs to be made smarter so it includes links to our read online page as well as the download options.

That matches with what Sean reporting seeing in the actual implementation.

PR #201, opened 2022-11-14
Sfr 1570 sk learn dependency update

mitri-slory:


scriptRunner.py
```diff
@@ -2,7 +2,7 @@
 import sys
 from main import loadEnvFile
 
-loadEnvFile('local-qa', fileString='config/{}.yaml')
+loadEnvFile('local-qa-old', fileString='config/{}.yaml')
```
mwbenowitz, line 5: Is this intended or is this a dangling change from some other work?

mwbenowitz (approved):
Update looks good, just wondering about some of the changes from the scripts!

scriptRunner.py
mitri-slory: This is intended since I thought this quick ticket was a good opportunity to adjust loadEnvFile for the scriptrunner  module and delete it from the other scripts.

PR #204, opened 2022-11-15
Replaced fasttext with gensim

mitri-slory:
Had to make a new PR to replace the old one because It was merged accidentally into main.

### DIFF
--- a/managers/oclcClassify.py
+++ b/managers/oclcClassify.py
@@ -1,4 +1,4 @@
-import fasttext
+from gensim.models import Word2Vec
 from lxml import etree
 import re
 import requests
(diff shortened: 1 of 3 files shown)


managers/oclcClassify.py
```diff
@@ -43,7 +43,7 @@ class ClassifyManager:
     ]
 
     # This is a compressed 917kb file that contains a language detection model
-    LANG_MODEL = fasttext.load_model('lid.176.ftz')
+    LANG_MODEL = Word2Vec.load('lid.176.ftz')
```
mwbenowitz, line 46: From the tests it seems like gensim can't read this file, not sure if there's another way to load it or if we need to take a different approach

PR #205, opened 2022-11-15
SFR-1543v2: Read Online Link in Publication Class

mitri-slory:
I created this new PR because the old one got merged accidentally into main before we completed it.

### DIFF
--- a/api/opds2/publication.py
+++ b/api/opds2/publication.py
@@ -165,8 +165,8 @@ class Publication:
         # Covers
         self.findAndAddCover(editionRecord)
 
-        host = 'digital-research-books-beta'\
-            if os.environ['ENVIRONMENT'] == 'production' else 'drb-qa'
+        #host = 'digital-research-books-beta'\
+            #if os.environ['ENVIRONMENT'] == 'production' else 'drb-qa'
 
         # Acquisition Links
         for item in editionRecord.items:
@@ -180,7 +180,7 @@ class Publication:
                 if 'reader' in link.flags:
                     if link.flags['reader'] == True:
                         self.addLink({
-                        'href': 'https://{}.nypl.org/read/{}'.format(host, link.id),
+                        'href': 'https://drb-qa.nypl.org/read/{}'.format(link.id),
                         'type': link.media_type,
                         'rel': 'http://opds-spec.org/acquisition/open-access'
                     })
--- a/api/opds2/publication.py
+++ b/api/opds2/publication.py
@@ -182,8 +182,8 @@ class Publication:
                         self.addLink({
                         'href': 'https://{}.nypl.org/read/{}'.format(host, link.id),
                         'type': link.media_type,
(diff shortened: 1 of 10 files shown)


api/opds2/publication.py
```diff
@@ -165,6 +165,16 @@ def parseEditionToPublication(self, editionRecord):
...
         self.findAndAddCover(editionRecord)
 
+        host = 'digital-research-books-beta'\
+        if os.environ['ENVIRONMENT'] == 'production' else 'drb-qa'
+
+        # Read Online Link
+        self.addLink({
+                'href': 'https://{}.nypl.org/read/{}'.format(host, editionRecord.id),
+                'rel': 'alternate',
+                'type': 'application/pdf'
```
mwbenowitz, line 175: I'd look down at L182-184 for what this should be. The `/read/{}` path should contain a `link.id` not an `edition.id`. And it should only be done when we have a flag of `readable:True`. So I think it needs to be done in that loop and have some kind of conditional for what kind of object should be passed to `self.addLink`

api/opds2/publication.py
```diff
@@ -165,6 +165,9 @@ def parseEditionToPublication(self, editionRecord):
         # Covers
         self.findAndAddCover(editionRecord)
 
+        #host = 'digital-research-books-beta'\
+            #if os.environ['ENVIRONMENT'] == 'production' else 'drb-qa'
+
         # Acquisition Links
         for item in editionRecord.items:
             for link in item.links:
```
mwbenowitz, line 173: Let's move this first `addLink` call to an `else` block if we aren't presenting it in the reader. Otherwise for reader links we'll be putting that object in the OPDS object twice

api/opds2/publication.py
```diff
@@ -173,6 +176,14 @@ def parseEditionToPublication(self, editionRecord):
                     'type': link.media_type,
                     'rel': 'http://opds-spec.org/acquisition/open-access'
                 })
+                # Read Online Link
+                if 'reader' in link.flags:
+                    if link.flags['reader'] == True:
+                        self.addLink({
+                        'href': 'https://drb-qa.nypl.org/read/{}'.format(link.id),
```
mwbenowitz, line 183: You commented out the `host` line above but we need to keep it (or the current host into this method call) because we shouldn't hard-code the host here (should be QA or production).

mwbenowitz (commented):
Two minor changes but otherwise this looks good!

api/opds2/publication.py
mitri-slory: Sounds good to me.

api/opds2/publication.py
mitri-slory: I'll uncomment the host line. I commented it out in the first place due to wanting to test specifically when the url was 'drb-qa' and I was getting environment errors with the host variable originally.

mwbenowitz (approved):
🎉

PR #208, opened 2022-12-02
SFR-1503: Added format filter for work and edition pages

mitri-slory:
In this PR, I updated the work and edition blueprints to extract the filter parameter and the format values of not just requestable, but also readable and downloadable formats due to `formatEditionOutput` already doing the work of filtering out links that don't match the given formats. Also, I updated the unit tests for APIUtils and the work and edition blueprints as well. This work (`096911fb-3875-4763-adaf-de97fcafc526`) is an example of a readable and downloadable work to test the format filter out. This work (`1f944742-2789-4cbc-98bd-a74505645a37`) is an example of a requestable work with a readable edition so this is another good test example.

### DIFF
--- a/api/utils.py
+++ b/api/utils.py
@@ -121,7 +121,8 @@ class APIUtils():
     def formatWorkOutput(
         cls, works, identifiers, showAll=True, formats=None, reader=None
     ):
-        if isinstance(works, list):
+        #Multiple formatted works with formats specified
+        if isinstance(works, list) and identifiers != None:
             outWorks = []
             workDict = {str(work.uuid): work for work in works}
 
@@ -144,6 +145,19 @@ class APIUtils():
                 outWorks.append(outWork)
 
             return outWorks
+        #Formatted work with a specific format given
+        elif formats != None and identifiers == None:
+            formattedWork = cls.formatWork(
+                works, None, showAll, formats, reader=reader
+            )
+
+            formattedWork['editions'].sort(
+                key=lambda x: x['publication_date']
+                if x['publication_date'] else 9999
+            )
+
+            return formattedWork
+        #Formatted work with no format specified
         else:
             formattedWork = cls.formatWork(
                 works, None, showAll, reader=reader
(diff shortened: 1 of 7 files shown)


api/utils.py
```diff
@@ -121,7 +121,8 @@ def formatPagingOptions(page, pageSize, totalHits):
     def formatWorkOutput(
         cls, works, identifiers, showAll=True, formats=None, reader=None
     ):
-        if isinstance(works, list):
+        #Multiple formatted works with formats specified
+        if isinstance(works, list) and identifiers != None:
```
mwbenowitz, line 125: I'm not sure the `identifiers != None` check necessary here

api/utils.py
```diff
@@ -144,6 +145,19 @@ def formatWorkOutput(
                 outWorks.append(outWork)
 
             return outWorks
+        #Formatted work with a specific format given
+        elif formats != None and identifiers == None:
+            formattedWork = cls.formatWork(
+                works, None, showAll, formats, reader=reader
```
mwbenowitz, line 151: I think making this `formats=formats` is more consistent with how this is invoked above (though there is no reason it doesn't work!)

mwbenowitz (approved):
This looks great and does a really good job of extending the logic we already had in the codebase to implement this new feature.

There are two things I would consider doing from a maintenance perspective:
1. Refactor the list comprehension that returns `filteredFormats`. That's now used in 3 places in the repo and that's the point that I think it starts to make sure we aren't repeating ourselves
2. I think the conditionals in `formatWorkOutput` could be streamlined now by moving some of the logic to doing some sanity checking around the arguments rather than controlling the overall flow. This is a somewhat larger refactor though.

api/utils.py
mitri-slory: Ok. I'll change the format parameter to match that then.

api/utils.py
```diff
@@ -100,6 +100,16 @@ def formatAggregationResult(cls, aggregations, parentKey=None):
...
         return aggs
+        
+    @staticmethod
+    def formatFilters(terms):
+        formats = [
+        mediaType for f in list(filter(
+            lambda x: x[0] == 'format', terms['filter']
+         ))
+         for mediaType in APIUtils.FORMAT_CROSSWALK[f[1]]
+     ]
```
mwbenowitz, line 111: under indented here!

mwbenowitz (approved):
Looks great!!

PR #210, opened 2022-12-15
SFR-1578: Updated identifier search for escaped characters

mitri-slory:
In this PR, I replaced the escapedQuery in the `identifierQuery` and `authorityQuery` method with the original query, so special characters aren't escaped when searching for specific identifiers like DOAB or DOI. Also, I removed the backslash when splitting the `authIdent` list by the pipe character since the query is no longer being escaped.

### DIFF
--- a/api/elastic.py
+++ b/api/elastic.py
@@ -239,7 +239,6 @@ class ElasticClient():
 
         #User will type in authority|identifier in the url
         authIdentList = authIdentText.split('|')
-        print(authIdentList)
 
         #When user only types in the identifier in the search bar
         if len(authIdentList) < 2:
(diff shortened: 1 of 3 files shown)


api/elastic.py
```diff
@@ -238,8 +238,8 @@ def identifierQuery(self, authorityList, authIdentText):
                                     'identifiers.authority', 'editions.identifiers.authority'])
 
         #User will type in authority|identifier in the url
-        #Need backslash to avoid the pipe character from being escaped due to the escapeSearchQuery method
-        authIdentList = authIdentText.split('\|')
+        authIdentList = authIdentText.split('|')
+        print(authIdentList)
```
mwbenowitz, line 242: nit: dangling `print` here!

mwbenowitz (approved):
Looks great! Honestly, this solution is way easier and simpler than what I had in mind!

api/elastic.py
mitri-slory: Thanks for catching that.

PR #212, opened 2023-01-19
SFR-1590_inCollection Property in Work and Edition API Responses

mitri-slory:
In this PR, I updated APIUtils with a new checkEditionInCollection method for the work and edition API responses which adds  the metadata objects of the collections these works/editions are part of within an array. I also adjusted the tests and fixed a small typo in the swagger document. Some feedback on simplifying the queries or conditionals would be helpful.

### DIFF
--- a/api/utils.py
+++ b/api/utils.py
@@ -3,12 +3,9 @@ from datetime import datetime
 from hashlib import scrypt
 from flask import jsonify
 from math import ceil
-from managers import DBManager
 from model import Collection, Edition
 import re
-import os
 from model.postgres.collection import COLLECTION_EDITIONS
-from .opds2 import Feed
 
 class APIUtils():
     QUERY_TERMS = [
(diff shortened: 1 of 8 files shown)


api/utils.py
```diff
@@ -218,23 +226,72 @@ def addWorkMeta(cls, work, **kwargs):
...
+                .filter(Edition.work_id == work.id):
+                    for collectEdit in dbClient.session.query(COLLECTION_EDITIONS) \
+                        .filter(COLLECTION_EDITIONS.c.edition_id == edit.id):
+                            #Avoiding duplicate collection ids
+                            if collectEdit.collection_id not in collectionIDs:
+                                for collection in dbClient.session.query(Collection) \
+                                .filter(Collection.id == collectEdit.collection_id):
+                                    numOfItems = 0
+                                    for edit in collection.editions:
+                                        numOfItems += 1
```
mwbenowitz, line 255: Could just be `len(collection.editions)` I think!

api/utils.py
```diff
@@ -218,23 +226,72 @@ def addWorkMeta(cls, work, **kwargs):
...
+    @classmethod
+    def checkEditionInCollection(cls, work, edition, dbClient):
+
+        collectionIDs = []
+
+        if work != None:
+            for edit in dbClient.session.query(Edition) \
+                .filter(Edition.work_id == work.id):
+                    for collectEdit in dbClient.session.query(COLLECTION_EDITIONS) \
+                        .filter(COLLECTION_EDITIONS.c.edition_id == edit.id):
```
mwbenowitz, line 248: I'm a little concerned about the depth of the nested `for` loops here since it could spawn many database queries that would slow down our response times. I'm fairly certain it should be able to turn this into a single query which would help with that, but let me know if there are limitations on that approach.

api/utils.py
```diff
@@ -3,8 +3,12 @@
 from hashlib import scrypt
 from flask import jsonify
 from math import ceil
+from managers import DBManager
```
mwbenowitz, line 6: I think this and `os` aren't needed anymore!

api/utils.py
```diff
@@ -218,23 +226,72 @@ def addWorkMeta(cls, work, **kwargs):
...
+                            metadataOBJ = {
+                                'id': collection.id,
+                                'title': collection.title,
+                                'creator': collection.creator,
+                                'description': collection.description,
+                                'numberOfItems': numOfItems
+                            }
+                            collectionIDs.append(metadataOBJ)
+
+        dbClient.closeConnection()
```
mwbenowitz, line 282: This may be a straggler, but I would remove it to avoid weird complications if the connection gets closed unexpectedly!

mwbenowitz (commented):
I've got some questions about the implementation, but the output looks good!

I think there may be some architectural decisions I've made here that we might want to re-think at some point. Needing to pass the client through so many methods isn't a great pattern, but I don't see any way around it right now.

api/utils.py
mitri-slory: That one line is equivalent to the for loop conditional I wrote so I'll replace it with your suggestion.

api/utils.py
mitri-slory: Agreed. Running the API wasn't working with the closeConnection line still there.

api/utils.py
mitri-slory: I'll attempt to turn this into a simplified query.

mwbenowitz (approved):
👍

PR #213, opened 2023-01-23
feat: Simplify dev setup for newrelic

sarangj:
This shouldn't be a common issue, since testing new relic changes
locally should be pretty rare, but I found this a tad confusing. Instead
of the conditional and commented out variables, I think we can just
point the dev at the appropriate configs. The other upside to this
approach is that with `monitor_enabled` set to false and with an empty
license key, the newrelic agent gives us nice logs without sending data
up:
```
drb_local_webapp    | 2023-01-23 17:53:08,786 (41/MainThread) newrelic.api.transaction DEBUG - record_log_event has been called but no transaction or application was running. As a result, the following event has not been recorded. message: 'Starting API...' level: 'INFO' timestamp 1674496388785. To correct this problem, supply an application object as a parameter to this record_log_event call.
drb_local_webapp    | {"timestamp":1674496388785,"message":"Starting API...","log.level":"INFO","logger.name":"processes.api","thread.id":139994393364288,"thread.name":"MainThread","process.id":41,"process.name":"MainProcess","file.name":"/src/processes/api.py","line.number":24,"entity.type":"SERVICE"}
```
Maybe noisy, but also perhaps useful?

### DIFF
--- a/main.py
+++ b/main.py
@@ -73,7 +73,7 @@ def createArgParser():
     parser.add_argument('-r', '--singleRecord',
                         help='Single record ID for ingesting an individual record (only applicable for DOAB)')
     parser.add_argument('options', nargs='*', help='Additional arguments')
-    
+
     return parser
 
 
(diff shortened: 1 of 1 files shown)


sarangj:
Looks like `developer_mode` is another option: https://discuss.newrelic.com/t/disable-new-relic-for-local-development/67199/8?u=simplyedev1

sarangj:
i'll update the changelog 👀

PR #216, opened 2023-01-26
Register background task apps for new relic

sarangj:
I *believe* this is the missing piece for covering our non web processes for full instrumentation.  Basically, we just need to register an app and pass that to the BackgroundTask context manager to get the newrelic agent to fully instrument and forward logs from these tasks.

To test this, I performed the Docker based newrelic agent installation on my machine, ran the ClassifyProcess locally, and observed logs and metrics in the 'DRB ETL Pipeline (DEV)' app in newrelic APM.

Using this, looks like we don't need to individually decorate methods being called in the task.  I only removed the decorators from the ClassifyProcess for now so that we can first verify this in prod.

https://jira.nypl.org/browse/SFR-1603

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Updated .gitignore to ignore vim generated `.swp` files and `ctags` file
 ### Fixed
 - drb_local_devSetUp Docker container exiting out due to ES authentication error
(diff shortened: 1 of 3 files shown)


mitri-slory (commented):
The changes you made seem to make sense to me, but having Mike's input and feedback would be best since he has more experience with New Relic then me.

PR #217, opened 2023-01-27
fix: Pass DBClient to work formatter in search

sarangj:
Looks like we weren't passing the client object, but failing with a default of `None`. Since the dbClient appears to be required for this to run, remove the default and pass it in.  Might make sense to require keyword args on some of these methods at some point.


mwbenowitz (approved):
Looks good to me!

Just FYI: You can tag a PR with `no-changelog` to skip that action for quick/hot fixes in the future.

PR #219, opened 2023-01-31
Hotfix-Updated ES configs in production.yaml

mitri-slory:
The production.yaml ES configs were overwriting the configs set in the CloudFormation template, so now they are commented out with only the timeout being defined.


sarangj (approved):
lgtm!

PR #220, opened 2023-01-31
Remove remaining background_task decorators

sarangj:
Verified the new intrumentation approach works in QA, so we don't need to decorate all these methods anymore

https://jira.nypl.org/browse/SFR-1603


mitri-slory (approved):
Good to see that the new instrumentation approach is working!

PR #221, opened 2023-02-01
fix: Bump python version for CI tests

sarangj:
Noticed we're still running 3.8 for our test workflows while running 3.9 in prod


sarangj:
ah interesting, got it! Just thought I'd throw this up here, but interested to know the reasoning!

mwbenowitz (approved):
I don't believe that there was a good reason for these 2 things to be on different versions. I'd have to dig through the commit history but I believe this probably diverged a while ago and since it hasn't caused any issues I didn't notice. But very good to have them aligned!

PR #222, opened 2023-02-02
SFR-1582:Added new collectionUpdate endpoint

mitri-slory:
In this PR, I have created a new endpoint called `collectionUpdate` that for right now is focused on having engineers on the backend able to replace a collection that's currently in the database with a new collection that the user creates. I'm currently working on tests and updating the swagger file to reflect these changes.

### DIFF
--- a/api/blueprints/drbCollection.py
+++ b/api/blueprints/drbCollection.py
@@ -10,6 +10,9 @@ from ..opdsUtils import OPDSUtils
 from ..utils import APIUtils
 from ..opds2 import Feed, Publication
 from logger import createLog
+from model import Work
+from model.postgres.collection import COLLECTION_EDITIONS
+from datetime import date
 
 logger = createLog(__name__)
 
@@ -97,7 +100,7 @@ def collectionUpdate(uuid, user=None):
     collectionData = request.json
     dataKeys = collectionData.keys()
 
-    if {'title', 'creator', 'description'}.issubset(set(dataKeys))\
+    if {'title', 'creator', 'description'}.issubset(set(dataKeys)) == False\
             or len(set(dataKeys) & set(['workUUIDs', 'editionIDs'])) == 0:
         errMsg = {
             'message':
@@ -117,17 +120,38 @@ def collectionUpdate(uuid, user=None):
         errMsg = {'message': 'Unable to locate collection {}'.format(uuid)}
         return APIUtils.formatResponseObject(404, 'fetchSingleCollection', errMsg)
 
-    #Creating the new collection that will replace the old collection
-    updatedCollection = dbClient.createCollection(
-        collectionData['title'],
-        collectionData['creator'],
-        collectionData['description'],
-        user,
-        workUUIDs=collectionData.get('workUUIDs', []),
-        editionIDs=collectionData.get('editionIDs', [])
-    )
+    workUUIDs=collectionData.get('workUUIDs', [])
+    editionIDs=collectionData.get('editionIDs', [])
 
-    collection = updatedCollection
+    collection.title = collectionData['title']
+    collection.creator = collectionData['creator']
+    collection.description = collectionData['description']
+
+    #Deleting the rows of collection_editions that were in the original collection
+    dbClient.session.execute(COLLECTION_EDITIONS.delete().where(COLLECTION_EDITIONS.c.collection_id == collection.id))
+
+    #Inserting rows of collection_editions based on workUUIDs/editionIDs arrays
+    if len(editionIDs) > 0:
+        for editID in editionIDs:
+            dbClient.session.execute(COLLECTION_EDITIONS.insert().values \
+                (collection_id = collection.id, edition_id = editID))
+
+    if len(workUUIDs) > 0:
+        collectionWorks = dbClient.session.query(Work)\
+            .join(Work.editions)\
+            .filter(Work.uuid.in_(workUUIDs))\
+            .all()
+
+        for work in collectionWorks:
+            editions = list(sorted(
+                    [ed for ed in work.editions],
+                    key=lambda x: x.publication_date
+                    if x.publication_date else date.today()
+            ))
+        
+            for edition in editions:
+                dbClient.session.execute(COLLECTION_EDITIONS.insert().values \
+                    (collection_id = collection.id, edition_id = edition.id))
 
     dbClient.session.commit()
 
@@ -136,7 +160,6 @@ def collectionUpdate(uuid, user=None):
     opdsFeed = constructOPDSFeed(collection.uuid, dbClient)
 
     return APIUtils.formatOPDS2Object(201, opdsFeed)
-
 
 @collection.route('/<uuid>', methods=['GET'])
 def collectionFetch(uuid):
--- a/api/blueprints/drbCollection.py
+++ b/api/blueprints/drbCollection.py
@@ -92,12 +92,12 @@ def collectionCreate(user=None):
 @collection.route('/update/<uuid>', methods=['POST'])
 @validateToken
 def collectionUpdate(uuid, user=None):
-    logger.info('Updating current collection')
+    logger.info('Handling collection update request')
 
     collectionData = request.json
     dataKeys = collectionData.keys()
 
-    if len(set(dataKeys) & set(['title', 'creator', 'description'])) < 3\
+    if {'title', 'creator', 'description'}.issubset(set(dataKeys))\
             or len(set(dataKeys) & set(['workUUIDs', 'editionIDs'])) == 0:
         errMsg = {
             'message':
@@ -111,7 +111,11 @@ def collectionUpdate(uuid, user=None):
     dbClient.createSession()
 
     #Getting the collection the user wants to replace
-    collection = dbClient.fetchSingleCollection(uuid)
+    try:
+        collection = dbClient.fetchSingleCollection(uuid)
+    except NoResultFound:
+        errMsg = {'message': 'Unable to locate collection {}'.format(uuid)}
+        return APIUtils.formatResponseObject(404, 'fetchSingleCollection', errMsg)
 
     #Creating the new collection that will replace the old collection
     updatedCollection = dbClient.createCollection(
--- a/api/blueprints/drbCollection.py
+++ b/api/blueprints/drbCollection.py
@@ -127,35 +127,17 @@ def collectionUpdate(uuid, user=None):
     collection.creator = collectionData['creator']
     collection.description = collectionData['description']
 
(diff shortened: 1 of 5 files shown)


api/blueprints/drbCollection.py
```diff
@@ -89,6 +89,50 @@ def collectionCreate(user=None):
...
+                ', with one of workUUIDs or editionIDs to create a collection'
+        }
+
+        return APIUtils.formatResponseObject(400, 'createCollection', errMsg)
+
+    dbClient = DBClient(current_app.config['DB_CLIENT'])
+    dbClient.createSession()
+
+    #Getting the collection the user wants to replace
+    collection = dbClient.fetchSingleCollection(uuid)
```
sarangj, line 114: If a non-existent uuid is passed and we get a `sqlalchemy.orm.exc.NoResultFound` (I *think* that's what we'd expect), should we catch and return a 404? Maybe not too high pri since this is only expected to be called by an engineer running a script...

api/blueprints/drbCollection.py
```diff
@@ -89,6 +89,50 @@ def collectionCreate(user=None):
 
     return APIUtils.formatOPDS2Object(201, opdsFeed)
 
+@collection.route('/update/<uuid>', methods=['POST'])
+@validateToken
+def collectionUpdate(uuid, user=None):
+    logger.info('Updating current collection')
```
sarangj, line 95: Maybe we should move this down to after we validate the input? Or change the message to something like "Handling collection update request" since this message might be misleading if the validation below fails?

api/blueprints/drbCollection.py
```diff
@@ -89,6 +89,50 @@ def collectionCreate(user=None):
...
+
+    collectionData = request.json
+    dataKeys = collectionData.keys()
+
+    if len(set(dataKeys) & set(['title', 'creator', 'description'])) < 3\
```
sarangj, line 100: Maybe just a stylistic preference, but you could also use `issubset`: 
```
if {'title', 'creator', 'description'}.issubset(dataKeys):
```

I *think* that does the same thing, but up to you!

api/blueprints/drbCollection.py
```diff
@@ -89,6 +89,50 @@ def collectionCreate(user=None):
...
+        return APIUtils.formatResponseObject(400, 'createCollection', errMsg)
+
+    dbClient = DBClient(current_app.config['DB_CLIENT'])
+    dbClient.createSession()
+
+    #Getting the collection the user wants to replace
+    collection = dbClient.fetchSingleCollection(uuid)
+
+    #Creating the new collection that will replace the old collection
+    updatedCollection = dbClient.createCollection(
```
sarangj, line 117: hmmm, do we need to add a new method to update the collection with the given uuid instead of creating a new one?

sarangj (commented):
couple questions / suggestions!

api/blueprints/drbCollection.py
mitri-slory: That sounds like a good idea. It's similar to lines 104-110 in the collectionFetch endpoint. 

```
try:
        opdsFeed = constructOPDSFeed(
            uuid, dbClient, sort=sort, page=page, perPage=perPage
        )
    except NoResultFound:
        errMsg = {'message': 'Unable to locate collection {}'.format(uuid)}
        return APIUtils.formatResponseObject(404, 'fetchCollection', errMsg)
```

api/blueprints/drbCollection.py
mitri-slory: Good suggestion. I'll change the message to reflect what's really going on.

api/blueprints/drbCollection.py
mitri-slory: That is a cleaner way of saying the same line. I'll implement it with this small change to dataKeys. 

`if {'title', 'creator', 'description'}.issubset(set(dataKeys))`

api/blueprints/drbCollection.py
mitri-slory: We're doing this for now since the base requirements for the ticket is to allow this endpoint to replace all the data of a current collection with a new collection. If we delete the old collection and create a new one, than the id of the new one will be different in the database instead of the same as the old collection. We want to retain the id of the original collection.

Later on, I plan to implement adding/deleting works and editions from a collection instead of simply replacing the whole collection object.

api/blueprints/drbCollection.py
```diff
@@ -89,6 +92,74 @@ def collectionCreate(user=None):
...
+    collection.description = collectionData['description']
+
+    #Deleting the rows of collection_editions that were in the original collection
+    dbClient.session.execute(COLLECTION_EDITIONS.delete().where(COLLECTION_EDITIONS.c.collection_id == collection.id))
+
+    #Inserting rows of collection_editions based on workUUIDs/editionIDs arrays
+    if len(editionIDs) > 0:
+        for editID in editionIDs:
+            dbClient.session.execute(COLLECTION_EDITIONS.insert().values \
+                (collection_id = collection.id, edition_id = editID))
```
sarangj, line 137: generally more pythonic to do `if editionIDs`, but no big deal.

Other than that, performance wise, it would be better to do one insert instead of separate inserts in a loop, ie:
```
dbClient.session.execute(COLLECTION_EDITIONS.insert().values([
    {"collection_id": collection.id, "edition_id": eid}
    for eid in editionIDs
])

File: api/blueprints/drbCollection.py [hunk 6]
```diff
@@ -89,6 +92,74 @@ def collectionCreate(user=None):
...
+                    [ed for ed in work.editions],
+                    key=lambda x: x.publication_date
+                    if x.publication_date else date.today()
+            ))
+        
+            for edition in editions:
+                dbClient.session.execute(COLLECTION_EDITIONS.insert().values \
+                    (collection_id = collection.id, edition_id = edition.id))
+
+    dbClient.session.commit()
```
sarangj, line 156: it might be good to separate some of this logic into methods on the `dbClient`, maybe 3 methods like `removeEditionsFromCollection`, `addEditionsToCollection` and `addWorkEditionsToCollection`?

api/blueprints/drbCollection.py
```diff
@@ -89,6 +92,74 @@ def collectionCreate(user=None):
...
+        for work in collectionWorks:
+            editions = list(sorted(
+                    [ed for ed in work.editions],
+                    key=lambda x: x.publication_date
+                    if x.publication_date else date.today()
+            ))
+        
+            for edition in editions:
+                dbClient.session.execute(COLLECTION_EDITIONS.insert().values \
+                    (collection_id = collection.id, edition_id = edition.id))
```
sarangj, line 154: similar here, once you get all the edition ids from the works, you should just be able to insert in one go, instead of inserting one at a time in a loop. Also, is there a need to sort them?  It shouldn't actually affect how they're stored in the DB, but lmk if I'm missing something!

sarangj (commented):
Nice, glad ya got the delete / insert logic working! Data flow, looks good to me, just a couple thoughts on separating out some of the logic for querying the db into the dbClient, but feel free to push back if you don't think it makes sense!

api/blueprints/drbCollection.py
mitri-slory: Nice advice on cleaning up the code and making it more efficient.

api/blueprints/drbCollection.py
mitri-slory: I agree. I was planning on deduplicating my code into separate methods today as well.

api/blueprints/drbCollection.py
mitri-slory: You're right on the sorted method. I got lines 146-149 from another method and didn't realize that the sorted method isn't needed for this function in particular.

mitri-slory:
@mwbenowitz Could you take a look at the swagger file update I made to see if the new endpoint will show up properly on the apidocs site when this branch is merged into main.

api/blueprints/drbCollection.py
```diff
@@ -252,3 +305,30 @@ def constructOPDSFeed(
...
+            .filter(Work.uuid.in_(workUUIDs))\
+            .all()
+
+    for work in collectionWorks:
+        editions = [ed for ed in work.editions]
+        
+
+        dbClient.session.execute(COLLECTION_EDITIONS.insert().values([ \
+        {"collection_id": collection.id, "edition_id": edition.id} \
+        for edition in editions \
```
mwbenowitz, line 333: I think we only want to insert one "preferred" edition into the collection per work. @sarangj I think you just implemented something similar for your most recent PR, would a similar approach work here?

swagger.v4.json
```diff
@@ -770,6 +770,85 @@
                 }
             }
         },
+        "/collection/update/{uuid}": {
+            "get": {
```
mwbenowitz, line 774: Should be `post` to align with the endpoint!

mwbenowitz (commented):
Looks good, just two comments! I think the OpenAPI spec here looks good except for the one minor note I left.

api/blueprints/drbCollection.py
sarangj: Yea, mine was kind of a gnarly SQL based solution: https://github.com/NYPL/drb-etl-pipeline/pull/225/files#diff-bb24eb4550556613d258592571b4d75aba40f31853c1b3fd384b886929b1d09eR59-R113

Here, we're not fetching as much data, so I think you could just sort the list of editions and pick the earliest one per collection?

sarangj: so like, i think if per work, you sort the list of editions by date create ascending and just use the first edition, that should do it.

api/blueprints/drbCollection.py
```diff
@@ -252,3 +304,31 @@ def constructOPDSFeed(
...
+    collectionWorks = dbClient.session.query(Work)\
+            .join(Work.editions)\
+            .filter(Work.uuid.in_(workUUIDs))\
+            .all()
+
+    for work in collectionWorks:
+        collectionEditions = dbClient.session.query(Edition)\
+            .filter(Edition.work_id == work.id)\
+            .order_by(Edition.date_created.asc())\
+            .all()
```
sarangj, line 330: Nice, this looks like the right logic here!  You *could* do a `.limit(1)` and then `.scalar()` instead of `.all()`, but no big deal really, just slighty less data sent over.

sarangj (approved):
Nice, I think if we have perf issues, we might look into refining the query in a loop bit, but I think it's fine for this, shouldn't be much data from what I understand!

api/blueprints/drbCollection.py
mitri-slory: I'll apply this logic and then I'll merge the change into main.

PR #224, opened 2023-02-08
Add 'automatic' collections DB entities

sarangj:
To support defining 'automatic' collections, define a new 'automatic collection definition' defining search terms to define an automatic collection. These terms will essentially mirror what we already use to hit ES when searching for works. We'll probably special case the example of the 'most recent' collection to just hit the DB, but otherwise we'll dynamically load the search terms into an ES query.

Otherwise, these automatic collections will simply be collections with a new collection 'type' of 'automatic' (defaulting all existing collections to 'static'), so we can reuse all the metadata fields on the existing collection.

Some specific things I have questions / would like feedback around:

  - Reusing the collection table for storing collection metadata (as opposed to creating a new table with separate PKs)
  - naming
  - should I add more constraints at the DB level? either way, I'd like to iron out the full implementation first before adding those.

### DIFF
--- a/migrations/versions/e46b30dd3ff5_automatic_collections.py
+++ b/migrations/versions/e46b30dd3ff5_automatic_collections.py
@@ -9,7 +9,6 @@ from alembic import op
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql
 
-
 # revision identifiers, used by Alembic.
 revision = 'e46b30dd3ff5'
 down_revision = '9e33fa16ba82'
@@ -18,10 +17,8 @@ depends_on = None
 
 
 def upgrade():
-    # TODO: Maybe add some CHECK constraints around some of these fields? Or just
-    # enforce in code upon creation.
     op.create_table(
-        "automatic_collection_definition",
+        "automatic_collection",
         sa.Column(
             'collection_id',
             sa.Integer,
@@ -29,16 +26,30 @@ def upgrade():
             primary_key=True,
         ),
         sa.Column(
-            'query',
-            sa.ARRAY(sa.String),
+            'keyword_query',
+            sa.String,
         ),
         sa.Column(
-            'sort',
-            sa.ARRAY(sa.String),
+            'author_query',
+            sa.String,
         ),
         sa.Column(
-            'filter',
-            sa.ARRAY(sa.String),
+            'title_query',
+            sa.String,
+        ),
+        sa.Column(
+            'subject_query',
+            sa.String,
+        ),
+        sa.Column(
+            'sort_field',
+            sa.String,
+            sa.CheckConstraint(r"sort_field is NULL OR sort_field IN ('title', 'author', 'date')"),
+        ),
+        sa.Column(
+            'sort_direction',
+            sa.String,
+            sa.CheckConstraint(r"sort_direction is NULL OR sort_direction IN ('ASC', 'DESC')"),
         ),
         sa.Column(
             'limit',
@@ -60,6 +71,6 @@ def upgrade():
 
 def downgrade():
     op.drop_column("collections", "type")
-    op.drop_table("automatic_collection_definition")
+    op.drop_table("automatic_collection")
     collection_type = postgresql.ENUM('static', 'automatic', name="collection_type")
     collection_type.drop(op.get_bind())
--- a/migrations/versions/e46b30dd3ff5_automatic_collections.py
+++ b/migrations/versions/e46b30dd3ff5_automatic_collections.py
@@ -1,0 +1,65 @@
+"""automatic_collections
+
+Revision ID: e46b30dd3ff5
+Revises: 9e33fa16ba82
+Create Date: 2023-02-06 11:44:29.000880
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+# revision identifiers, used by Alembic.
+revision = 'e46b30dd3ff5'
+down_revision = '9e33fa16ba82'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # TODO: Maybe add some CHECK constraints around some of these fields? Or just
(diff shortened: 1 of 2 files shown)


migrations/versions/e46b30dd3ff5_automatic_collections.py
```diff
@@ -0,0 +1,65 @@
...
+down_revision = '9e33fa16ba82'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # TODO: Maybe add some CHECK constraints around some of these fields? Or just
+    # enforce in code upon creation.
+    op.create_table(
+        "automatic_collection_definition",
```
mitri-slory, line 24: I suggest changing the name of this table to "automatated_collections" just to shorten it and make it clearer.

migrations/versions/e46b30dd3ff5_automatic_collections.py
```diff
@@ -0,0 +1,65 @@
...
+        ),
+        sa.Column(
+            'filter',
+            sa.ARRAY(sa.String),
+        ),
+        sa.Column(
+            'limit',
+            sa.Integer,
+            nullable=False,
+        ),
```
mitri-slory, line 47: I think that at the very least these columns should have CHECK constraints to specify the filters that are allowed and a defined limit to the number of works/editions that can be in an automated collection.

mitri-slory (commented):
Overall, this is definitely a good start to adding this data model for the automated collections. I have a few comments below addressing your specific questions. Also, I agree that having the new automated collections table be linked to the current collections table makes the most sense since automated collections are still collections and should be a part of the original table as well.

migrations/versions/e46b30dd3ff5_automatic_collections.py
sarangj: Yea, I actually think I'm going to move these array columns to JSONB anyway, and I can implement check constraints on that.

For the `limit` clause, I'm not as concerned about limiting that, but lemme think about that. Or maybe I could just err on the side of a really high default, but I think implementing pagination properly should avoid us overwhelming ES.

sarangj:
After looking this over more, thinking more about @mitri-slory 's feedback and researching strategies for validating arrays and jsonb columns in postgrs, I'm gonna make two pretty significant changes:

  1. Punt on the `filter` portion of this, at least at the outset
  2. Ditch the array / jsonb idea and just use individual flat columns for each field that can be queried for simpler type safety

migrations/versions/e46b30dd3ff5_automatic_collections.py
```diff
@@ -18,27 +17,39 @@
...
+            'author_query',
+            sa.String,
+        ),
+        sa.Column(
+            'title_query',
+            sa.String,
+        ),
+        sa.Column(
+            'subject_query',
+            sa.String,
```
mitri-slory, line 42: This is a good idea to make the type of queries the user puts in very clear for us backend engineers.

mitri-slory (approved):
This looks good to me. Having @mwbenowitz advice on this migration change would be good to have as well since I've done only one migration myself on this codebase so far.

migrations/versions/e46b30dd3ff5_automatic_collections.py
```diff
@@ -0,0 +1,76 @@
...
+        sa.Column(
+            'sort_direction',
+            sa.String,
+            sa.CheckConstraint(r"sort_direction is NULL OR sort_direction IN ('ASC', 'DESC')"),
+        ),
+        sa.Column(
+            'limit',
+            sa.Integer,
+            nullable=False,
+        ),
```
mwbenowitz, line 58: Should `limit` be required? I see the need for some kind of functionality here but what if we wanted to create a collection around an author and as our catalog grows it exceeds this limit? I could see an argument for a max collection size being defined somewhere universally, but this might create some unexpected behavior?

migrations/versions/e46b30dd3ff5_automatic_collections.py
```diff
@@ -0,0 +1,76 @@
...
+        sa.Column(
+            'sort_field',
+            sa.String,
+            sa.CheckConstraint(r"sort_field is NULL OR sort_field IN ('title', 'author', 'date')"),
+        ),
+        sa.Column(
+            'sort_direction',
+            sa.String,
+            sa.CheckConstraint(r"sort_direction is NULL OR sort_direction IN ('ASC', 'DESC')"),
+        ),
```
mwbenowitz, line 53: Should these be required with a default value of something like `title ASC` (or even `uuid ASC`)? Without a sort value we might see ordering change in these queries? Providing default behavior might ensure more consistency for patrons looking at these collections.

mwbenowitz (approved):
This looks like a solid migration! Just a couple of questions about functionality. Which might be best addressed by product?

migrations/versions/e46b30dd3ff5_automatic_collections.py
sarangj: hmm good point, perhaps we could require *either* one of the `_query` fields *or* (inclusive) the `limit` field. 

Though now I wonder if there's any utility to having a limit at all if I add a bit of logic to https://github.com/NYPL/drb-etl-pipeline/pull/225 to implement pagination in the DB query (and elastic query for search based stuff).  I guess a question for Sean that I can reach out about.

migrations/versions/e46b30dd3ff5_automatic_collections.py
sarangj: Ah yea, maybe uuid would make sense, as it doesn't implement a semantically meaningful sort, but still keeps the sort stable

migrations/versions/e46b30dd3ff5_automatic_collections.py
sarangj: I think I'll just make `limit` optional and do the pagination thing I mentioned in the next PR.

PR #225, opened 2023-02-10
Implement non search based automatic collection 

sarangj:

As a first step, add the new models and implement the logic to fetch
editions for any collections not requiring text search. In these cases,
such as with the main example of a 'Most Recent' collection, we can just
hit the DB. This also includes a script to generate the 'Most Recent'
collection.

In a later diff, I'll implement handling of all the text search fields
that we are currenly ignoring.

Note, this branch is based on the migration branch https://github.com/NYPL/drb-etl-pipeline/pull/224 so this PR is currently against the branch on that PR. Once the migration goes in, I'll reset the base on this one to main before merging!

### DIFF
--- a/api/automaticCollectionUtils.py
+++ b/api/automaticCollectionUtils.py
@@ -3,21 +3,23 @@ from logger import createLog
 logger = createLog(__name__)
 
 
-def fetchAutomaticCollectionEditions(dbClient, collectionId):
+def fetchAutomaticCollectionEditions(dbClient, collectionId, page: int, perPage: int):
     """Given a collection id for an automatic collection, perform the given
     search and return a list of collection editions
     """
     automaticCollection = dbClient.fetchAutomaticCollection(collectionId)
     if not automaticCollection:
         logger.warning("Found invalid automatic collection %s with no search definition", collectionId)
-        return []
+        return (0, [])
 
     # Skip ES and go straight to the DB if there's no actual search to be done
     if not _requiresSearch(automaticCollection):
-        return dbClient.fetchSortedEditions(
+        return dbClient.fetchAllPreferredEditions(
             sortField=automaticCollection.sort_field,
             sortDirection=automaticCollection.sort_direction,
             limit=automaticCollection.limit,
+            page=page,
+            perPage=perPage,
         )
 
     # TODO: Hit ES for query based collections
--- a/api/automaticCollectionUtils.py
+++ b/api/automaticCollectionUtils.py
@@ -1,0 +1,35 @@
+from logger import createLog
+
+logger = createLog(__name__)
+
+
+def fetchAutomaticCollectionEditions(dbClient, collectionId):
+    """Given a collection id for an automatic collection, perform the given
+    search and return a list of collection editions
+    """
+    automaticCollection = dbClient.fetchAutomaticCollection(collectionId)
+    if not automaticCollection:
+        logger.warning("Found invalid automatic collection %s with no search definition", collectionId)
+        return []
+
+    # Skip ES and go straight to the DB if there's no actual search to be done
+    if not _requiresSearch(automaticCollection):
+        return dbClient.fetchSortedEditions(
+            sortField=automaticCollection.sort_field,
+            sortDirection=automaticCollection.sort_direction,
+            limit=automaticCollection.limit,
+        )
+
+    # TODO: Hit ES for query based collections
+    raise ValueError("Automatic collection search is not yet implemented")
+
+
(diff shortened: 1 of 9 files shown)


api/automaticCollectionUtils.py
```diff
@@ -0,0 +1,35 @@
...
+    # Skip ES and go straight to the DB if there's no actual search to be done
+    if not _requiresSearch(automaticCollection):
+        return dbClient.fetchSortedEditions(
+            sortField=automaticCollection.sort_field,
+            sortDirection=automaticCollection.sort_direction,
+            limit=automaticCollection.limit,
+        )
+
+    # TODO: Hit ES for query based collections
+    raise ValueError("Automatic collection search is not yet implemented")
```
mitri-slory, line 24: Really like your comments here and the TODO one for reminding yourself to fetch query based automated collections when you further develop this branch.

scripts/addMostRecentAutomaticCollection.py
```diff
@@ -0,0 +1,82 @@
...
+                ) VALUES (:collectionId, :sortField, :sortDirection, :limit)
+                """,
+            ),
+            {
+                "collectionId": collection_id,
+                "sortField": "date",
+                "sortDirection": "DESC",
+                "limit": "100",
+            },
+        )
```
mitri-slory, line 78: After these execute statements don't forget to type `dbManager.session.commit` or these changes won't actually occur in the database.

mitri-slory: You only need to type the commit line one time at the end of this main method.

mitri-slory (commented):
Looking really good so far. Once we see how the migration goes, we might need to adjust some parts of this PR, but for right now it makes sense. I left a few comments below as well.

api/automaticCollectionUtils.py
sarangj: thanks!

scripts/addMostRecentAutomaticCollection.py
sarangj: actually, the `with session.begin()` context manager takes care of that! Once the with block is done, the context manager commits, or rolls back if there is an exception

mitri-slory: Oh that's nice to know. I just looked it up the session API and that's a good thing to remember.

api/db.py
```diff
@@ -56,6 +56,29 @@ def fetchSingleEdition(self, editionID, showAll=False):
             )\
             .filter(Edition.id == editionID).first()
 
+    def fetchSortedEditions(self, sortField: str, sortDirection: str, limit: int):
```
mwbenowitz, line 59: 👍 to the code, feels like the name is a bit more general than the function though. How would you feel about one of these options?
1. Update the function name to reflect how they're sorted.
2. Make this a bit more generic so that `sortField` could be any valid `Edition` field. I'm not 100% sure that would make sense (I like the sanity check here).

api/db.py
```diff
@@ -56,6 +56,29 @@ def fetchSingleEdition(self, editionID, showAll=False):
...
+            "title": Edition.title,
+            "date": Edition.date_created,
+        }.get(sortField)
+        if not sortClause:
+            raise ValueError(f"Invalid sort param {sortField}")
+        if sortDirection == "DESC":
+            sortClause = sortClause.desc()
+
+        return (
+            self.session.query(Edition)
```
mwbenowitz, line 70: I have one concern here, which is that this could/will return many editions for a single work in the collection (since they all get created together). I _think_ what we want is really to get the `Work` record and then send back what we would generally call the "preferred" edition. That is a bit more complicated but not too bad?

mwbenowitz (commented):
Again: solid implementation of the automatic collection idea, here just mainly a thought about what will get pulled into our first collection. I think its an edge case, so want to talk about how best to handle that!

api/db.py
sarangj: I just changed the function name, but maybe not in a way that really solves option 1. For option 2 (making this more generic), I'm not sure I want to implement the magic to allow arbitrary sort fields passed in, but I did add to the docstring a note to consider it, thoughts?

api/db.py
sarangj: Latest commit should address that using some window function magic!

mitri-slory (approved):
These changes look good. The addition of the pagination and extracting the 'limit' logic out of the DBClient makes sense to me.

sarangj:
got the migration up in QA, so this is safe to merge now!

PR #226, opened 2023-02-15
Add DEFAULT_COVER_URL to local-compose env

sarangj:
Had to set this variable to make some of the endpoints in my webapp docker container, seems good to make sure it works for everyone!


mitri-slory (approved):
I was having similar issues with a KeyError with the DEFAULT_COVER_URL variable when running the createCollection endpoint with the sample compose file and adding the variable like you suggested fixed the issue.

PR #229, opened 2023-02-17
Fix preferred editions query

sarangj:
See commit messages, first one does a basic filter (in lieu of a potential future per enhancement that's more robust) and the second one fixes some deduping issues (see the commit message for details).


mitri-slory (approved):
These changes are nice ways to deduplicate the code and the limit addition is a good performance solution as well. If the removal of the joinedload parameters doesn't affect fetching the editions we want, than we can proceed without them.

PR #230, opened 2023-02-22
SFR-1607: Added rights data object to Publication & Feed opds2 directory

mitri-slory:
For the front-end, the publication rights data block which appears in the `/edition/{id} ` and search API responses needs to also be present in the /collection/{uuid} API response. I chose to add a Rights module to the opds2 directory and then imported the module into the Feed and Publication modules to properly add the data block to the API response. I adjusted some of the tests as well due this addition and will add more in a future commit.


mwbenowitz (commented):
This looks really solid from a code perspective. I appreciate the approach of extending the current model and sticking to the established patterns.

My only question/concern is about where we're adding the rights documents to the responses. OPDS2 is a flexible standard, but this takes us further away from the standard implementation.

The option I have in mind is adding the rights data to the `Metadata` section, which is more in line with what already exists. The metadata is based on the Schema.org standard, which has provisions for rights.

Perhaps it would be good to sit down and talk this through as we have several other extensions in the works for OPDS2 and I'd be curious to figure out how this might fit in with that.

tests/unit/test_opds_publication.py
```diff
@@ -193,7 +182,8 @@ def test_parseEditionToPublication_ReaderFlagFalse(self, testPubEls, mocker):
             languages=[None, {'iso_3': 'tst'}, {'iso_3': 'oth'}],
             date_created='testCreated',
             date_modified='testModified',
-            items=[mocker.MagicMock(links=[mocker.MagicMock(id='testID', url='testURL', media_type='testType', flags={'reader': False})])],
+            items=[mocker.MagicMock(links=[mocker.MagicMock(id='testID', url='testURL', media_type='testType', flags={'reader': False})],
+                                    rights=[mocker.MagicMock(source='testSource', license='testLicense', rightsStatement='testStatement')])],
```
sarangj, line 186: it would be good to add an `assert` to test that `addField` was called for this right below where we do `metadata.addField.assert_has_calls`

sarangj (approved):
looks great to me! Just one non blocking comment regarding the existing unit test!

PR #231, opened 2023-02-24
SFR-1614_New collection endpoint to selectively update metadata

mitri-slory:
This endpoint will be tailored to have the user choose which metadata values to update through request parameters. In later commits, I will update the swagger file and unit tests in consideration of this new endpoint.

### DIFF
--- a/api/blueprints/drbCollection.py
+++ b/api/blueprints/drbCollection.py
@@ -165,6 +165,8 @@ def collectionUpdate(uuid):
     collection.creator = creator
     collection.description = description
 
+    dbClient.session.commit()
+
     try:
         opdsFeed = constructOPDSFeed(uuid, dbClient)
     except NoResultFound:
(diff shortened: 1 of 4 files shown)


api/blueprints/drbCollection.py
```diff
@@ -143,6 +143,38 @@ def collectionUpdate(uuid, user=None):
...
+    #Getting the collection the user wants to update
+    try:
+        collection = dbClient.fetchSingleCollection(uuid)
+    except NoResultFound:
+        errMsg = {'message': 'Unable to locate collection {}'.format(uuid)}
+        return APIUtils.formatResponseObject(404, 'fetchSingleCollection', errMsg)
+
+    collection.title = title
+    collection.creator = creator
+    collection.description = description
```
sarangj, line 166: one thing to consider: if the request doesn't contain one of these fields, we'll essentially be overwriting the given field to `None`. We may want to only update the fields that are present in the request object!

sarangj (commented):
logic looks great, just one consideration below!

sarangj (approved):
noice, looks good!

PR #232, opened 2023-02-24
HOTFIX-Added type parameter to dbClient.createCollection

mitri-slory:
Due to the recent database migration for automated collections, the createCollection endpoint returned a 500 integrity error due to the createCollection method in the dbClient not specifying a 'type' as a parameter which is now a column in the collections table. Now, I have defaulted the 'type' in the endpoint and dbClient method to be static at all times to resolve this issue.


mwbenowitz (approved):
👍

PR #233, opened 2023-02-27
Add logic for search based collection 

sarangj:
If we encounter an automatic collection with search parameters, build
up an elastic search query and return the results.

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -119,6 +119,37 @@ class DBClient():
             editionsQuery.offset(offset).limit(perPage).all(),
         )
 
+    def fetchEditions(self, editionIDs):
+        """Fetch the editions in the given list, respecting the order
+        of the passed in list
+        """
+
+        # First build a CTE of the passed in ids and their index in the
+        # given list
+        editionIdCTE = sa.select(
+            sa.values(
+                sa.column("idx", sa.Integer),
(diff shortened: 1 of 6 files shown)


api/db.py
```diff
@@ -119,6 +119,37 @@ def fetchAllPreferredEditions(
...
+
+        # First build a CTE of the passed in ids and their index in the
+        # given list
+        editionIdCTE = sa.select(
+            sa.values(
+                sa.column("idx", sa.Integer),
+                sa.column("edition_id", sa.Integer),
+                name="subquery",
+            ).data(list(enumerate(editionIDs)))
+        ).cte("edition_ids")
```
mitri-slory, line 135: I like your idea of using a CTE for building this query. I've never had experience using CTEs before, so your use of it is very interesting as a way to create more complex queries. With your use of it here, I might start using them for some queries I create in the future.

mitri-slory (approved):
Your logic for these automated search-based collections looks great to me. I'm not sure if we truly need the script you created to be merged into main, but @mwbenowitz might be a better judge on that one. The only other suggestion would be to update the Changelog before merging.

sarangj:
changelog updated!  Ya, I think I'll drop the script commit before merging!

PR #234, opened 2023-03-02
Support automatic collections in POST /collection 

sarangj:
https://jira.nypl.org/browse/SFR-1611

Accept some new data in the request body that indicates the user wants
to create an automatic collection. Do some validation and then do the
creation.

Note, I had to augment the sqlalchemy models to allow for creating both
entities in a single transaction.

### DIFF
--- a/api/blueprints/drbCollection.py
+++ b/api/blueprints/drbCollection.py
@@ -63,12 +63,9 @@ def collectionCreate(user=None):
     collectionData = request.json
     dataKeys = collectionData.keys()
 
-    if len(set(dataKeys) & set(['title', 'creator', 'description'])) < 3\
-            or len(set(dataKeys) & set(['workUUIDs', 'editionIDs'])) == 0:
+    if len(set(dataKeys) & set(['title', 'creator', 'description'])) < 3:
         errMsg = {
-            'message':
-                'title, creator and description fields are required'
-                ', with one of workUUIDs or editionIDs to create a collection'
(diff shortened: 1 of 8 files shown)


api/blueprints/drbCollection.py
```diff
@@ -63,28 +63,51 @@ def collectionCreate(user=None):
...
+        newCollection = dbClient.createAutomaticCollection(
+            collectionData['title'],
+            collectionData['creator'],
+            collectionData['description'],
+            owner=user,
+            sortField=autoDef.get("sortField"),
+            sortDirection=autoDef.get("sortDirection"),
+            limit=autoDef.get("limit"),
+            **{field: autoDef.get(field) for field in queryFields},
+        )
```
mitri-slory, line 101: What would happen if a user puts in the editionIDs/workUUIDs and 'autodef' in the request json body? Would the ids be prioritized or the 'autodef' parameter. I think this edge case should be considered.

mitri-slory (commented):
This is looking really good so far. I have one comment about a certain edge case, but once that's resolved the PR should be ready to be merged in.

api/blueprints/drbCollection.py
sarangj: Fair point! Right now, we'd just ignore the `autoDef` but it probably makes sense to raise an error if they do that, I can add that!

mitri-slory: Sounds good.

PR #235, opened 2023-03-02
SFR-1636: Make collection links identifiable

mitri-slory:
In this PR, I added an identifier value to the publication links for each edition to categorize the links as readable, downloadable, requestable, catalog, and other in the fetchCollection API response.

### DIFF
--- a/api/opds2/publication.py
+++ b/api/opds2/publication.py
@@ -185,7 +185,7 @@ class Publication:
                     })
                     # Non-readable links
                     else:
-                        identifier = self.setLinkIdentifier(link)
+                        identifier = self.getLinkIdentifier(link)
                         self.addLink({
                         'href': link.url,
                         'type': link.media_type,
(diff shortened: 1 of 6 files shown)


api/opds2/publication.py
```diff
@@ -242,6 +250,23 @@ def setPreferredLink(self, editions):
             'type': firstLink.media_type,
             'rel': 'http://opds-spec.org/acquisition/open-access'
         })
+    
+    def setLinkIdentifier(self, link):
```
sarangj, line 254: one nit on the name of this method: I'd normally expect a method starting with `set` to mutate `self` in some way, but this just returns a value. How would you feel about renaming this to `getLinkIdentifier` or `determineLinkIdentifier` (or anything that indicates that this isn't doing a mutation)?

sarangj (commented):
logic looks great, just one nit on naming!

api/opds2/publication.py
mitri-slory: Makes sense. I'll change the method to `getLinkidentifier`

PR #237, opened 2023-03-09
SFR-1633: Updated OCLC Classify API

mitri-slory:
With this PR, I updated the OCLC Classify API with an updated root url, adding a header to the query for the classification process, and updated the API keys for the qa and production config files.

### DIFF
--- a/managers/oclcClassify.py
+++ b/managers/oclcClassify.py
@@ -125,8 +125,8 @@ class ClassifyManager:
             [str] -- A string of XML data comprising of the body of the
             Classify response.
         """
-        classifyResp = requests.get(self.query, \
-            headers={'X-OCLC-API-Key': os.environ['OCLC_API_KEY']}, \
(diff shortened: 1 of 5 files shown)


managers/oclcClassify.py
```diff
@@ -124,7 +125,10 @@ def execQuery(self):
             [str] -- A string of XML data comprising of the body of the
             Classify response.
         """
-        classifyResp = requests.get(self.query, timeout=10)
+        classifyResp = requests.get(self.query, \
+            headers={'X-OCLC-API-Key': os.environ['OCLC_API_KEY']}, \
```
sarangj, line 129: also, you *shouldn't* need these line break characters within a function call (unless I'm missing something!)

sarangj (commented):
generally looks good, but we probably wanna change the environment variable name to avoid a collision!

PR #238, opened 2023-03-09
SFR-1615: Added new deleteWorkEditionCollection endpoint

mitri-slory:
This new endpoint is focused on deleting specific editions/works from a collection. I will add tests and update the swagger file in future commits.

### DIFF
--- a/api/blueprints/drbCollection.py
+++ b/api/blueprints/drbCollection.py
@@ -139,7 +139,7 @@ def _validateAutoCollectionDef(autoDef: dict) -> str:
 
 @collection.route('/replace/<uuid>', methods=['POST'])
 @validateToken
-def collectionReplace(uuid, user=None):
+def collectionReplace(uuid):
     logger.info('Handling collection replacement request')
 
     collectionData = request.json
@@ -178,7 +178,7 @@ def collectionReplace(uuid, user=None):
         addEditionsToCollection(dbClient, collection, editionIDs)
 
     if workUUIDs:
-        addWorkEditionsToCollection(dbClient, collection, workUUIDs)
+        addWorksToCollection(dbClient, collection, workUUIDs)
 
     dbClient.session.commit()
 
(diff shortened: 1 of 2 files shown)


api/blueprints/drbCollection.py
```diff
@@ -448,11 +491,35 @@ def _buildPublications(editions):
...
-def removeEditionsFromCollection(dbClient, collection):
+def removeEditionsFromCollection(dbClient, collection, editionIDs=None, workUUIDs=None):
 
     '''Deleting the rows of collection_editions that were in the original collection'''
 
-    dbClient.session.execute(COLLECTION_EDITIONS.delete().where(COLLECTION_EDITIONS.c.collection_id == collection.id))
+    #Delete every row of the collection
+    if editionIDs == None and workUUIDs == None:
+        dbClient.session.execute(COLLECTION_EDITIONS.delete()\
+            .where(COLLECTION_EDITIONS.c.collection_id == collection.id))
```
sarangj, line 501: This behavior strikes me as a *little* unexpected. If I pass `None` to a method to remove things, I would probably expect nothing happen to happen.  Maybe this part should be a separate function, like `removeAllEditionsFromCollection`?   Thoughts?

That said, it looks like the calling code as written will never reach this case.

api/blueprints/drbCollection.py
```diff
@@ -448,11 +491,35 @@ def _buildPublications(editions):
...
+    #Delete the rows that match the editionIDs
+    if editionIDs != None:
+        for eid in editionIDs:
+            dbClient.session.execute(COLLECTION_EDITIONS.delete()\
+                .where(COLLECTION_EDITIONS.c.edition_id == eid))
```
sarangj, line 506: You could pretty easily avoid the loop here I think:

```
dbClient.session.execute(COLLECTION_EDITIONS.delete().where(COLLECTION_EDITIONS.c.edition_id.in_(editionIDs))
```
But these will all get executed in a single transaction as is anyway, so probably not a huge deal!

sarangj (approved):
looks good to me, logic in the route handlers is very clear, just a couple thoughts / questions!

api/blueprints/drbCollection.py
mitri-slory: This case will be called for the collectionReplace endpoint specifically because in that endpoint there is no editionIDs or workUUIDs passed as parameters. I'll take you advice and put this one edge case into into its own method though.

api/blueprints/drbCollection.py
mitri-slory: Good suggestion on simplifying the code.

PR #240, opened 2023-03-23
fix: Don't query collections in loop for list

sarangj:
Provide a little private helper so that if we already have a fresh collection object, we don't need to refetch it when constructing the feed.

I'm not sure this will fix the problems in QA, but worth a shot?

### DIFF
--- a/api/blueprints/drbCollection.py
+++ b/api/blueprints/drbCollection.py
@@ -100,7 +100,7 @@ def collectionCreate(user=None):
 
     logger.info('Created collection {}'.format(newCollection))
 
-    opdsFeed = constructOPDSFeed(newCollection.uuid, dbClient)
+    opdsFeed = _constructOPDSFeedForCollection(newCollection, dbClient)
 
     return APIUtils.formatOPDS2Object(201, opdsFeed)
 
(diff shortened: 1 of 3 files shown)


api/blueprints/drbCollection.py
```diff
@@ -379,11 +379,9 @@ def collectionList():
     )
 
     for collection in collections:
-        uuid = collection.uuid
+        path = '/collection/{}'.format(collection.uuid)
 
-        path = '/collection/{}'.format(uuid)
-
-        group = constructOPDSFeed(uuid, dbClient, perPage=5, path=path)
+        group = constructOPDSFeed(collection, dbClient, perPage=5, path=path)
```
mitri-slory, line 384: Shouldn't the method call be `_constructOPDSFeedForCollection` since we're passing collection as the parameter and not the uuid.

mitri-slory (commented):
This fix looks good and I have one suggestion that might be a typo in a comment I made below.

api/blueprints/drbCollection.py
sarangj: Thanks for the catch!

sarangj:
I would say this is worth revisiting, but since it wasn't the actual cause, I'm gonna close this!

PR #242, opened 2023-03-30
Add script to run prod DB migration

sarangj:
Whipped this up to be able to run the prod migration from a non prod machine. This does assume a couple things:

  1. You have a profile in your local aws config called 'digital-dev' (this could maybe be passed as a param, but it would be easy enough to alter the script before running :shrug)
  2. You have installed jq (https://stedolan.github.io/jq/) which is an awesome command line JSON parser

A pretty bare bones script, but thought it would be helpful to at least document how I ran the migration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Supported ES backed automatic collections
 - Add ability to create automatic collections
 - Updated OCLC Classify manager with new API key and header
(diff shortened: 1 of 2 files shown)


mitri-slory (approved):
This script looks good. Smart idea to add this script for easier production migration cycles in the future.

PR #243, opened 2023-04-03
Document release process

sarangj:
I think this is a pretty standard use of git tags and github releases, but I figured I'd write it down!


mitri-slory (approved):
This is a really good addition to the README file especially for future engineers during the onboarding process.

PR #244, opened 2023-04-10
HOTFIX_DeleteCollectionAuthentication

mitri-slory:
This hotfix allows any authenticated user to delete a collection even if they're not the owners of that collection.


sarangj (approved):
Seems reasonable, curious how this came up, QA?

mitri-slory:
@sarangj This issue came from Sean trying to delete collections he didn't create in QA, so this is our solution for the meantime.

PR #246, opened 2023-04-17
SFR-1647v2_DeleteDuplicateAuthors&Contributors

mitri-slory:
This PR addresses an issue that came up with running the original version of this script where any author/contributor JSONB object that had more than 2 author/contributor names would have the next element after the first one delete itself because the `nextIndex` would stay at index of 1 instead of always being one more than than the `currIndex`. The change to the for loop should fix this issue and logging was added for testing and dry run attempts. Lastly, the edition for loop is outside of the work for loop because the query originally would only find editions that were in works with multiple authors/contributors instead of any edition with multiple contributors which broadens the query.

### DIFF
--- a/scripts/deleteDuplicateAgents.py
+++ b/scripts/deleteDuplicateAgents.py
@@ -1,7 +1,8 @@
 import os
 import re 
+import logging 
 
-from model import Work
+from model import Work, Edition
 from managers import DBManager
 from Levenshtein import jaro_winkler
 from logger import createLog
@@ -9,7 +10,9 @@ from sqlalchemy import or_, func
 
 logger = createLog(__name__)
 
-def main():
+logging.basicConfig(filename='duplicateAgents.log', encoding='utf-8', level=logging.INFO)
+
+def main(dryRun=True):
 
     '''Deleting current duplicate authors/contributors with improved Levenshtein implementation'''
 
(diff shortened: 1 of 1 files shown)


scripts/deleteDuplicateAgents.py
```diff
@@ -60,38 +77,50 @@ def deleteDuplicateAgents(agent, count, type):
             break
                    
         cleanCurrAgent = re.sub(r'[.,:\(\)\-0-9]+', '', currAgent['name'].lower())
+        logging.info(cleanCurrAgent)
 
-        for nextAgent in agent[1:]:
+        for nextAgent in agent[currIndex+1:]:
```
sarangj, line 82: ah i see the issue, makes sense!

scripts/deleteDuplicateAgents.py
```diff
@@ -34,23 +37,37 @@ def main():
...
 
-            for edition in work.editions:
-                if edition.contributors and len(edition.contributors) > 1: 
-                    edition.contributors = deleteDuplicateAgents(edition.contributors, \
-                                                                countEditContrib, 'edition contributor')
-                    dbManager.session.add(edition)
+    for edition in dbManager.session.query(Edition) \
+        .filter(func.jsonb_array_length(Edition.contributors) > 1) \
+        .yield_per(batchSize):
+            if edition.contributors and len(edition.contributors) > 1:
```
sarangj, line 52: do you need this `if`?  Seems like you're already filtering in the query.

sarangj (approved):
lg, one question on a redundant conditional, but functionally the changes make sense!

PR #248, opened 2023-04-18
fix: Install pybind11 as fasttext dependency

sarangj:
pybind11 is not an explicit pip dependency of fasttext, but is instead required for the build step of fasttext (I guess there's a cpp binary or something in there that gets compiled during build-wheels). So explicitly install pybind11 to see if that does the trick. While we're here, pin the versions of both libs

i also don't think this will work


sarangj:
right again

PR #253, opened 2023-05-12
SFR-1684_Title&AuthorChicagoISAC

mitri-slory:
This PR adds the title and author metadata to the Chicago ISAC json file that will be created and eventually mapped into the DRB PSQL database. It also modifies the `cleanedText` method to make it more robust and compact. The change to the `deleteDuplicates` methods is simply to avoid the log file running when `scriptrunner.py` is called for other scripts.

### DIFF
--- a/scripts/chicago_isac_scraper.py
+++ b/scripts/chicago_isac_scraper.py
@@ -15,6 +15,7 @@ request_urls = [
     'https://isac.uchicago.edu/research/publications/chicago-assyrian-dictionary',
     'https://isac.uchicago.edu/research/publications/chicago-hittite-dictionary',
     'https://isac.uchicago.edu/research/publications/chicago-hittite-dictionary-supplements-chds',
+    'https://isac.uchicago.edu/research/publications/oriental-institute-seminars-ois',
     'https://isac.uchicago.edu/research/publications/late-antique-and-medieval-islamic-near-east-lamine',
     'https://isac.uchicago.edu/research/publications/materials-assyrian-dictionary',
     'https://isac.uchicago.edu/research/publications/materials-and-studies-kassite-history-mskh',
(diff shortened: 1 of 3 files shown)


scripts/chicago_isac_scraper.py
```diff
@@ -438,31 +592,74 @@ def cleaningText(metaData):
...
+                metaData[i][key] = re.sub(charPattern, '', metaData[i][key])
+                metaData[i][key] = re.sub(isbnPattern, '', metaData[i][key])
+                metaData[i][key] = re.sub(authorPattern, '', metaData[i][key])
+                metaData[i][key] = re.sub(titlePattern, '', metaData[i][key])
+                metaData[i][key] = metaData[i][key].strip()
+    
     return metaData
 
+def cleaningAuthorMetadata(authors):
+    authorPattern = re.compile(r'ed\.|eds\.|([\d]+)\.| and |([\d]+)|Originally published in ([\d]+)\.|\n')
```
sarangj, line 611: not a huge deal since this is a script, but it can make sense to define this pattern outside the function at the top level so you don't recompile the pattern every time, but the perf difference is pretty negligible I would assume!

scripts/chicago_isac_scraper.py
```diff
@@ -438,31 +592,74 @@ def cleaningText(metaData):
...
+                                          OIP [\d]+. |SAOC [/d]+. ')
+                metaData[i][key] = re.sub(charPattern, '', metaData[i][key])
+                metaData[i][key] = re.sub(isbnPattern, '', metaData[i][key])
+                metaData[i][key] = re.sub(authorPattern, '', metaData[i][key])
+                metaData[i][key] = re.sub(titlePattern, '', metaData[i][key])
+                metaData[i][key] = metaData[i][key].strip()
+    
     return metaData
 
+def cleaningAuthorMetadata(authors):
```
sarangj, line 610: One other thing that's not too urgent in a script but more a nit: it can be easier to read function names in a more imperative (? not sure if that's the right word lol) tone, ie, something like `cleanAuthorMetadata`

sarangj (approved):
Nice nice, couple non blocking suggestions!

scripts/chicago_isac_scraper.py
mitri-slory: I can do that.

PR #254, opened 2023-05-19
Pin python deps

sarangj:
Logged into a prod EC2 instance, did a `docker exec -it <container> /bin/bash` into the running DRB python process, then ran `pip freeze -r requirements.txt` to get the exct versions running in prod.

Note, flask in prod is currently running 2.2.3, but I left it as is here at 2.2.4, as that was an intentional change introduced in https://github.com/NYPL/drb-etl-pipeline/commit/17e4b8587c72541588ef6a9fdc4557539dfcd12e not yet deployed to prod.


mitri-slory (approved):
Good quality assurance for the future.

PR #255, opened 2023-05-24
Sfr-1673_ChicagoISACMapping

mitri-slory:
This PR creates a mapping and process for the Chicago ISAC catalog metadata with changes to other methods and the web scraper script to get more accurate data.


processes/chicagoISAC.py
```diff
@@ -0,0 +1,114 @@
...
+                    manifestURI,
+                    source,
+                    'application/webpub+json',
+                    flags
+                ])
+                record.has_part.insert(0, linkString)
+                for i in range(1, len(record.has_part)):
+                    if len(record.has_part[i]) > 1:
+                        urlArray = record.has_part[i]
+                        record.has_part.pop(i)
```
sarangj, line 80: hmm modifying an array while iterating over its indices could be problematic.  You might need to build a list of indices to delete, then after the loop is finished, delete all the elements in another loop (probably iterating backwards?)

sarangj (commented):
Nice! A couple of suggestions and questions.

mappings/chicagoISAC.py
```diff
@@ -0,0 +1,42 @@
...
+        #Formatting for multiple isbns
+        if ',' in self.record.identifiers[0]:
+            identifierArray = self.record.identifiers[0].split(', ')
+            updatedIdentifierArray = []
+            i = 0
+            while len(updatedIdentifierArray) < len(identifierArray):
+                if i == len(identifierArray)-1:
+                    updatedIdentifierArray.append(identifierArray[i].strip())
+                else:
+                    updatedIdentifierArray.append(f'{identifierArray[i].strip()}|isbn')
```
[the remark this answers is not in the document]
mitri-slory: Every element is an isbn based on the metadata gathered from ISAC and for the if conditional we just add the identifier, 92883737, by itself since the '|isbn' string will be automatically added to the last element of the identifiers list mapping.

processes/chicagoISAC.py
```diff
@@ -0,0 +1,114 @@
...
+        # S3 Configuration
+        self.s3Bucket = os.environ['FILE_BUCKET']
+        self.createS3Client()
+
+    def runProcess(self):
+        with open('chicagoISAC_metadata.json') as f:
+                chicagoISACData = json.load(f)
+
+        for dict in chicagoISACData:
+            self.processChicagoISACRecord(dict)
```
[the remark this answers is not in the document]
mitri-slory: Good call!

mappings/chicagoISAC.py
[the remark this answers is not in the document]
sarangj: ah so the mapping stuff automatically adds 'isbn' to the full comma separated string? Got it.

mappings/chicagoISAC.py
```diff
@@ -0,0 +1,42 @@
...
+        return {
+            'title': ('title', '{0}'),
+            'authors': ('authors', '{0}|||true'),
+            'dates': [('publicationDate', '{0}|publication_date')],
+            'publisher': ('publisher', '{0}'),
+            'identifiers': 
+                [('isbn', '{0}|isbn')]
+            ,
+            'is_part_of': ('series', '{0}|series'),
+            'spatial': ('publisherLocation', '{0}|publissherLocation'),
```
sarangj, line 18: small thing I missed, is that a typo? Looks like you have an extra 's' in 'publissherLocation'

mitri-slory: That is a type. Thanks for pointing it out.

processes/chicagoISAC.py
mitri-slory: I don't believe this will be problematic since before typing this code the mapping for a ISAC record with multiple incoming urls would be [manifestURL, [pdfURL, pdfURL2]] when we want it to be like this [manifestURL, [pdfURL, pdfURL2]. The only element that would be popped would be the second element of the has_part array which will always be an array of pdfURLS and then we would append those elements of that array of pdfURLS back into the has_part array as strings.

processes/chicagoISAC.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def processChicagoISACRecord(self, record):
+        try:
+            chicagoISACRec = ChicagoISACMapping(record)
+            chicagoISACRec.applyMapping()
+            self.storePDFManifest(chicagoISACRec.record)
+            self.addDCDWToUpdateList(chicagoISACRec)
+            
+        except (MappingError, HTTPError, ConnectionError, IndexError, TypeError) as e:
+            logger.debug(e)
```
sarangj, line 49: Could be good to do `logger.exception` here, which will log the stack trace for you

processes/chicagoISAC.py
sarangj: I see! I think it would be really helpful to add this as a comment explaining the loop logic!

processes/met.py
```diff
@@ -16,7 +16,7 @@
 class METProcess(CoreProcess):
     MET_ROOT_URL = 'https://libmma.contentdm.oclc.org/digital'
 
-    # The documentatino for these API queries is here: https://help.oclc.org/Metadata_Services/CONTENTdm/Advanced_website_customization/API_Reference/CONTENTdm_API/CONTENTdm_Server_API_Functions_-_dmwebservices
+    # The documentation for these API queries is here: https://help.oclc.org/Metadata_Services/CONTENTdm/Advanced_website_customization/API_Reference/CONTENTdm_API/CONTENTdm_Server_API_Functions_-_dmwebservices
```
sarangj, line 19: nice

tests/unit/test_isac_mapping.py
```diff
@@ -0,0 +1,56 @@
...
+
+class TestISACMapping:
+    @pytest.fixture
+    def testMapping(self):
+        class TestISAC(ChicagoISACMapping):
+            def __init__(self):
+                self.source = {'isbn': '1112223334445'}
+                self.mapping = None
+        
+        return TestISAC()
```
sarangj, line 14: nice use of fixtures!

tests/unit/test_isac_mapping.py
```diff
@@ -0,0 +1,56 @@
+import pytest
+
+from mappings.chicagoISAC import ChicagoISACMapping
+
+
+class TestISACMapping:
```
sarangj, line 6: FYI, with `pytest`, you don't really need classes here! You can just define all these functions in the module scope and the tests will just work. Feel free to merge as is, just an fyi!

tests/unit/test_met_process.py
```diff
@@ -6,7 +6,7 @@
 from tests.helper import TestHelpers
 
 
-class TestMUSEProcess:
+class TestMetProcess:
```
sarangj, line 9: nice catch!

sarangj (approved):
a couple small pieces of feedback, but looks good overall!

PR #256, opened 2023-06-01
SFR-1697_ISACRecordClassificationError

mitri-slory:
This PR fixes the classification error that I ran into when attempting to cluster the ISAC records.

### DIFF
--- a/managers/oclcClassify.py
+++ b/managers/oclcClassify.py
@@ -18,6 +18,7 @@ class ClassifyManager:
     Returns:
         [str] -- A string of XML data comprising of the Classify response body.
     """
+    #Forward slash at the end is needed before the query parameters for the classification API to work
     CLASSIFY_ROOT = 'https://metadata.api.oclc.org/classify/'
 
     LOOKUP_IDENTIFIERS = [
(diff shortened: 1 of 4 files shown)


managers/oclcClassify.py
```diff
@@ -18,7 +18,7 @@ class ClassifyManager:
     Returns:
         [str] -- A string of XML data comprising of the Classify response body.
     """
-    CLASSIFY_ROOT = 'https://metadata.api.oclc.org/classify'
+    CLASSIFY_ROOT = 'https://metadata.api.oclc.org/classify/'
```
sarangj, line 21: It might be worth s code comment explaining that this slash is necessary for the endpoint to work!

managers/oclcClassify.py
```diff
@@ -60,7 +60,8 @@ def getClassifyResponse(self):
 
         try:
             self.execQuery()
-        except (ConnectTimeout, HTTPError, ReadTimeout):
+        except (ConnectTimeout, HTTPError, ReadTimeout) as e:
+            logger.exception(e)
```
sarangj, line 64: Awesome!

sarangj (approved):
Awesome, just one suggestion on where we could add some context, but looks good!

managers/oclcClassify.py
mitri-slory: Agreed

PR #257, opened 2023-06-01
Add ISAC to Source Priority dictionary

mitri-slory:
This hotfix adds ISAC to the Source Priority dictionary which allows ISAC works to be searched for in the API and on the frontend.


sarangj:
Ah, TIL!

PR #262, opened 2023-07-07
SFR-1717_LOCAPIProcess

mitri-slory:
This branch focuses on creating a process for ingesting the LOC Open Access and Digitized Book Collections. I've been able to ingest the works into the records table of my local database successfully. The only problem I've run into is attempting to store an epub manifest for the works that have epub links which results in a blocking RabbbitMQ connection error. This is why lines 29-33 and line 166 are commented out for right now.

### DIFF
--- a/mappings/UofSC.py
+++ b/mappings/UofSC.py
@@ -12,7 +12,7 @@ class UofSCMapping(JSONMapping):
             'authors': [('authors', '{0}|||true')],
             'dates': [('publicationDate', '{0}|publication_date')],
             'publisher': [('publisher', '{0}')],
-            'spatial': ('publisherLocation', '{0}|publisherLocation'),
+            'spatial': ('publisherLocation', '{0}'),
             'identifiers': 
                 [('isbn', '{0}|isbn')]
             ,
--- a/mappings/loc.py
+++ b/mappings/loc.py
@@ -1,4 +1,5 @@
 from .json import JSONMapping
+import ast
 
 class LOCMapping(JSONMapping):
     def __init__(self, source):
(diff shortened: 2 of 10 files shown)


mappings/UofSC.py
```diff
@@ -12,7 +12,7 @@ def createMapping(self):
             'authors': [('authors', '{0}|||true')],
             'dates': [('publicationDate', '{0}|publication_date')],
             'publisher': [('publisher', '{0}')],
-            'spatial': ('publisherLocation', '{0}|publisherLocation'),
+            'spatial': ('publisherLocation', '{0}'),
```
sarangj, line 15: is this and the ISAC change below meant to be in this PR?

mappings/loc.py
```diff
@@ -8,32 +9,98 @@ def __init__(self, source):
...
 
     def applyFormatting(self):
+        self.record.has_part = []
         self.record.source = 'loc'
-        self.record.source_id = self.record.identifiers[0]
         self.record.medium = self.record.medium[0]
-        self.record.spatial = self.record.spatial.split(':')[0].strip(' ')
+
+        #Convert string repr of list to actual list
+        itemList = ast.literal_eval(self.record.identifiers[1])
```
sarangj, line 38: hmmm, what does `self.record.identifiers` look like here? If it's just a json string, I think it would be better to do a `json.loads`, would be a bit clearer / more standard. Unless the format of the string is not valid json and is something else?

mappings/loc.py
```diff
@@ -8,32 +9,98 @@ def __init__(self, source):
...
+            self.record.rights = []
+
+        #Languages Formatting
+        if 'language' in itemList:
+            languageArray = []
+            for i, elem in enumerate(itemList['language']):
+                languageArray.append(f'||{elem}')
+            self.record.languages = languageArray
+        else:
+            self.record.languages = []
```
sarangj, line 104: Thanks for the comment blocks, they are helpful! However, maybe we could just separate these into individual methods? Ie, I think it would read easier if instead of this last language formatting block, we extracted the logic here into a `format_languages` method, and have the code here look like `self.record.languages = formatLanguage(itemList)`, similar for all the other logical blocks here!

processes/loc.py
```diff
@@ -0,0 +1,208 @@
...
+        self.s3Bucket = os.environ['FILE_BUCKET']
+        self.createS3Client()
+
+    def runProcess(self):
+        count = 0
+        count2 = 0
+        sp = 1
+        openAccessURL = '{}&sp={}'.format(LOC_ROOT_OPEN_ACCESS, sp)
+        try:
+            while sp > 0:
```
sarangj, line 45: hmm, I'm a bit confused how this works. Seems like `sp` starts at 1 and only increases? Unless I'm missing something!

processes/loc.py
```diff
@@ -0,0 +1,208 @@
...
+        # self.createRabbitConnection()
+        # self.createOrConnectQueue(self.fileQueue, self.fileRoute)
+
+        # S3 Configuration
+        self.s3Bucket = os.environ['FILE_BUCKET']
+        self.createS3Client()
+
+    def runProcess(self):
+        count = 0
+        count2 = 0
```
sarangj, line 41: Might be more readable to name these according to what they're counting, ie, `openAccessRequestCount` and `digitizeRequestCount`?

processes/loc.py
```diff
@@ -0,0 +1,208 @@
...
+                openAccessURL = '{}&sp={}'.format(LOC_ROOT_OPEN_ACCESS, sp)
+                jsonData = self.fetchPageJSON(openAccessURL)
+                with open("locAPI_metadata_OPAccess.json", "w", encoding='utf-8') as write_file:
+                    json.dump(jsonData.json(), write_file, ensure_ascii = False, indent = 6)
+                with open("locAPI_metadata_OPAccess.json") as f:
+                    LOCData = json.load(f)
+                    for metaDict in LOCData['results']:
+                        if 'pdf' in metaDict['resources'][0].keys() or 'epub_file' in metaDict['resources'][0].keys():
+                            self.processLOCRecord(metaDict)
+                            count += 1
```
sarangj, line 55: question, seems like we're writing to a file, only to read the file and process the records. Is there any need to do any file IO at all? Seems like we could just process the results of `fetchPageJSON` directly

processes/loc.py
```diff
@@ -0,0 +1,208 @@
...
+                    logger.debug(f'Count for OP Access: {count}')
+                sp += 1
+
+        
+        except Exception or HTTPError as e:
+            logger.exception(e)
+            logger.debug('OPEN ACCESS Collection Ingestion Complete')
```
sarangj, line 62: Would this be a misleading log statements? Maybe it should mention that there was an error

processes/loc.py
```diff
@@ -0,0 +1,208 @@
...
+
+        sp = 1
+        try:
+            while sp > 0:
+                digitizedURL = '{}&sp={}'.format(LOC_ROOT_DIGIT, sp)
+                jsonData = self.fetchPageJSON(digitizedURL)
+                with open("locAPI_metadata_digitzed.json", "w", encoding='utf-8') as write_file:
+                    json.dump(jsonData.json(), write_file, ensure_ascii = False, indent = 6)
+                with open("locAPI_metadata_digitzed.json") as f:
+                    LOCData = json.load(f)
```
sarangj, line 73: similar feedback with regards to potentially unnecessary file IO

processes/loc.py
```diff
@@ -0,0 +1,208 @@
...
+        try:
+            while sp > 0:
+                openAccessURL = '{}&sp={}'.format(LOC_ROOT_OPEN_ACCESS, sp)
+                jsonData = self.fetchPageJSON(openAccessURL)
+                with open("locAPI_metadata_OPAccess.json", "w", encoding='utf-8') as write_file:
+                    json.dump(jsonData.json(), write_file, ensure_ascii = False, indent = 6)
+                with open("locAPI_metadata_OPAccess.json") as f:
+                    LOCData = json.load(f)
+                    for metaDict in LOCData['results']:
+                        if 'pdf' in metaDict['resources'][0].keys() or 'epub_file' in metaDict['resources'][0].keys():
```
sarangj, line 53: This line might read a little clearer if you define a little var:
```
resources = metaDict['resources'][0]
if 'pdf' in resources or 'epub_file' in resources:
```

processes/loc.py
```diff
@@ -0,0 +1,208 @@
...
+                jsonData = self.fetchPageJSON(digitizedURL)
+                with open("locAPI_metadata_digitzed.json", "w", encoding='utf-8') as write_file:
+                    json.dump(jsonData.json(), write_file, ensure_ascii = False, indent = 6)
+                with open("locAPI_metadata_digitzed.json") as f:
+                    LOCData = json.load(f)
+                    for metaDict in LOCData['results']:
+                        if 'pdf' in metaDict['resources'][0].keys() or 'epub_file' in metaDict['resources'][0].keys():
+                            self.processLOCRecord(metaDict)
+                            count2 += 1
+                    logger.debug(f'Count for Digitzed: {count2}')
```
sarangj, line 78: misspelled 'Digitized' 😄

sarangj (commented):
I know you're still working on testing the rabbit part, but generally looks good, just some feedback / suggestions!

mappings/UofSC.py
mitri-slory: Yes it is. This is just a typo that both mappings had that I wanted to get rid of now.

mappings/loc.py
mitri-slory: I mapped the `items` block to `self.record.identifiers[1]` and it comes in as a string representation of a dictionary with single quotes while a valid JSON strings needs to be in double quotes. Thus, I get the `Expecting property name enclosed in double quotes` error when using `json.loads`. I could use` json.dumps` then use `json.loads` but that's extra lines of code compared to the one with the `ast.literal_eval` method I'm currently using.

mappings/loc.py
mitri-slory: That makes sense. I'll implement that into my next commit.

processes/loc.py
mitri-slory: The `sp` flag represents the page number of the search results for the LOC works. The sp value will keep increasing and when the `sp` value goes past the last page number, then an HTTP error will occur which is where the `except `block will catch that error and it will be an indication of when the whole collection is ingested.

processes/loc.py
mitri-slory: Sounds good

processes/loc.py
mitri-slory: You're right and I did it this way because I wanted to keep track of what LOC works that are being processed if an error occurs, so I can know which specific work is causing an error. However, i could simply log out the current url and title of the work every time I run through the process instead to keep track that instead.

processes/loc.py
mitri-slory: Nice catch

mappings/loc.py
sarangj: hmm, I see. One follow up, when I throw that string into the `literal_eval` method, it seems to not like it:
```
>>> ast.literal_eval("['identifier': 'test']")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py", line 62, in literal_eval
    node_or_string = parse(node_or_string.lstrip(" \t"), mode='eval')
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
  File "<unknown>", line 1
    ['identifier': 'test']
                 ^
SyntaxError: invalid syntax
```

is that the right string? And could you maybe add a comment describing the string here so it's more clear why we're using `literal_eval`?

mappings/loc.py
```diff
@@ -8,32 +10,111 @@ def __init__(self, source):
...
+                spatialArray.append(pubLocation)
+        return (pubArray, spatialArray)
+    
+    #Extent Formatting
+    def formatExtent(self, itemList):
+        extentArray = []
+
+        if 'medium' in itemList:
+            for elem in itemList['medium']:
+                extentArray.append(elem)
```
sarangj, line 87: seems like you're just returning `itemList['medium']` here right? Do you need the loop at all?

sarangj (approved):
looks good! Just a couple small questions / suggestions!

mappings/loc.py
mitri-slory: It's supposed to be a dictionary like "{'identifier': "test'}". If you put curly braces, then it should work for you.

processes/loc.py
mitri-slory: I'll add a comment to clarify my reasoning and I'll put a very high limit for sp just in case.

mappings/loc.py
mitri-slory: You're right. I'll just use the extend method for the extentArray instead to get all the elements of the medium array.

PR #263, opened 2023-07-10
Fix DB perf issues

sarangj:
See each commit, but both commits are essentially to fix issues with running queries in loops. The first refactors a method call to accept `collection` objects instead of uuids to avoid an extra roundtrip to the DB.

The second one eager loads collection relationships to again avoid roundtrips in the DB in some loops

### DIFF
--- a/api/db.py
+++ b/api/db.py
@@ -178,20 +178,37 @@ class DBClient():
         return (baseQuery.count(), baseQuery.offset(offset).limit(size).all())
 
     def fetchSingleCollection(self, uuid):
-        return self.session.query(Collection)\
-            .options(joinedload(Collection.editions))\
-            .filter(Collection.uuid == uuid).one()
+        return (
+            self.session.query(Collection)
+                .options(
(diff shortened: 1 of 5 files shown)


api/db.py
```diff
@@ -178,20 +178,37 @@ def fetchNewWorks(self, page=0, size=50):
...
+            self.session.query(Collection)
+                .options(
+                    joinedload(Collection.editions),
+                    joinedload(Collection.editions, Edition.links),
+                    joinedload(Collection.editions, Edition.items),
+                    joinedload(Collection.editions, Edition.items, Item.links),
+                    joinedload(Collection.editions, Edition.items, Item.rights),
+                )
+                .filter(Collection.uuid == uuid).one()
+        )
```
mitri-slory, line 191: Good change in making the query load all the database models at once to avoid repeated queries to different models.

mitri-slory (approved):
Looks great and hopefully these changes resolve our performance issues.

PR #265, opened 2023-07-11
Wrap DDL in dev setup process in `sqlalchemy.text`

sarangj:
Noticed when spinning up a new environment that these SQL statements fail because our version of sqlalchemy doesn't accept raw strings, but rather requires wrapping them in executable text objects.

I also switched over to f strings while I was at it

### DIFF
--- a/processes/developmentSetup.py
+++ b/processes/developmentSetup.py
@@ -5,6 +5,7 @@ import gzip
 import os
 from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
 import requests
+import sqlalchemy as sa
 from sqlalchemy.exc import ProgrammingError
 from time import sleep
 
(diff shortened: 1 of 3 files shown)


processes/developmentSetup.py
```diff
@@ -70,16 +71,24 @@ def initializeDB(self):
...
+                    sa.text(
+                        f"CREATE USER {os.environ['POSTGRES_USER']} "
+                        f"WITH PASSWORD '{os.environ['POSTGRES_PSWD']}'",
+                    ),
+                )
+                conn.execute(
+                    sa.text(
+                        f"GRANT ALL PRIVILEGES ON DATABASE {os.environ['POSTGRES_NAME']} "
+                        f"TO {os.environ['POSTGRES_USER']}",
+                    ),
```
mitri-slory, line 91: So when you use two f strings in the sa.text field, they will combine into one string? I'm assuming you did it like this since you don't want one very long f string.

mitri-slory (approved):
Besides one question below, this looks great! It's good that you caught these errors.

processes/developmentSetup.py
sarangj: pretty much, though I'm really just passing one string to `text` (notice the lack of comma between the two lines). Python automatically interprets this as one string, and yep, did it so that it's not one huge line!

PR #266, opened 2023-07-20
SFR-1718v2_AutoIngestionLOC

mitri-slory:
This PR focuses on adding automatic ingestion for the LOC items. This auto ingestion will be weekly instead of daily due to the LOC site not being as updated consistently compared to other sources in DRB.

### DIFF
--- a/mappings/loc.py
+++ b/mappings/loc.py
@@ -56,7 +56,7 @@ class LOCMapping(JSONMapping):
 
         self.record.languages = self.formatLanguages(itemList)
 
-    #Identifier/SourceID Formatting
+    #Identifier/SourceID Formatting to return (string, string, string)
     def formatIdentifierSourceID(self, itemList):
         newIdentifier = itemList
         lccnNumber = self.record.identifiers[0][0]  #lccnNumber comes in as an array and we need the string inside the array
@@ -68,7 +68,7 @@ class LOCMapping(JSONMapping):
             callNumber = None
         return (lccnNumber, callNumber, sourceID)
     
-    #Publisher/Spatial Formatting
+    #Publisher/Spatial Formatting to return (array, string)
     def formatPubSpatial(self, itemList):
         pubArray = []
         spatialString = None
@@ -92,7 +92,7 @@ class LOCMapping(JSONMapping):
         else:
             return ([], None)
     
-    #Extent Formatting
+    #Extent Formatting to return string
     def formatExtent(self, itemList):
         extentString = ''
 
@@ -103,7 +103,7 @@ class LOCMapping(JSONMapping):
             
         return None
     
-    #Subjects Formatting
+    #Subjects Formatting to return array
     def formatSubjects(self, itemList):
         subjectArray = []
 
@@ -113,7 +113,7 @@ class LOCMapping(JSONMapping):
         
         return subjectArray
     
-    #Rights Formatting
+    #Rights Formatting to return string
     def formatRights(self, itemList):
         rightsString = ''
 
@@ -124,7 +124,7 @@ class LOCMapping(JSONMapping):
             
         return None
         
-    #Languages Formatting
+    #Languages Formatting to return array
     def formatLanguages(self, itemList):
         languageArray = []
 
--- a/mappings/loc.py
+++ b/mappings/loc.py
@@ -10,13 +10,13 @@ class LOCMapping(JSONMapping):
     def createMapping(self):
         return {
             'title': ('title', '{0}'),
-            'alternative': [('other_title', '{0}')], #One other_title in items block and one outside of it 
-            'medium': [('original_format', '{0}')],
+            'alternative': ('other_title', '{0}'),
+            'medium': ('original_format', '{0}'),
             'authors': ('contributor', '{0}|||true'),
             'dates': ('dates', '{0}|publication_date'),
             'publisher': ('item', '{0}'),
             'identifiers': [
-                ('number_lccn', '{0}|loc'),
+                ('number_lccn', '{0}|lccn'),
                 ('item', '{0}'),
             ],
             'contributors': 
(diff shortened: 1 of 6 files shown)


processes/loc.py
```diff
@@ -37,27 +39,49 @@ def __init__(self, *args):
...
+        self.commitChanges()
+    
+
+    def importLOCRecords(self, fullImport=False, startTimestamp=None):
+
+        if not fullImport:
+            if not startTimestamp:
+                startTimestamp = datetime.utcnow() - timedelta(days=7)
+            else:
+                startTimestamp = datetime.strptime(startTimestamp, '%Y-%m-%dT%H:%M:%S')
```
sarangj, line 59: I have a small suggestion to simplify some of these flags / logic, but lmk what you think / if I'm missing something!

It seems like there are 3 cases we actually expect to see:

  1.   fullImport=True and startTimestamp=None
  2.  fullImport=False and startTimestamp=None (which means weekly)
  3. fullImport=False and startTimestamp is set

I think we can cover this same logic *without* the fullImport flag, by passing in a timestamp to `importLOCRecords` even for the weekly ingest.  So move the timestamp logic for a weekly ingest up to the `if self.process = 'weekly'` block in `runProcess`.

Then you can get rid of the if block here at the start of `importLOCRecords`. It is also more intuitive as a reader (at least to me) that when you call `importLOCRecords()` with no args, it's a full import, and if you call it like `importLOCRecords(timestamp)` it always uses the start timestamp.  Right now, it's a little confusing that if you call `self.importLOCRecords()` with no timestamps, it's actually importing a week's worth of records implicitly.

Does that make sense? Or am I missing something here?

processes/loc.py
```diff
@@ -7,10 +7,11 @@
 from mappings.loc import LOCMapping
 from managers import WebpubManifest
 from logger import createLog
+from datetime import datetime, timedelta
 
 logger = createLog(__name__)
 
-LOC_ROOT_OPEN_ACCESS = 'https://www.loc.gov/collections/open-access-books/?fo=json&fa=access-restricted%3Afalse&c=2&at=results'
+LOC_ROOT_OPEN_ACCESS = 'https://www.loc.gov/collections/open-access-books/?fo=json&fa=access-restricted%3Afalse&c=2&at=results&sp=2'
```
sarangj, line 14: curious about this change. It seems like below, you dynamically append `sp` to the url, so will this cause the url to have two `sp` values?

sarangj (commented):
Generally looks good! One suggestion on some of the timestamp logic, but otherwise makes sense (and I know you're still figuring out the epub thing)

processes/loc.py
mitri-slory: Thanks for catching that. I was testing out the urls locally and forgot to get rid of the sp value I set there.

processes/loc.py
mitri-slory: I agree with your assessment and I can see how it's set up now can be confusing. I'll incorporate the timeStamp variable only and get rid of the fullImport variable.

mappings/loc.py
```diff
@@ -51,41 +59,49 @@ def applyFormatting(self):
...
         sourceID = lccnNumber
+        if 'call_number' in newIdentifier.keys():
+            newIdentifier['call_number'][0] = f'{newIdentifier["call_number"][0]}|call_number'
+            callNumber = newIdentifier['call_number'][0].strip(' ')
+        else: 
+            callNumber = None
         return (lccnNumber, callNumber, sourceID)
     
     #Publisher/Spatial Formatting
     def formatPubSpatial(self, itemList):
```
sarangj, line 72: I think maybe a docstring explaining the input / output a bit (doesn't need to be super detailed) would be helpful, since there's a lot of logic in this method!

processes/loc.py
```diff
@@ -7,11 +7,12 @@
...
 from managers import WebpubManifest
 from logger import createLog
+from datetime import datetime, timedelta
 
 logger = createLog(__name__)
 
-LOC_ROOT_OPEN_ACCESS = 'https://www.loc.gov/collections/open-access-books/?fo=json&fa=access-restricted%3Afalse&c=2&at=results'
-LOC_ROOT_DIGIT = 'https://www.loc.gov/collections/selected-digitized-books/?fo=json&fa=access-restricted%3Afalse&c=2&at=results' 
+LOC_ROOT_OPEN_ACCESS = 'https://www.loc.gov/collections/open-access-books/?fo=json&fa=access-restricted%3Afalse&c=50&at=results&sb=timestamp_desc'
+LOC_ROOT_DIGIT = 'https://www.loc.gov/collections/selected-digitized-books/?fo=json&fa=access-restricted%3Afalse&c=50&at=results&sb=timestamp_desc'
```
sarangj, line 15: what does the `c=50` param do here? Just curious!

mappings/loc.py
```diff
@@ -51,41 +59,49 @@ def applyFormatting(self):
...
         pubArray = []
-        spatialArray = []
-        for elem in itemList['created_published']:
-            if ':' not in elem:
-                createdPublishedList = elem.split(',', 1)
-                pubLocation = createdPublishedList[0].strip(' ')
-                if ',' in createdPublishedList[1]:
-                    pubOnly = createdPublishedList[1].split(',')[0].strip(' ')
+        spatialString = None
+        if 'created_published' in itemList.keys():
```
sarangj, line 75: more of just a nit, since it's functionally the same, but here, you don't need to call `.keys()` to check dictionary membership, you can just do `if 'created_published' in itemList.keys():`

Or if you wanna get fancy with the walrus operator, you can do:

```
if createdPublished := itemList.get('created_published', []):
    ...do stuff with `createdPublished
```

Up to you, just thought I'd throw it out there since there's a lot of instances in this file of checking dict membership and doing stuff with / iterating over values
```

sarangj (approved):
A couple questions / suggestions, feel free to apply / ignore as makes sense to you!

processes/loc.py
mitri-slory: It returns the results/works per page in the json response that's returned

mappings/loc.py
mitri-slory: 'but here, you don't need to call .keys() to check dictionary membership, you can just do if 'created_published' in itemList.keys():' The line you recommended is the same as what I wrote in the code. Did you mean to type something else? Also, the walrus operator is a interesting technique I'll look into using.

mappings/loc.py
mitri-slory: Sounds good.

PR #268, opened 2023-07-25
HOTFIX_ISACHasPartConditional

mitri-slory:
This hotfix adds one if conditional to avoid running the `updatePubLocation&Links` script for ISAC records with an empty `has_part` array.

### DIFF
--- a/scripts/updatePubLocationAndLinks.py
+++ b/scripts/updatePubLocationAndLinks.py
@@ -28,18 +28,20 @@ def main():
             record.spatial = re.sub(charPattern, '', record.spatial)
 
             #Reversing download and reader values of webpub mainfest in has_part attribute 
-            if record.has_part:
-                hasPartManifest = record.has_part[0].split('|')
+            if not record.has_part:
+                continue
+            
+            hasPartManifest = record.has_part[0].split('|')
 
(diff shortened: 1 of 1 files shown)


scripts/updatePubLocationAndLinks.py
```diff
@@ -28,17 +28,18 @@ def main():
...
-            if hasPartManifest[len(hasPartManifest)-2] == 'application/webpub+json' \
-            and hasPartManifest[-1] == '{"catalog": false, "download": true, "reader": false, "embed": false}':
-                
-                hasPartManifest[-1] = '{"catalog": false, "download": false, "reader": true, "embed": false}'
-                newHasPart = []
-                newHasPart.append('|'.join(hasPartManifest))
-                for i in record.has_part[1:]:
-                    newHasPart.append(i)
-                record.has_part = newHasPart
+            if record.has_part:
```
sarangj, line 31: I sometimes like to invert to avoid more nesting:
```
if not record.has_part:
    continue
```
Also keeps the git history cleaner! But this seems fine here, just thought I'd mention!

sarangj (approved):

scripts/updatePubLocationAndLinks.py
mitri-slory: Sounds like a good organizational tip to remember. I'll add your implementation to the script.

PR #269, opened 2023-08-01
SFR-1742_AuditUForCABooks

mitri-slory:
This PR focuses on counting the number of public domain University of California books from Hathitrust that are in the DRB database and returning a list of their ids from the database. The json file was created from a column of source ids from the Google Sheet that had the current public domain University of California books from Hathitrust.


sarangj (approved):
ah, very neat, this will be helpful to know!

Might consider moving that JSON file under the scripts/ dir, but otherwise looks great!

PR #271, opened 2023-08-18
Add blurb about checking ECS console during deploy

sarangj:
Got bitten a bit on the last release when I realized the github actions succeeded even though the eventual ECS update failed, since the deploy action just triggers an ECS update, but doesn't poll till it succeeds.

Note, I didn't link the exact ECS cluster / service because this repo is public and I figure we don't wanna leak the names of our IAM resources?

Also, it may make sense to just use something like https://github.com/marketplace/actions/ecs-wait in the future, thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,13 +66,13 @@ All services share a single entry point in `main.py` file. This script dynamical
 
 To set up a local environment there is a special process to initialize a database and search cluster which is the `DevelopmentSetupProcess`. However, it's recommended to run the `DevelopmentSetupProcess` and `APIProcess` at the same time to build the most efficient local environment. Before running a command, it's required to set these config variables in the sample-compose.yaml file:
 
-`HATHI_API_KEY`: 
+`HATHI_API_KEY`:
 `HATHI_API_SECRET`:
 `OCLC_API_KEY`:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -133,6 +133,8 @@ We use git tags to tag releases and github's release feature to deploy.  The ste
      ('v0.12.0')
   9. Add a quick 1-2 sentence summary, make sure 'Set as the latest release' is enabled and hit 'Publish release'
   10. Check the repo's `Actions` tab to observe the progress of the deployment to production
+    - Note that the deployment job merely kicks off an ECS service update. To fully verify success, you'll need to check the
+      `Deployments` tab for the relevant service / cluster in the ECS console.
```
mitri-slory, line 137: The addition of this blurb helps with clarifying the deployment process more since I didn't even know about this part either.

mitri-slory (approved):
This blurb makes sense to add. This link https://github.com/marketplace/actions/ecs-wait can be something we utilize for the next deployment as a test. If it works fine, then maybe we can permanently add it to the deployment steps.

sarangj:
Good idea! Maybe we put up a github actions change when we do the next release!

PR #275, opened 2023-09-05
Hotfix: escape "*" in ES queries to prohibit wildcard use in search

Apophenia:
This change will, at least for the time being, limit the usage of "*" in queries passed to ElasticSearch, since the volume of wildcard queries the app is receiving is creating instability.


mwbenowitz (approved):
🚀

PR #277, opened 2023-11-28
SFR_1826_NYPLLoginFlag

mitri-slory:
This PR adds a script to add an nypl_login flag to the Links objects of DRB. For right now, the nypl_login flag will only be inserted for links that have requestable formats like application/html+edd and application/x.html+edd with a True boolean value. My second commit adds the nypl_login flag to nypl records that are requestable and these records seem to be the only ones that have the requestable formats.


Apophenia (approved):
Thanks, this looks good to me!

PR #278, opened 2023-12-14
SFR-1828_APIS3ObjectLink

mitri-slory:
This PR adds a new API that returns the a unique presigned url for S3 objects from the DRB buckets on AWS. The first commit I made to this PR has changes to the API and tests that are outdated now which I explain in the comments below. However, the first PR includes changes I made to swagger and other files which I recommend to look over as well for feedback.


api/blueprints/drbS3.py
mitri-slory, line 1: This API was the initial format where the link returned was the object url to the S3 object, but I realized later on that we needed to return a unique link to users so it the url won't be shared amongst users outside the organization.

tests/unit/test_api_s3_blueprint.py
mitri-slory, line 1: These tests were made with the initial return to an object url I had in mind for the S3 API, so this is outdated.

swagger.v4.json
```diff
@@ -245,7 +245,7 @@
                     }
                 }
             }
-        },
+        },
```
Apophenia, line 248:
```suggestion
        },
```

Apophenia: If you can revert this whitespace change I think this will be ready to go!

Apophenia (changes requested):
Logic is all good, not necessarily a merge blocker but if you could revert the whitespace change in the Swagger JSON that would be perfect.

Apophenia (approved):

PR #279, opened 2023-12-15
SFR_1864_NoNYPLRecordsCreated

mitri-slory:
This PR intends to fix a sqlalchemy.exc:ObjectNotExecutableError which is an issue that is preventing the NYPL process from running in QA and prod which has prevented the addition of new records directly from NYPL since late last year. This StackOverflow forum [https://stackoverflow.com/questions/69490450/objectnotexecutableerror-when-executing-any-sql-query-using-asyncengine](url) had the same problem I encountered and the solution is what I followed for this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - New script to add nypl_login flag to Links objects
 - Added nypl_login flag to nypl mapping
 ## Fixed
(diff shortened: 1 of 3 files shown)


Apophenia:
Looks good to me! Thanks for the background!

Apophenia (approved):

PR #280, opened 2023-12-18
Add fulfill route and tests

Apophenia:
Does not have 100% test coverage; when everything is hooked up I will add test cases for the positive cases of "yes, this is working" and also figure out how to mock the expired token.

Don't love how much logic lives directly in the Blueprint but given the error structure I'm not sure how to best test it if I were to bundle the auth logic in another file, but open to refactoring.

### DIFF
--- a/api/blueprints/drbFulfill.py
+++ b/api/blueprints/drbFulfill.py
@@ -1,0 +1,62 @@
+import json
+import os
+
+import jwt
+
+from flask import Blueprint, request
+from ..utils import APIUtils
+from logger import createLog
+
+JWT_ALGORITHM = ''
+logger = createLog(__name__)
+
+fulfill = Blueprint('fulfill', __name__, url_prefix='/fulfill')
+
+@fulfill.route('/<uuid>', methods=['GET'])
+def workFulfill(uuid):
+    logger.info('Checking if authorization is needed for work {}'.format(uuid))
+
+    requires_authorization = True
+
+    if requires_authorization:
+        try:
+            bearer = request.headers.get('Authorization')
+            token = bearer.split()[1]
+
+            jwt_secret = os.environ['NYPL_API_CLIENT_PUBLIC_KEY']
+            decoded_token =(jwt.decode(token, jwt_secret, 'RS256', 
+                                   audience="app_myaccount"))
+            if json.loads(json.dumps(decoded_token))['iss'] == "https://www.nypl.org":
+                statusCode = 200
+                responseBody = uuid
+            else:
+                statusCode = 401
+                responseBody = 'Invalid access token'
+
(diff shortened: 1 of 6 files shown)


api/blueprints/drbFulfill.py
```diff
@@ -0,0 +1,62 @@
...
+        except jwt.exceptions.ExpiredSignatureError:
+            statusCode = 401
+            responseBody = 'Expired access token'
+        except (jwt.exceptions.DecodeError, UnicodeDecodeError, IndexError, AttributeError):
+            statusCode = 401
+            responseBody = 'Invalid access token'
+        except ValueError:
+            logger.warning("Could not deserialize NYPL-issued public key")
+            statusCode = 500
+            responseBody = 'Server error'
```
mitri-slory, line 45: Good to catch all these exception cases

config/sample-compose.yaml
```diff
@@ -45,6 +45,12 @@ HATHI_API_ROOT: https://babel.hathitrust.org/cgi/htd
 OCLC_QUERY_LIMIT: '390000'
 #OCLC_API_KEY: 
 
+# NYPL AUTH CONFIGURATION
+NYPL_API_CLIENT_PUBLIC_KEY: >
+  -----BEGIN PUBLIC KEY-----
+  MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA44ilHg/PxcJYsISHMRyoxsmez178qZpkJVXg7rOMVTLZuf05an7Pl+lX4nw/rqcvGQDXyrimciLgLkWu00xhm6h6klTeJSNq2DgseF8OMw2olfuBKq1NBQ/vC8U0l5NJu34oSN4/iipgpovqAHHBGV4zDt0EWSXE5xpnBWi+w1NMAX/muB2QRfRxkkhueDkAmwKvz5MXJPay7FB/WRjf+7r2EN78x5iQKyCw0tpEZ5hpBX831SEnVULCnpFOcJWMPLdg0Ff6tBmgDxKQBVFIQ9RrzMLTqxKnVVn2+hVpk4F/8tMsGCdd4s/AJqEQBy5lsq7ji1B63XYqi5fc1SnJEQIDAQAB
+  -----END PUBLIC KEY-----
```
mitri-slory, line 52: This is a good addition to the sample compose file for the local environment.

mitri-slory (approved):
The new endpoint looks really good. The logic in the blueprint is fine since some of the other blueprints have a lot of logic as well when calling methods. I would only recommend finishing the rest of the tests you want to implement before fully merging this branch into main.

Apophenia:
I'm going to go ahead and merge this, and add the one test I think I'm missing (need to mock an exception thrown for an expired token) with my next PR so that it can be reviewed.

PR #281, opened 2023-12-28
Change Hathi data files source to new URL

Apophenia:
Change an outdated Hathi file listing to a JSON file at a new URL containing the same information.


mwbenowitz (approved):
👍

PR #282, opened 2024-01-09
SFR-1865_S3APIAuthentication

mitri-slory:
This branch adds a new API and endpoint for the AWS S3 objects in DRB. Right now, there's only endpoint and it returns a unique, presigned link to an S3 object with user authentication based on the psql database. I wonder if I should have the user and password be based on something else besides the database so any feedback on that point specifically would be appreciated.

### DIFF
--- a/api/blueprints/drbS3.py
+++ b/api/blueprints/drbS3.py
@@ -1,9 +1,10 @@
 from flask import Blueprint, request, current_app
+from base64 import b64decode
+from ..db import DBClient
 from ..utils import APIUtils
+from functools import wraps
 from logger import createLog
 from functools import wraps
-from botocore.exceptions import ClientError
-from urllib.parse import urlparse
 
 import requests
 import argparse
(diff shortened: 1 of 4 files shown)


api/blueprints/drbS3.py
```diff
@@ -13,8 +14,44 @@
...
+            )
+
+        dbClient = DBClient(current_app.config['DB_CLIENT'])
+        dbClient.createSession()
+
+        user = dbClient.fetchUser(user)
+
+        if not user or APIUtils.validatePassword(password, user.password, user.salt) is False:
+            return APIUtils.formatResponseObject(
+                401, 'authResponse', {'message': 'invalid user/password'}
```
Apophenia, line 40: I only learned this myself a few weeks ago, but [as part of the HTTP standard](https://www.rfc-editor.org/rfc/rfc9110.html#status.401), an endpoint returning a 401 response should also add a WWW-Authenticate header describing of what kind of authentication the endpoint accepts (in this case, basic auth). 

I think it will become a little more important since we will potentially be accepting different kinds of auth at different endpoints.

When I needed to implement this, [I extended the FormatResponseObject in this PR to take a header object](https://github.com/NYPL/drb-etl-pipeline/pull/284/files#diff-1927a2a7260bf138319644bceb09095e47b82f13b79db71504801f8705c00394R485), so that is a potential approach you could take in the future.

Apophenia (commented):
Hi Dmitri, thanks so much for your work on this PR. I talked with Sean and David about this PR, and since I have some security concerns about it (which aren't related directly to the code you wrote here), we are going to wait to discuss this further/merge this until after Limited Access is out.

I do think we should think carefully about what authentication method we use for this service and that's one of the things I think could be a larger discussion before we merge this. My other questions are mostly about the scope of objects that this PR would allow access to, since I'm not sure what the AWS credentials in QA/production would allow in scope (they're different from the credentials I have). I think we should make sure before merging this that it does not allow retrieving file objects managed by other portfolios within Digital, unless the developers who manage those buckets are aware of the possibility.

Over all I think this looks like a good contribution that could provide some utility in the future, but I think we should hold off merging it until we feel good about our security approach.

PR #283, opened 2024-01-11
SFR-1866_Bardo&HathiConfigUpdate

mitri-slory:
This PR updates the Bardo CCE API and Hathi Data Files URLS in the production and qa config. The Bardo CCE API URL was outdated so local running of the NYPL Process was unsuccessful and worked soon after updating this url.


Apophenia (approved):
This looks good to me. Is there any (external or internal) listing of the Bardo CCE API URL, so that if it changed again, we would know to update it?

mitri-slory:
There are no listings for the Bardo CCE API URL.

PR #284, opened 2024-01-18
Sfr 1829 complete fulfill endpoint

Apophenia:
## Description
This pull request completes the fulfill endpoint. It also refactors the fulfill endpoint in order to fix control flow and error issues. The fulfill endpoint blueprint is now called "itemFulfill" because it fulfills items, not works, and uses a link ID as its identifier.

## Motivation and Context
This work is needed in order to return pre-signed URLs from the fulfill endpoint. Additionally, it includes some updates to the DB client code and a slightly different structure from the other blueprints that utilize the DB in an effort to reduce hanging database connections if an unexpected error occurs.


mitri-slory (approved):
Looks great. Good to see the fixes to the endpoint and additions to the sample compose file.

PR #285, opened 2024-01-24
SFR-1775_University Michigan Ingestion

mitri-slory:
This PR adds a process and mapping for the University of Michigan books. It also includes unit tests for the process and mapping as well.


.gitignore
```diff
@@ -13,3 +13,6 @@ config/local*
 tags
 # Vim
 *.swp
+# Ingest JSON files
```
Apophenia, line 16: This wouldn't block merging this PR, but maybe it would be good to just add a utility directory of something like /ingest_json to the .gitignore so that we can add, remove, or rename files as needed without worrying about adding or removing them individually to the gitignore

Apophenia (commented):
Hi Dmitri, I think this looks good, I just have questions for my own reference about the "_100pct.pdf" convention here. Is it the case that all of the items in the bucket have either "_100pct.pdf" or "_060pct.pdf" appended and we just check to see which one appears in the bucket for any given title? Do some of the titles have both (if you happen to know)?

Apophenia (approved):

.gitignore
mitri-slory: Sounds like a good idea. I'll implement that as a commit.

PR #286, opened 2024-02-01
SFR-1877: rewrite URLs as fulfillment URLs

Apophenia:
## Description
This pull request rewrites the URLs of links for objects that are not EDD links and also nypl_login required. This should cover items within the scope of Limited Access. 

Other options (like rewriting the links in the database and referring to a third, new column of S3 urls) were considered, but none of them seemed like they would work without heavily rewriting some of the code related to FRBRized output.

As it stands, almost all of the work is done at the "Format..." level for each hierarchy where links are part of the JSON formatted output. While this makes the change fairly superficial and it must be implemented in many places, it is hopefully transparent enough that this link overwriting can be easily found and modified and/or removed later if needed.

## Motivation and Context
The link URLs in the database represent private s3 URLs. If the s3 URLs are sent as part of the JSON, there won't be any way for users (or the front end) to retrieve the contents of the URLs. The fulfill endpoint already contains the logic needed to generate a fulfill URL that points to the private item bucket via a time-gated URL that requires user authentication; however, the private URLs to s3 must be rewritten with these fulfillment URLs before the API output is returned.

### DIFF
--- a/api/utils.py
+++ b/api/utils.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 from datetime import datetime
 from hashlib import scrypt
 from flask import jsonify
+from itertools import repeat
 from math import ceil
 from model import Collection, Edition
 import re
@@ -139,7 +140,7 @@ class APIUtils():
 
     @classmethod
     def formatWorkOutput(
-        cls, works, identifiers, dbClient, showAll=True, formats=None, reader=None
+        cls, works, identifiers, dbClient, request, showAll=True, formats=None, reader=None,
     ):
         #Multiple formatted works with formats specified
         if isinstance(works, list):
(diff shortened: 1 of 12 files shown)


api/utils.py
```diff
@@ -232,16 +233,25 @@ def addWorkMeta(cls, work, **kwargs):
...
             edition, editionWorkTitle, editionWorkAuthors, editionInCollection, records, formats, showAll=showAll, reader=reader
         )
 
+        if formattedEdition.get("instances"):
+            for instance in formattedEdition['instances']:
+                for item in instance['items']:
+                    # Map over item links and patch with pre-signed URL where necessary
+                    item['links']= list(map(APIUtils.replacePrivateLinkUrl, item['links'], repeat(request)))
+
+        return formattedEdition
```
mitri-slory, line 252: Good way to catch item links which the presigned url should replace.

mitri-slory (approved):
All looks great to me. I believe it's ready to be merged.

PR #288, opened 2024-02-01
Change local development port to 5050 and record change in docs

Apophenia:
MacOS Monterey introduced a new AirPlay feature enabled by default which uses port 5000. While the feature can be disabled, freeing up the port, the port cannot be changed and will continue to conflict with default settings of Flask, Node, and some PSGI/Perl dev servers. While we unfortunately can't fix this, we can change the default port of the project to 5050.

I spent some time attempting to abstract this into an environment variable, but drew a blank about how to dynamically change the port forwarding in the docker-compose setup. Changing the hard-coded values was a relatively straightfoward fix by comparison, but I am open to suggestions about how to assign and open this port dynamically and keep it working with the Docker setup.


Apophenia:
Also tagging @ayan1229, @jackiequach, and @clarissarichard for transparency - once this is merged, the default port for DRB's API when running locally specifically will be 5050 instead of 5000, so any saved tests in Insomnia, Postman, etc may need to be updated.

PR #289, opened 2024-02-01
SFR-1863 Remove Tugboat configuration from codebase

Apophenia:
Tugboat is no longer used and every build fails. This commit removes the Tugboat configuration files.


Apophenia:
Hey! No Tugboat :)

mwbenowitz (approved):

PR #292, opened 2024-02-13
SFR-1891_DeleteUMPManifestLinks

mitri-slory:
This PR adds a new script to delete UMP manifest links from the links table that should not have been created at this time. One question is if the UofM manifest directory in the S3 manifest bucket should be deleted as well?


mwbenowitz (approved):
Looks Good!

PR #293, opened 2024-02-21
Remove reference to Tugboat temporary builds from README

Apophenia:
I assume that this is referencing Tugboat's builds, which no longer automatically open on PR.


mwbenowitz (approved):
👍

PR #295, opened 2024-02-28
HOTFIX_FixFormatFilter500Error

mitri-slory:
This hotfix intends to fix the 500 error that started to occur in the DRB API after using the search endpoint and using the format filter. The error came in after adding the items media type sorting code line which led to a index error due to some items having an empty link list. Removing this line might revert some changes to the frontend with item sorting but the links are properly sorted now due to the change to the sortByMediaType method in APIUtils from a previous hotfix change [https://github.com/NYPL/drb-etl-pipeline/pull/294](url)

### DIFF
--- a/api/utils.py
+++ b/api/utils.py
@@ -366,7 +366,7 @@ class APIUtils():
             editionDict['items'].append(itemDict)
 
         editionDict['items']\
-            .sort(key=lambda x: (cls.SOURCE_PRIORITY[x['source']], cls.sortByMediaType(x['links'][0])))
+            .sort(key=lambda x: (cls.SOURCE_PRIORITY[x['source']]))
 
         if records is not None:
             itemsByLink = {}
--- a/api/utils.py
+++ b/api/utils.py
@@ -365,9 +365,22 @@ class APIUtils():
 
             editionDict['items'].append(itemDict)
 
-        editionDict['items']\
-            .sort(key=lambda x: (cls.SOURCE_PRIORITY[x['source']]))
+        emptyListFlag = False
+        
+        #This for loop is meant to check if a empty links array exists in a item dictionary
+            #If one exists, then sorting by source priority and empty/non-empty priority occurs
+            #Otherwise, sorting by source and link media type occurs
+        for itemDict in editionDict['items']:
+            if itemDict['links'] == []:
+                editionDict['items']\
+                    .sort(key=lambda x: (cls.SOURCE_PRIORITY[x['source']], x['links'] == []))
+                emptyListFlag = True
+                break
 
+        if emptyListFlag == False:
+            editionDict['items']\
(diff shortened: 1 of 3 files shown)


mwbenowitz (commented):
Before approving this want to be clear that this is reverting a change to sorting, but we are _not_ expecting this to impact the new limited access feature? (essentially we'll still surface the correct link on both search result and detail pages?)

api/utils.py
```diff
@@ -365,9 +365,22 @@ def formatEdition(
...
-
+        emptyListFlag = False
+        
+        #This for loop is meant to check if a empty links array exists in a item dictionary
+            #If one exists, then sorting by source priority and empty/non-empty priority occurs
+            #Otherwise, sorting by source and link media type occurs
+        for itemDict in editionDict['items']:
+            if itemDict['links'] == []:
+                editionDict['items']\
+                    .sort(key=lambda x: (cls.SOURCE_PRIORITY[x['source']], x['links'] == []))
```
mwbenowitz, line 376: This looks good but is `x['links'] == []` correct? Seems like that would sort empty lists to the top and `!=` is what we want. Not sure though, just want to check!

mwbenowitz (approved):
Approach looks good and handles this well, just want to be sure that the sorting will take place in the right direction!

api/utils.py
mitri-slory: That is the right way to correct sorting for empty lists since the test `test_formatEdition_v2_reader_flag_2` has a item with a empty list and a item with a non empty list that both have a gutenberg source in which the non empty list becomes the first element in the list after the formattedEdition method is called. This is the result:

`{'id': 'ed1', 'work': {}, 'publication_date': 2000, 'items': [{'id': 'it2', 'links': [{'link_id': 'li2', 'mediaType': 'application/webpub+json', 'url': 'testURI', 'flags': {'test': True}}], 'rights': [], 'physical_location': {}, 'source': 'gutenberg', 'item_id': 'it2', 'location': None}, {'id': 'it4', 'links': [], 'rights': [], 'physical_location': {}, 'source': 'gutenberg', 'item_id': 'it4', 'location': None}, {'id': 'it1', 'links': [{'link_id': 'li1', 'mediaType': 'application/epub+xml', 'url': 'testURI', 'flags': {'test': True, 'reader': False}}], 'rights': [{'source': 'test', 'license': 'testLicense', 'rightsStatement': 'testStatement'}], 'physical_location': {'name': 'test'}, 'source': 'hathitrust', 'item_id': 'it1', 'location': 'test'}], 'links': [{'link_id': 'co1', 'mediaType': 'image/png', 'url': 'testCover'}], 'edition_id': 'ed1', 'work_uuid': 'uuid1', 'inCollections': None, 'work_title': None, 'work_authors': None}`

PR #296, opened 2024-03-18
SFR-1900_ParseDownloadRequests

mitri-slory:
This PR adds a new script to parse download requests for UMP books from AWS server access logs into a json file.


scripts/parseDownloadRequests.py
```diff
@@ -0,0 +1,89 @@
...
+                parseTuple = parseInfo(logObject, requestRegex, referrerRegex, timeStampRegex, fileIDRegex)
+                if parseTuple:
+                    umpDownloadArray.append(parseTuple)
+    umpDownloadCSV = numpy.array(umpDownloadArray)
+    with open('data3.csv', 'w') as f: 
+        f.write(str(umpDownloadCSV))
+
+def load_batch():
+    paginator = s3_client.get_paginator('list_objects_v2')
+    page_iterator = paginator.paginate(Bucket= bucketName, Prefix= 'logs/946183545209/us-east-1/ump-pdf-repository/2024/03/13/')
```
Apophenia, line 45: Can we set the file name and the log prefix as variables at the top of the page (along with the regex constants)? I imagine even if we run this manually we might want an easy way to change the output file names and prefixes, for example.

scripts/parseDownloadRequests.py
```diff
@@ -0,0 +1,89 @@
...
+            for link in dbManager.session.query(Link) \
+                .join(ITEM_LINKS) \
+                .filter(ITEM_LINKS.c.item_id == item.id) \
+                .filter(Link.media_type == 'application/pdf') \
+                .filter(Link.url.contains(linkGroup.strip())).all(): 
+                    itemEditID = item.edition_id
+                    for edit in dbManager.session.query(Edition) \
+                        .filter(Edition.id == itemEditID):
+                            titleParse = edit.title
+                            idParse = edit.id
```
Apophenia, line 81: I'm not sure I 100% follow the parse logic here (especially with the linkGroup split) but I am guessing the main functionality here is matching log entries to titles and IDs from the database?

Apophenia (commented):
I think this looks good, but I think one more pass to improve script re-usability/future automation might be useful.

scripts/parseDataRequests.json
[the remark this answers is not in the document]
mitri-slory: When you "this file" are you referring to the `parseDataRequests.json` file? I don't intend to merge them in since they are test outputs like you said. I'll delete them before I merge them.

scripts/parseDownloadRequests.py
mitri-slory: That works for me. Makes sense to have these variable be easy to change on a global scale in the script.

scripts/parseDownloadRequests.py
mitri-slory: The general logic is that I'm trying to extract an UofM item object that has a link object with a `application/pdf` media type and has the fileID string from the key name of the UofM pdfs on S3 as a substring of its url attribute. The `linkGroup.strip()` method is to get rid of whitespace that appeared when I used regex to parse out the fileID from the log files. Once I have that item object, I use the item id to match with an edition id, so I can extract the edition title and an unique identifier which is the edition id in that instance.

Apophenia: Sounds great!

scripts/parseDownloadRequests.py
```diff
@@ -42,10 +41,10 @@ def main():
 
 def load_batch():
     paginator = s3_client.get_paginator('list_objects_v2')
-    page_iterator = paginator.paginate(Bucket= bucketName, Prefix= 'logs/946183545209/us-east-1/ump-pdf-repository/2024/03/13/')
+    page_iterator = paginator.paginate(Bucket= bucketName, Prefix=logPrefix)
     return page_iterator
 
-def parseInfo(logObject, requestRegex, referrerRegex, timeStampRegex, fileIDRegex):
+def parseInfo(logObject):
```
Apophenia, line 47: I think for the purposes of the script I don't mind whether these values are fed in as parameters or not; the only thing I felt strongly about was not hard-coding e.g. the prefix when the function was called & that has been changed. 

I think this looks OK too, but if we do abstract this code in the future (for example, to ingest books from other presses) we might want to think about how this function signature should be shaped.

Apophenia (approved):

PR #297, opened 2024-04-30
SFR-1966_FulfillURLUofM

mitri-slory:
This PR focuses on updating the UofM manifests with fulfill endpoints to replace the pdf and epub links within the manifest JSON file. All the UofM files will be affected for now even though a few are now Public Domain since reverting this change can be done at a later time once this ticket is resolved. Also, the commented out class definition at the top of the script has to do with one question I have. The question is if the S3Manger method, `putObjectInBucket`, should be utilized instead of the regular `s3_client.put_object` method since the S3Manger method is used when we currently put manifests in a S3 bucket like in the UofM process.

### DIFF
--- a/scripts/fulfillURLManifest.py
+++ b/scripts/fulfillURLManifest.py
@@ -1,0 +1,110 @@
+import boto3
+import json
+import os
+import logging
+from botocore.exceptions import ClientError
+from model import Link
+from managers import DBManager, s3
+
+s3_client = boto3.client("s3")
+
+bucketName = 'drb-files-qa'
+
+# class fulfillURLManifest():
+
+#     def __init__(self):
+
+#         # Connect to database
+#         self.generateEngine()
+#         self.createSession()
+
+#         # S3 Configuration
+#         self.s3Bucket = bucketName
+#         self.createS3Client()
+
+def main():
+
+    '''Replacing pdf/epub links in with fulfill urls in manifest JSON files'''
+
+    batches = load_batch()
+    for batch in batches:
+        for c in batch['Contents']:
+            currKey = c['Key']
+            metadataObject = s3_client.get_object(Bucket= bucketName, Key= f'{currKey}')
+            update_batch(metadataObject, bucketName, currKey)
+    
+
+def update_batch(metadataObject, bucketName, currKey):
+
+    dbManager = DBManager(
+        user=os.environ.get('POSTGRES_USER', None),
+        pswd=os.environ.get('POSTGRES_PSWD', None),
+        host=os.environ.get('POSTGRES_HOST', None),
+        port=os.environ.get('POSTGRES_PORT', None),
+        db=os.environ.get('POSTGRES_NAME', None)
+    )
+
+    dbManager.generateEngine()
+
+    dbManager.createSession()
+
+    metadataJSON = json.loads(metadataObject['Body'].read().decode("utf-8"))
+    
+    metadataJSON = linkFulfill(metadataJSON, dbManager)
+    metadataJSON = readingOrderFulfill(metadataJSON, dbManager)
+    metadataJSON = resourceFulfill(metadataJSON, dbManager)
+    metadataJSON = tocFulfill(metadataJSON, dbManager)
+
+    fulfillManifest = json.dumps(metadataJSON, ensure_ascii = False, indent = 6)
+
+    try:
+        return s3_client.put_object(Bucket=bucketName, Key='manifests/UofM/0472030132.json', \
+                            Body=fulfillManifest, ACL= 'public-read', \
+                            ContentType = 'application/json'
+            )
+    except ClientError as e:
+        logging.error(e)
+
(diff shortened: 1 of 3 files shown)


scripts/fulfillURLManifest.py
```diff
@@ -0,0 +1,110 @@
...
+            or i['type'] == 'application/epub+xml':
+                for link in dbManager.session.query(Link) \
+                    .filter(Link.url == i['href'].replace('https://', '')):
+                        i['href'] = f'http://127.0.0.1:5050/fulfill/{link.id}'
+                        return metadataJSON
+
+def tocFulfill(metadataJSON, dbManager): 
+    for i in metadataJSON['toc']:
+        if 'pdf' in i['href'] \
+            or 'epub' in i['href']:
```
Apophenia, line 98: Is the table of contents section formed differently from the links, resources, etc such that the logic here needs to be different?

scripts/fulfillURLManifest.py
```diff
@@ -0,0 +1,110 @@
...
+                for link in dbManager.session.query(Link) \
+                    .filter(Link.url == i['href'].replace('https://', '')):
+                        i['href'] = f'http://127.0.0.1:5050/fulfill/{link.id}'
+                        return metadataJSON
```
Apophenia, line 102: Once again just want to check my own understanding, and then pose an additional question: it looks like this code uses SQLAlchemy to filter the Links table, retrieving the link id of each entry with a matching URL existing in the Link url column. The process then returns an updated href that replaces the previous url with an updated fulfill link with the link id from the Links table.

If the logic is the same for links, resources, readingOrder, etc, is there any way to consolidate these lines of code (maybe by writing a helper function to encapsulate the logic that's the same)? It seems like these functions have very similar shapes and signatures and have some shared/identical lines of code that could perhaps be extracted to a more general function.

Apophenia (commented):
I've shared a few thoughts about what I think might make this code a little more re-usable if we do use a similar process for books from other presses in the future. Let me know what you think. I think the logic is solid.

scripts/fulfillURLManifest.py
```diff
@@ -0,0 +1,110 @@
...
+def main():
+
+    '''Replacing pdf/epub links in with fulfill urls in manifest JSON files'''
+
+    batches = load_batch()
+    for batch in batches:
+        for c in batch['Contents']:
+            currKey = c['Key']
+            metadataObject = s3_client.get_object(Bucket= bucketName, Key= f'{currKey}')
+            update_batch(metadataObject, bucketName, currKey)
```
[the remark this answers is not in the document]
mitri-slory: Yeah you're exactly right. For this ticket, I'm just focusing on manually updating the UofM manifests currently in S3. SFR-1967 will focus on future LA works and will deal with uncommenting the `storePDFmanifest` method.

scripts/fulfillURLManifest.py
mitri-slory: Yes it is since the toc section does not have a 'type' key value like the links, resources, etc sections. 

"toc": [{"href": "", "title": ""}]

scripts/fulfillURLManifest.py
```diff
@@ -0,0 +1,110 @@
...
+    
+    metadataJSON = linkFulfill(metadataJSON, dbManager)
+    metadataJSON = readingOrderFulfill(metadataJSON, dbManager)
+    metadataJSON = resourceFulfill(metadataJSON, dbManager)
+    metadataJSON = tocFulfill(metadataJSON, dbManager)
+
+    fulfillManifest = json.dumps(metadataJSON, ensure_ascii = False, indent = 6)
+
+    try:
+        return s3_client.put_object(Bucket=bucketName, Key=currKey, \
```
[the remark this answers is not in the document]
mitri-slory: I agree with your assessment Lyndsey. I'll work on adding putObjectInBucket as the method instead since I also thought it's better for error handling and future proofing the script.

scripts/fulfillURLManifest.py
mitri-slory: Your understanding of the code is correct. I was also thinking of consolidating these lines of code since they share similar logic and now with your feedback I will make a function that has the same logic.

scripts/fulfillURLManifest.py
```diff
@@ -4,24 +4,12 @@
...
-#     def __init__(self):
-
-#         # Connect to database
-#         self.generateEngine()
-#         self.createSession()
-
-#         # S3 Configuration
-#         self.s3Bucket = bucketName
-#         self.createS3Client()
-
```
mitri-slory: I changed my mind on using the putObjectInBucket method and chose to use the S3 client method put_object instead since these are QA manifests we're focusing on now so the extra error handling isn't needed for now and when the fulfill endpoints are integrated into future LA manifests in prod they will automatically use the putObjectInBucket method that's already in place.

Apophenia (approved):
Looks good, I also appreciate the more specific naming!

PR #298, opened 2024-05-15
SFR-1967_FulfillFutureLAManifests

mitri-slory:
This PR focuses on adding fulfill endpoints to future LA manifests after they have been ingested and clustered. A new flag called fulfill_limited_access has been added to prevent these LA works to be Read Online on the frontend before the process  addFulfillManifest is run.

### DIFF
--- a/config/production.yaml
+++ b/config/production.yaml
@@ -57,6 +57,10 @@ NYPL_LOCATIONS_BY_CODE: https://nypl-core-objects-mapping-qa.s3.amazonaws.com/by
 # API_CLIENT_ID and API_CLIENT_SECRET must be configured in secrets file
 NYPL_API_CLIENT_TOKEN_URL: https://isso.nypl.org/oauth/token
 
+# DRB API Credentials
+DRB_API_HOST: '0.0.0.0'
+DRB_API_PORT: '80'
+
 # GITHUB API Credentials
 # GITHUB_API_KEY must be configured in secrets file
 GITHUB_API_ROOT: https://api.github.com/graphql
--- a/config/production.yaml
+++ b/config/production.yaml
@@ -58,7 +58,7 @@ NYPL_LOCATIONS_BY_CODE: https://nypl-core-objects-mapping-qa.s3.amazonaws.com/by
 NYPL_API_CLIENT_TOKEN_URL: https://isso.nypl.org/oauth/token
 
 # DRB API Credentials
-DRB_API_HOST: '0.0.0.0'
+DRB_API_HOST: 'digital-research-books-beta.nypl.org'
 DRB_API_PORT: '80'
 
 # GITHUB API Credentials
--- a/config/qa.yaml
+++ b/config/qa.yaml
@@ -58,7 +58,7 @@ NYPL_LOCATIONS_BY_CODE: https://nypl-core-objects-mapping-qa.s3.amazonaws.com/by
 NYPL_API_CLIENT_TOKEN_URL: https://isso.nypl.org/oauth/token
 
 # DRB API Credentials
-DRB_API_HOST: '0.0.0.0'
+DRB_API_HOST: 'drb-qa.nypl.org'
 DRB_API_PORT: '80'
 
 # GITHUB API Credentials
--- a/config/qa.yaml
+++ b/config/qa.yaml
@@ -57,6 +57,10 @@ NYPL_LOCATIONS_BY_CODE: https://nypl-core-objects-mapping-qa.s3.amazonaws.com/by
 # API_CLIENT_ID and API_CLIENT_SECRET must be configured in secrets file
 NYPL_API_CLIENT_TOKEN_URL: https://isso.nypl.org/oauth/token
 
+# DRB API Credentials
+DRB_API_HOST: '0.0.0.0'
+DRB_API_PORT: '80'
+
 # GITHUB API Credentials
 # GITHUB_API_KEY must be configured in secrets file
 GITHUB_API_ROOT: https://api.github.com/graphql
--- a/config/sample-compose.yaml
+++ b/config/sample-compose.yaml
@@ -62,6 +62,11 @@ NYPL_API_CLIENT_PUBLIC_KEY: |
   EQIDAQAB
   -----END PUBLIC KEY-----
 
+  # DRB API Credentials
+  
+DRB_API_HOST: '0.0.0.0'
+DRB_API_PORT: '5050'
+
 # Bardo CCE API URL
 BARDO_CCE_API: http://sfr-bardo-copyright-development.us-east-1.elasticbeanstalk.com/search
 
--- a/config/sample-compose.yaml
+++ b/config/sample-compose.yaml
@@ -64,7 +64,7 @@ NYPL_API_CLIENT_PUBLIC_KEY: |
 
   # DRB API Credentials
   
-DRB_API_HOST: '0.0.0.0'
+DRB_API_HOST: drb_local_webapp
 DRB_API_PORT: '5050'
 
 # Bardo CCE API URL
--- a/processes/UofM.py
+++ b/processes/UofM.py
@@ -45,7 +45,7 @@ class UofMProcess(CoreProcess):
             UofMRec = UofMMapping(record)
             UofMRec.applyMapping()
             self.addHasPartMapping(record, UofMRec.record)
-            #self.storePDFManifest(UofMRec.record)
+            self.storePDFManifest(UofMRec.record)
             self.addDCDWToUpdateList(UofMRec)
             
         except (MappingError, HTTPError, ConnectionError, IndexError, TypeError) as e:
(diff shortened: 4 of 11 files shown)


processes/addFulfillManifest.py
```diff
@@ -0,0 +1,138 @@
...
+
+        return metadataJSON
+
+    def tocFulfill(self, metadataJSON, dbManager): 
+        for i in metadataJSON['toc']:
+            if 'pdf' in i['href'] \
+                or 'epub' in i['href']:
+                    for link in dbManager.session.query(Link) \
+                        .filter(Link.url == i['href'].replace('https://', '')):
+                            i['href'] = f'http://127.0.0.1:5050/fulfill/{link.id}'
```
Apophenia, line 117: Hi Dmitri, thanks for your work on this! This script will run on QA and production in order to add manifests on those servers in the future, right? If that is the case, I don’t think a hard coded URL to the dev environment will work since the processes will run remotely. I had to deal with the same issue of dynamically guiding the urls in many of the endpoints when adding the fulfill endpoints to the JSON (e.g. the work endpoint).

mitri-slory: Appreciate the feedback Lyndsey. You're right on a hard coded url not working in this case, so I'll update these urls to be more dynamic for the prod and QA environments.

processes/addFulfillManifest.py
```diff
@@ -0,0 +1,140 @@
...
+
+        batches = self.load_batch()
+        for batch in batches:
+            for c in batch['Contents']:
+                if c["LastModified"] > startTimeStamp.replace(tzinfo=timezone.utc):
+                    currKey = c['Key']
+                    metadataObject = self.getObjectFromBucket(currKey, self.s3Bucket)
+                    self.update_batch(metadataObject, self.s3Bucket, currKey)
+
+    def update_batch(self, metadataObject, bucketName, currKey):
```
mwbenowitz, line 61: I'd rename this function since it only deals with one manifest at a time (this implies it deals with the whole batch)

mwbenowitz (commented):
The functionality looks good, I left some comments about some specifics of the implementation that I think could be updated to be clearer and a bit more performant.

processes/addFulfillManifest.py
```diff
@@ -0,0 +1,140 @@
...
+            self.updateManifest(startTimeStamp)
+
+    def updateManifest(self, startTimeStamp=None):
+
+        '''Load batch of LA works from past 24 hours'''
+
+        batches = self.load_batch()
+        for batch in batches:
+            for c in batch['Contents']:
+                if c["LastModified"] > startTimeStamp.replace(tzinfo=timezone.utc):
```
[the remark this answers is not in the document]
mitri-slory: I'm a little confused on what you're suggesting. Are you suggesting to put this block of highlighted code into the `load_batches` method or something else?

processes/UofM.py
```diff
@@ -90,7 +90,7 @@ def addHasPartMapping(self, resultsRecord, record):
                     urlPDFObject,
                     'UofM',
                     'application/pdf',
-                    '{"catalog": false, "download": true, "reader": false, "embed": false}'
+                    '{"catalog": false, "download": true, "reader": false, "embed": false, "nypl_login": true}'
```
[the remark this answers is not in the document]
mitri-slory: Got an update from Elena and she says only the MIT Press books need to have downloads disabled. UofM books can  be downloaded and viewed. This block of code can be kept now.

processes/addFulfillManifest.py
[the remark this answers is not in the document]
mwbenowitz: When you call `paginator.paginate` I think you should be able to pass in a parameter to fetch only the objects that match this `LastModified` query which would a) eliminate the need for this check and b) speed up the process because you won't need to load every item from this press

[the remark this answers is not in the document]
mwbenowitz: fwiw `load_batch` could also be a generator function using `yield` if you want to experiment with that pattern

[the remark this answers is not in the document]
mitri-slory: Appreciate the clarification and I found a way to use the JSON query language JSEM to do this as well.

config/production.yaml
```diff
@@ -57,6 +57,10 @@ NYPL_LOCATIONS_BY_CODE: https://nypl-core-objects-mapping-qa.s3.amazonaws.com/by
 # API_CLIENT_ID and API_CLIENT_SECRET must be configured in secrets file
 NYPL_API_CLIENT_TOKEN_URL: https://isso.nypl.org/oauth/token
 
+# DRB API Credentials
+DRB_API_HOST: '0.0.0.0'
```
mwbenowitz, line 61: I believe this should be `digital-research-books-beta.nypl.org` Remember that these are URLs that clients will use to connect to our servers so they need to be a resolvable domain.

config/qa.yaml
```diff
@@ -57,6 +57,10 @@ NYPL_LOCATIONS_BY_CODE: https://nypl-core-objects-mapping-qa.s3.amazonaws.com/by
 # API_CLIENT_ID and API_CLIENT_SECRET must be configured in secrets file
 NYPL_API_CLIENT_TOKEN_URL: https://isso.nypl.org/oauth/token
 
+# DRB API Credentials
+DRB_API_HOST: '0.0.0.0'
```
mwbenowitz, line 61: Same comment as above but for QA

config/sample-compose.yaml
```diff
@@ -62,6 +62,11 @@ NYPL_API_CLIENT_PUBLIC_KEY: |
   EQIDAQAB
   -----END PUBLIC KEY-----
 
+  # DRB API Credentials
+  
+DRB_API_HOST: '0.0.0.0'
```
mwbenowitz, line 67: The default should be some version of localhost. `0.0.0.0` exposes the port to the world (rather than just your local machine) so this technically introduces a security risk while running it locally. Generally it's not exploitable but it's best practice to just avoid this

processes/addFulfillManifest.py
```diff
@@ -33,51 +25,42 @@ def __init__(self, *args):
 
         # S3 Configuration
         self.s3Bucket = os.environ['FILE_BUCKET']
+        self.host = os.environ['DRB_API_HOST']
+        self.port = os.environ['DRB_API_PORT']
+        self.objKey = 'manifests/UofM/'
```
mwbenowitz, line 30: I'd call this `prefix` or something. When its referenced in this class it gets confusing because in s3 keys usually refer to individual objects.

processes/addFulfillManifest.py
```diff
@@ -33,51 +25,42 @@ def __init__(self, *args):
...
         metadataJSONCopy = copy.deepcopy(metadataJSON)
         
-        metadataJSON = self.linkFulfill(metadataJSON, dbManager)
-        metadataJSON = self.readingOrderFulfill(metadataJSON, dbManager)
-        metadataJSON = self.resourceFulfill(metadataJSON, dbManager)
-        metadataJSON = self.tocFulfill(metadataJSON, dbManager)
+        metadataJSON = self.linkFulfill(self, metadataJSON)
+        metadataJSON = self.readingOrderFulfill(self, metadataJSON)
+        metadataJSON = self.resourceFulfill(self, metadataJSON)
+        metadataJSON = self.tocFulfill(self, metadataJSON)
```
mwbenowitz, line 63: Don't need to pass `self` as the first argument here

processes/addFulfillManifest.py
```diff
@@ -88,53 +71,51 @@ def update_batch(self, metadataObject, bucketName, currKey):
...
+                or 'epub' in toc['href']:
+                    for link in self.session.query(Link) \
+                        .filter(Link.url == toc['href'].replace('https://', '')):
+                            toc['href'] = f'http://{self.host}:{os.environ[self.port]}/fulfill/{link.id}'
                             link.flags['fulfill_limited_access'] = True
         return metadataJSON
 
     @staticmethod
-    def fulfillReplace(metadata, dbManager):
+    def fulfillReplace(self, metadata):
```
mwbenowitz, line 111: If we're passing self here this doesn't need to be a static method

mwbenowitz (commented):
Left a few more comments. Overall looking good but there are a few more tweaks that need to be made around variables, etc.

config/production.yaml
mitri-slory: This makes sense. Thanks for the correction.

config/qa.yaml
mitri-slory: I'll change it to the host name to `drb-qa.nypl.org`

config/sample-compose.yaml
mitri-slory: Appreciate the explanation.

processes/addFulfillManifest.py
mitri-slory: Agreed. I'll change the name.

processes/addFulfillManifest.py
mitri-slory: You're right. Forgot to change this.

processes/addFulfillManifest.py
mitri-slory: Thanks for catching this. I thought I removed static method decorators from all the methods.

Apophenia:
Thanks for making these changes! I think this looks good :)

PR #301, opened 2024-06-07
SFR-1980_DeprecateUTCNowMethod

kylevillegas93:
## Description 
This change deprecates the `datetime.utcnow()` method and replaces it with `datetime.now(timezone.utc).replace(tzinf=None)` based on the following [documentation](https://discuss.python.org/t/deprecating-utcnow-and-utcfromtimestamp/26221/2). 

The main difference between `datetime.utcnow()` and `datetime.now(timezone.utc)` is that the latter is timezone aware while the former is timezone unaware: 

`print(datetime.utcnow)` ---> `2024-06-07 19:07:33.968121`
`print(datetime.now(timezone.utc))` ---> `2024-06-07 19:07:42.528055+00:00`

This change only deprecates `utcnow()` and does not make the system timezone aware.  If we want to make the system timezone aware, I think that will be a larger change to ensure we are not comparing timezone unaware and aware datetime objects. Decisions like should we make audit information in the database also timezone aware will need to be made. 

## Testing
`make test`


Apophenia (commented):
Awesome, thanks for catching this. I think the `datetime.utcnow()` method definitely needs to be updated, and this looks like the best way to do it to retain compatibility with 3.9 & our current stack.

Apophenia (approved):
Thanks for all of your work on this -- looks great!

mitri-slory (approved):
Looks good Kyle! Thanks for taking the time to do this.

PR #302, opened 2024-06-12
HOTFIX_FixFulfillURLManifest

mitri-slory:
This hotfix intends to fix the typos found in the original format of the fulfill endpoints in the UofM manifests. The structure of it now works as intended with an example that was used by the web reader on the front end. The example was this: `https://drb-api-qa.nypl.org/fulfill/9350262`


kylevillegas93 (approved):
looks good

Apophenia:
I think this looks good! If we use these variables in the future (e.g. in integration tests we want to run locally) we might have to think about how to deal with the port in the URL building, but I think given that we are only ever going to run these scripts to generate the manifest files, this should work for QA and production.

PR #304, opened 2024-06-20
SFR_1987_RightStatusUofM

mitri-slory:
This PR focuses on adding the Rights attribute to the UofM books.


mappings/UofM.py
```diff
@@ -75,6 +70,40 @@ def formatIdentifiers(self):
...
+        if '|' in self.record.subjects:
+            subjectList = self.record.subjects.split('|')
+            newSubjectList = [f'{subject}||' for subject in subjectList] 
+            return newSubjectList
+        else:
+            subjectList.append(f'{self.record.subjects}||')
+            return subjectList
+    
+    def formatRights(self):
+        # Parse rights codes
```
kylevillegas93, line 86: I think the code is clear that we are parsing the rights codes. My rec is to remove it!

mappings/UofM.py
```diff
@@ -75,6 +70,40 @@ def formatIdentifiers(self):
...
+        rightsElements = self.record.rights.split('|') if self.record.rights else [''] * 5
```
kylevillegas93, line 87: Hmm - one comment that might be helpful is an example of what the rights can look like and why there is this pipe delimiter.

kylevillegas93: My recommendation is if self.record.rights is None then let's keep it as None otherwise we may have rights with empty strings.

mappings/UofM.py
```diff
@@ -75,6 +70,40 @@ def formatIdentifiers(self):
...
+        licenseMeta = ''
+        statementMeta = ''
+        print(rightsMetadata)
+
+        if rightsMetadata == 'in copyright':
+            licenseMeta = 'in_copyright'
+            statementMeta = 'In Copyright'
+        if rightsMetadata == 'public domain':
+            licenseMeta = 'public_domain'
+            statementMeta = 'Public Domain'
```
kylevillegas93, line 98: are there any other possible values for rightsMetadata?

mappings/UofM.py
```diff
@@ -75,6 +70,40 @@ def formatIdentifiers(self):
...
+            newSubjectList = [f'{subject}||' for subject in subjectList] 
+            return newSubjectList
+        else:
+            subjectList.append(f'{self.record.subjects}||')
+            return subjectList
+    
+    def formatRights(self):
+        # Parse rights codes
+        rightsElements = self.record.rights.split('|') if self.record.rights else [''] * 5
+        rightsMetadata = rightsElements[0]
```
kylevillegas93, line 88: rightsMetadata is a bit vague. is `rightsStatus` clearer?

mappings/UofM.py
```diff
@@ -75,6 +70,40 @@ def formatIdentifiers(self):
...
+        licenseMeta = ''
+        statementMeta = ''
```
kylevillegas93, line 90: Are there cases where we want the rights to be `UofM|||`?

mappings/UofM.py
```diff
@@ -16,6 +16,7 @@ def createMapping(self):
                 ('ISBN', '{0}|isbn'),
                 ('OCLC', '{0}|oclc')
             ],
+            'rights': ('DRB Rights Classification', '{0}||||'),
```
kylevillegas93, line 19: Is the rights classification specific to DRB or is it general? Could this just be Rights Classification?

mappings/UofM.py
```diff
@@ -75,6 +70,40 @@ def formatIdentifiers(self):
...
+        # Parse rights codes
+        rightsElements = self.record.rights.split('|') if self.record.rights else [''] * 5
+        rightsMetadata = rightsElements[0]
+        licenseMeta = ''
+        statementMeta = ''
+        print(rightsMetadata)
+
+        if rightsMetadata == 'in copyright':
+            licenseMeta = 'in_copyright'
+            statementMeta = 'In Copyright'
```
kylevillegas93, line 95: What are the differences between these two fields and what are they used for?

kylevillegas93 (changes requested):
Overall looks good! I just have a few questions and minor comments.

mappings/UofM.py
```diff
@@ -75,6 +70,40 @@ def formatIdentifiers(self):
...
+            subjectList.append(f'{self.record.subjects}||')
+            return subjectList
+    
+    def formatRights(self):
+        # Parse rights codes
+        rightsElements = self.record.rights.split('|') if self.record.rights else [''] * 5
+        rightsMetadata = rightsElements[0]
+        licenseMeta = ''
+        statementMeta = ''
+        print(rightsMetadata)
```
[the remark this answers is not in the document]
mitri-slory: I can do that

mappings/UofM.py
mitri-slory: I agree

mappings/UofM.py
mitri-slory: Ok I can leave a comment that does that.

mappings/UofM.py
mitri-slory: There are rights_statements and rights_date which are not necessary to have and are not currently available on Airtable.

mappings/UofM.py
mitri-slory: I agree that `rightsStatus` is clearer since we're only tracking if the books are in copyright or PD and no other rights metadata.

mappings/UofM.py
mitri-slory: No to reach the formatRights method, there will have to be a value in the Rights mapping so there won't be a case for UofM||||. To make this clearer, I'll delete those two lines and move this code block 

```
rightsObject = 'UofM|{}||{}|'.format(
                licenseMeta,
                statementMeta
            ) 

            return rightsObject
```

inside both conditional statements.

mappings/UofM.py
mitri-slory: This name is the name of the column in AIrtable with the rights classification. When converting the csv format of Airtbale into a json file, the name remains as a key which is used in this mapping to get the value of that key.

mappings/UofM.py
mitri-slory: The two fields are similar with the statement usually being a longer form version of the license.

mitri-slory: Both identidy the current rights status of a particular book.

mappings/UofM.py
```diff
@@ -75,6 +70,49 @@ def formatIdentifiers(self):
...
+        if '|' in self.record.subjects:
+            subjectList = self.record.subjects.split('|')
+            newSubjectList = [f'{subject}||' for subject in subjectList] 
+            return newSubjectList
+        else:
+            subjectList.append(f'{self.record.subjects}||')
+            return subjectList
+    
+    def formatRights(self):
+        # The pipe delimiter is to separate the Rights table attributes into this format:
```
kylevillegas93, line 86: Let's use a multi-line comment here

mappings/UofM.py
```diff
@@ -75,6 +70,49 @@ def formatIdentifiers(self):
...
+        # source|license|reason|statement|date 
+        # which makes it easy to place the right data into the columns when clustered
+
+        rightsElements = self.record.rights.split('|') if self.record.rights else [''] * 5
```
kylevillegas93, line 90: I'd say if the rights are None, let's just return None. so we can have 

```
if not self.record.rights: return None

rightsElements = self.record.rights.split('|')
```

mappings/UofM.py
```diff
@@ -75,6 +70,49 @@ def formatIdentifiers(self):
...
+        if rightsStatus == 'in copyright':
+            licenseMeta = 'in_copyright'
+            statementMeta = 'In Copyright'
+
+            rightsObject = 'UofM|{}||{}|'.format(
+                licenseMeta,
+                statementMeta
+            ) 
+
+            return rightsObject
```
kylevillegas93, line 102: let's make this single line: let's make this single line:

```
return 'UofM|{}||{}|'.format('in_copyright', 'In Copyright') 
```

mappings/UofM.py
```diff
@@ -75,6 +70,49 @@ def formatIdentifiers(self):
...
+        if rightsStatus == 'public domain':
+            licenseMeta = 'public_domain'
+            statementMeta = 'Public Domain'
+
+            rightsObject = 'UofM|{}||{}|'.format(
+                licenseMeta,
+                statementMeta
+            ) 
+
+            return rightsObject
```
kylevillegas93, line 113: Same as above: 

`return 'UofM|{}||{}|'.format('public_domain', 'Public Domain')

mappings/UofM.py
```diff
@@ -75,6 +70,49 @@ def formatIdentifiers(self):
...
+
+
```
kylevillegas93, line 115: Let's return None here if the right status is neither public domain nor in copyright?

mappings/UofM.py
mitri-slory: Sure

mappings/UofM.py
mitri-slory: Sure. I can do this and I'll delete the if conditional in the `applyFormatting `method:

```
if self.record.rights:
            self.record.rights = self.formatRights()
```

mappings/UofM.py
mitri-slory: I can do that

kylevillegas93 (approved):
Looks good - nice work!

processes/UofM.py
```diff
@@ -118,16 +141,29 @@ def storePDFManifest(self, record):
...
+                    record.has_part.insert(0, linkString)
+                    break
+
+                if 'public_domain' in record.rights:
+                    linkString = '|'.join([
+                        itemNo,
+                        manifestURI,
+                        source,
+                        'application/webpub+json',
+                        '{"catalog": false, "download": false, "reader": true, "embed": false'
```
kylevillegas93, line 162: Need a closing curly brace } here!

PR #305, opened 2024-06-26
SFR-2054_PubProjectSourceFieldMigration

mitri-slory:
This PR so far focuses on adding a new column named contract_source to the record and items PSQL tables since both tables have a source field already so it makes sense to have both source type fields in these tables. Once this PR is good to go it doesn't need to be approved and a comment saying that the PR is ready for database migration will suffice.  Approving this PR at that stage is unnecessary since I will then run the database migration in the QA environment and once the migration works the next half of this PR will be focused on updating the DRB API to allow the filtering of works based on contract_source. Those changes to the DRB API will be committed to this branch as well.

### DIFF
--- a/migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
+++ b/migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
@@ -1,4 +1,4 @@
-"""add contract source to records and items
+"""add publisher_project_source field to records and items
 
 Revision ID: 54e57fb2e1c6
 Revises: e46b30dd3ff5
@@ -17,10 +17,10 @@ depends_on = None
 
 
 def upgrade():
-    op.add_column('records', sa.Column('contract_source', sa.ARRAY, index=True))
-    op.add_column('items', sa.Column('contract_source', sa.ARRAY, index=True))
+    op.add_column('records', sa.Column('publisher_project_source', sa.ARRAY(sa.Unicode), index=True))
+    op.add_column('items', sa.Column('publisher_project_source', sa.ARRAY(sa.Unicode), index=True))
 
 
 def downgrade():
-    op.drop_column('records', 'contract_source')
-    op.drop_column('items', 'contract_source')
+    op.drop_column('records', 'publisher_project_source')
+    op.drop_column('items', 'publisher_project_source')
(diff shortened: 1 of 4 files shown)


migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
```diff
@@ -0,0 +1,26 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '54e57fb2e1c6'
+down_revision = 'e46b30dd3ff5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('records', sa.Column('contract_source', sa.ARRAY, index=True))
```
kylevillegas93, line 20: Should we name this `contract_sources` since there can be many?

migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
```diff
@@ -0,0 +1,26 @@
...
+    op.add_column('items', sa.Column('contract_source', sa.ARRAY, index=True))
```
kylevillegas93, line 21: Are we pretty positive that this will be a list of strings and we won't need to store any additional information besides the name of the contract source? 

Also can this be nullable? We may have to specify in the nullable param. 

https://alembic.sqlalchemy.org/en/latest/ops.html

Apophenia:
Thanks so much for your work on this, @mitri-slory!

I just talked to Will and I think we're going to hold on merging this in. We're planning to schedule a meeting early next week (both you and @kylevillegas93) to talk about the requirements for analytics and reporting (which is ultimately what this field pertains to rather than the general user experience of DRB). Maybe once we better understand the business needs here (particularly if this field will be used to populate an analytics or reporting dashboard for publishers) we'll have a better idea of what role this field will play, and whether we should rename it.

migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
mitri-slory: You're right and I'll make that change.

migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
mitri-slory: If I add the code `sa.String` as a parameter to the `sa.ARRAY` like `sa.ARRAY(sa.String)`, then the list_type (the parameter that is being filled in for `sa.ARRAY`) will be guaranteed to be string values. Also, adding a column like in the other database migrations will be nullable by default because the nullable parameter is only used when altering a table or creating a table which the records and items table already allow null values.

PR #306, opened 2024-07-18
Sfr-2054_PubSourceFieldMigration

mitri-slory:
This PR so far focuses on adding a new column named publisher_project_source to the record and items PSQL tables since both tables have a source field already so it makes sense to have both source type fields in these tables.

### DIFF
--- a/migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
+++ b/migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
@@ -1,4 +1,4 @@
-"""add contract source to records and items
+"""add publisher_project_source field to records and items
 
 Revision ID: 54e57fb2e1c6
 Revises: e46b30dd3ff5
@@ -17,10 +17,10 @@ depends_on = None
 
 
 def upgrade():
-    op.add_column('records', sa.Column('contract_source', sa.ARRAY, index=True))
-    op.add_column('items', sa.Column('contract_source', sa.ARRAY, index=True))
+    op.add_column('records', sa.Column('publisher_project_source', sa.ARRAY(sa.Unicode), index=True))
+    op.add_column('items', sa.Column('publisher_project_source', sa.ARRAY(sa.Unicode), index=True))
 
 
 def downgrade():
-    op.drop_column('records', 'contract_source')
-    op.drop_column('items', 'contract_source')
+    op.drop_column('records', 'publisher_project_source')
+    op.drop_column('items', 'publisher_project_source')
--- a/migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
+++ b/migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
@@ -17,8 +17,8 @@ depends_on = None
 
 
 def upgrade():
-    op.add_column('records', sa.Column('publisher_project_source', sa.ARRAY(sa.Unicode), index=True))
-    op.add_column('items', sa.Column('publisher_project_source', sa.ARRAY(sa.Unicode), index=True))
+    op.add_column('records', sa.Column('publisher_project_source', sa.Unicode, index=True))
+    op.add_column('items', sa.Column('publisher_project_source', sa.Unicode, index=True))
 
 
 def downgrade():
--- a/migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
+++ b/migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
@@ -1,0 +1,26 @@
+"""add contract source to records and items
+
+Revision ID: 54e57fb2e1c6
+Revises: e46b30dd3ff5
+Create Date: 2024-06-26 15:04:24.092549
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '54e57fb2e1c6'
+down_revision = 'e46b30dd3ff5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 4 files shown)


migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def upgrade():
+    op.add_column('records', sa.Column('publisher_project_source', sa.ARRAY(sa.Unicode), index=True))
+    op.add_column('items', sa.Column('publisher_project_source', sa.ARRAY(sa.Unicode), index=True))
+
+
+def downgrade():
+    op.drop_column('records', 'publisher_project_source')
+    op.drop_column('items', 'publisher_project_source')
```
kylevillegas93, line 26: nit: let's add a newline at the end of the file.

migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = '54e57fb2e1c6'
+down_revision = 'e46b30dd3ff5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('records', sa.Column('publisher_project_source', sa.ARRAY(sa.Unicode), index=True))
+    op.add_column('items', sa.Column('publisher_project_source', sa.ARRAY(sa.Unicode), index=True))
```
kylevillegas93, line 21: For contract source, I understood that there could be multiple sources but in my head I thought that 
the project source would be based on these Airtable values. Is that correct? If so, should this just be a string rather than an array of strings or is true that an item/record could be potentially provided by more than 1 backlist project source? 

Screenshot 2024-07-19 at 9 24 29 AM

kylevillegas93: And sorry 1 last question! We need the column in both tables because we need to go from manifest/pdf access log -> Item Links/Items -> Editions -> Records to populate the report data?

migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
mitri-slory: Ok I can do that.

migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
mitri-slory: I'm thinking the same thing that the Publisher project source is based on those Airtable values you posted, but I just wanted to get some insight if we should change this field to a string field rather than a array field. What are your thoughts on this @Apophenia?

mitri-slory: That wasn't my intention for adding the column to both tables. I did it since both table have a source field so it would be good organization to have the other source field in those tables. Also, the process of getting the log data to reporting would be this: manifest/pdf access log -> Item Links/Items -> Editions -> Works. Records remain in the backend and they don't appear in the API, so there is no view or download count for the records. The records are created to be clustered and create the works, editions, items, links, etc.

Apophenia: For question 1: I think in discussion with Will we've established that it can be a single value. I guess my understanding was that the value did not need to be granular enough to represent e.g. reverted rights, but all of those would just be "UMass Press Backlist," but at the same time, if we used the value in this column, we could still include "UMass Press Reverted Rights," for example, by drawing a report composed of both "UMass Press Backlist" and "UMass Press Reverted Rights." Is this worth flagging with product to clarify?

Apophenia: For question 2: I feel like we worked out with Mike that the value couldn't just be represented on the record (but would need to additionally be on the edition or item?) but now I unfortunately can't remember the logic by which we arrived there.

mitri-slory: I think keeping it as single value for now works so I'll change the type to a string value. I don't think this is worth flagging to product about. What do you think @kylevillegas93?

mitri-slory: I'll tag Mike in this PR just so we can clarify if we need to add this new field to any other table besides records and items.

mitri-slory: @mwbenowitz

kylevillegas93: Just asked for clarification - https://nypl.slack.com/archives/C04NNMPP3D0/p1721412191766289

mitri-slory: That's good to know we can make it a string value now.

mitri-slory: Also, @Apophenia Mike responded to our questions about the field being in the records and items table and he said "From a technical perspective I don’t think it needs to be anywhere besides those two. As long as you’re sure that you can grab it from the Item table easily to populate it where you need it, that should be good!" The items table is enough to get our data from the S3 logs so we should be good to go with having the field be in just the records and items table.

Apophenia: That makes sense to me @mitri-slory, thanks for checking!

Apophenia:
Looks good to me too! thanks!

PR #308, opened 2024-07-23
SFR-2055_PubProjectSourceUofM

mitri-slory:
This PR is intent out filling out the publisher_project_source field for current and future UofM books. After this is approved, I will ingest and cluster the UofM books so the current books have a non null value in this field.

### DIFF
--- a/mappings/UofM.py
+++ b/mappings/UofM.py
@@ -25,6 +25,7 @@ class UofMMapping(JSONMapping):
         self.record.has_part = []
         self.record.spatial = 'Michigan'
         self.record.source = 'UofM'
+        self.record.publisher_project_source = 'UofMichigan Press'
 
         if self.record.authors:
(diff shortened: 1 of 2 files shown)


mappings/UofM.py
```diff
@@ -25,6 +25,7 @@ def applyFormatting(self):
         self.record.has_part = []
         self.record.spatial = 'Michigan'
         self.record.source = 'UofM'
+        self.record.publisher_project_source = 'UofMichigan Press'
```
kylevillegas93, line 28: Should this be UofMichigan Backlist to match the project name?

mitri-slory: The name is the name of the UofM table on Airtable so it's fine the way it is I believe.

kylevillegas93 (approved):

PR #309, opened 2024-07-23
SFR-1935 Add editionID validation

kylevillegas93:
## Description
- After adding better error handling, the logs now show that the editionID that was being passed in was a string and not an integer
- This change ensures the path param editionID is a valid integer before querying the database. If not, it returns a 400 response back to the client

## Testing
`make test`

### DIFF
--- a/api/blueprints/drbEdition.py
+++ b/api/blueprints/drbEdition.py
@@ -22,6 +22,9 @@ def editionFetch(editionID):
                 or current_app.config['READER_VERSION']
             filteredFormats = APIUtils.formatFilters(terms)
 
+            if not isValidEditionID(editionID):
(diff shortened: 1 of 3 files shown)


api/blueprints/drbEdition.py
```diff
@@ -39,3 +42,7 @@ def editionFetch(editionID):
     except Exception as e: 
         logger.error(e)
         return APIUtils.formatResponseObject(500, RESPONSE_TYPE, { 'message': 'Unable to fetch edition with id {}'.format(editionID) })
+
+
+def isValidEditionID(editionID) -> bool:
+    return isinstance(editionID, int) or (isinstance(editionID, str) and editionID.isdigit())
```
mitri-slory, line 48: This is a good method to add to ensure the editionID is a integer even in string format.

PR #311, opened 2024-07-25
SFR-2065: Adding more logging to proxy API

kylevillegas93:
## Description
- Adding more logging to understand why proxying through S3 PDFs fails
- When we retrieve metadata about the S3 url it returns 403
- When we retrieve metadata about the cleaned S3 url it returns 400 and the cleaning process malforms the tokens
- @jackiequach had a great callout and we should just not proxy to S3. 

## Testing
`make test`

### DIFF
--- a/api/blueprints/drbUtils.py
+++ b/api/blueprints/drbUtils.py
@@ -81,7 +81,7 @@ def getProxyResponse():
                     urlParts.scheme, urlParts.netloc, cleanUrl
                 )
         else:
-            logger.warn(f'Unable to proxy clean URL {cleanUrl}')
+            logger.warn(f'Unable to proxy clean url {cleanUrl}')
             cleanUrl = proxyUrl
             break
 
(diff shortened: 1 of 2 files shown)


api/blueprints/drbUtils.py
```diff
@@ -51,30 +57,31 @@ def totalCounts():
...
 
         statusCode = headResp.status_code
-        print(statusCode, cleanUrl)
+        
+        logger.info(f'HEAD response from {cleanUrl} returned status code {statusCode}')
+
         if statusCode in [200, 204]:
             break
         elif statusCode in [301, 302, 303, 307, 308]:
-            print(headResp.headers)
```
Apophenia, line 69: Do the headers not need to be logged, or are they captured somewhere else?

kylevillegas93: Also the print statement logs are hard to find in new relic because they don't have any of the other log metadata :(

mitri-slory (approved):
These are good updates to introduce more logging to this API.

PR #314, opened 2024-07-26
SFR-1951: Implement call to Worldcat to retrieve OCLC number

Apophenia:
This is the first part of a number of PRs for [SFR-1951](https://newyorkpubliclibrary.atlassian.net/browse/SFR-1951). This PR does not currently retrieve an OCLC number, but about 10,000 OCLC numbers for a given keyword query, so that refinement will come in an upcoming PR!

**Things this PR currently adds:**

1. Authentication for the OCLC v2 Search APIs via token retrieval
2. A function that queries the OCLC v2 Search API when provided a structured query string

**Main things I am still looking to build out in this PR:**
1. **Token refresh:** Unless someone sees a good reason now to, I think we should actually use a similar system to Project R and store an OCLC token in the environment, and refresh that token regularly. Right now, the code requests a new token from OCLC for each request. If it actually did this while running ingest processes of e.g. Hathi materials, we'd quickly cap our daily request limit if the token requests are counted against our daily request limit. It looks like each token lasts for about 20 minutes, so we could significantly cut down the number of requests by reusing the token over the interval.
3. **Unit tests:** new functions are not unit tested (yet)
4. **OCLC-number-centric manager refactor:** the OCLC manager instantiates with an OCLC number, which made sense when this manager was always invoked after retrieving OCLC numbers from Classify. However, we'll now be utilizing catalog manager functions before we have an OCLC number to instantiate the manager with. Since we no longer have a separate set of credentials or a separate service in Classify, it seems like we could benefit from refactoring the OCLC manager to behave as a generic manager where OCLC numbers are passed to methods if needed. 
5. **Query mapping:** As with token refresh, I don't think I fully captured the work scoped in the TAD in the tickets I made. Rhere's more work to be done in translating the data we get from the ingest pipeline to something we can send to OCLC's query service, since they require a specific format when querying. After doing some more preliminary testing, the search function casts such a wide net that we will definitely need a more specific query than generic keyword searching. Something like this may or may not already exist in DRB (probably in the Classify manager) so I will have to do some more work to figure out the scope of this - it may be that a simple title/author mapping will suffice, or we might need something more detailed.
6. **Make integration tests run in Github Actions** - figure out how to configure Github Actions so that the new integration tests can access the environment variables & run.

[SFR-1951]: https://newyorkpubliclibrary.atlassian.net/browse/SFR-1951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


managers/oclcCatalog.py
```diff
@@ -27,6 +35,47 @@ def queryCatalog(self):
...
+            )
+        except (Timeout, ConnectionError):
+            print(f'Failed to query {bibsEndpoint} with query {query}')
+            return self.queryCatalog()
+
+        if bibsResp.status_code != 200:
+            print(f'OCLC Catalog Request failed with status {bibsResp.status_code}')
+            return None
+
+        return bibsResp.json()
```
Apophenia, line 81: I went ahead and returned the dict from json here. This is asymmetric with the other function returning data in the file, but that function hands over to an XML parser, whereas we just want the dict values.

managers/oclcCatalog.py
```diff
@@ -27,6 +35,47 @@ def queryCatalog(self):
...
+
+    def getToken(self):
+        try:
+            authResp = requests.post(
+                self.OCLC_AUTH_URL,
+                headers={ 'Content-Type': 'application/x-www-form-urlencoded'},
+                auth=(self.oclcClientID, self.oclcClientSecret),
+            )
+        except (Timeout, ConnectionError):
+            print(f'Failed to retrieve token from {self.OCLC_AUTH_URL}')
```
Apophenia, line 49: Should we use the logger in favor of print statements, and if so should I go ahead and refactor the file to forward all of these to the logger?

kylevillegas93: Hmm - ya I think let's add a logger to this class!

tests/integration/test_oclc_catalog_manager.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+from config.loadEnv import loadEnvFile
+from managers import OCLCCatalogManager
+
+class TestOCLCCatalogManager:
+    @pytest.fixture
+    def testInstance(self):
+        loadEnvFile('sample-compose', fileString='config/local.yaml')
```
kylevillegas93, line 8: Should we point this to local-compose?

tests/integration/test_oclc_catalog_manager.py
```diff
@@ -0,0 +1,18 @@
...
+        # TODO: In the future the catalog manager shouldn't take the OCLC no
+        return OCLCCatalogManager(1)
+
+    def testOCLCv2Auth(self, testInstance):
+        testInstance.getToken()
+        assert len(testInstance.token) > 10
+        assert testInstance.tokenExpiresAt is not None
+
```
kylevillegas93, line 16: I think we can remove this test as this should be handled by the other manager/integration tests

kylevillegas93 (approved):
Looks good! Let's get this merged in on Monday morning.

tests/integration/test_oclc_catalog_manager.py
kylevillegas93: Done!

tests/integration/test_oclc_catalog_manager.py
kylevillegas93: Done!

tests/integration/test_oclc_catalog_manager.py
Apophenia: Ah sorry, I didn't see this comment but thank you for taking care of this!

PR #315, opened 2024-07-26
NO-REF: Refactor OCLC Catalog Manager

kylevillegas93:
## Description
- This simple refactor passes the oclc no as a param to query the catalog rather than a param to instantiate the manager class

## Testing
`make test`

### DIFF
--- a/tests/unit/test_oclcCatalog_process.py
+++ b/tests/unit/test_oclcCatalog_process.py
@@ -69,8 +69,8 @@ class TestOCLCCatalogProcess:
 
         testInstance.processCatalogQuery('{"oclcNo": 1, "owiNo": 1}')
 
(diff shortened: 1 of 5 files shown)


tests/unit/test_oclcCatalog_process.py
```diff
@@ -19,6 +19,7 @@ def testInstance(self, mocker):
         class TestCatalogProcess(CatalogProcess):
             def __init__(self, process, customFile, ingestPeriod):
                 self.statics = {}
+                self.oclcCatalogManager = mocker.MagicMock()
```
Apophenia, line 22: great!

Apophenia (approved):
Thanks so much for making these changes. I think there's probably further space for refactoring the managers and their instantiation within the processes, but this alone is a great refactor.

PR #316, opened 2024-07-30
SFR-2079: Integration test make command

kylevillegas93:
## Description
- Adds integration test make command

## Testing
`make unit; make integration`

## TODO
- We will tackle integration tests and CI in a later phase of tech debt but not as part of the OCLC work.


Apophenia (approved):
Thanks for these changes, looks good!

PR #322, opened 2024-08-01
SFR-2070: Fix to Local Cluster Process

mitri-slory:
This PR intends to fix the ElasticConnection error that occurred when running the Cluster Process locally with the local-compose and sample-compose yaml files. The solution was pretty simple fortunately and it involved changing the host name for the Elasticsearch host from the name of the ES docker bucket to 'localhost.' All the host names for the services in the sample-compose file were updated as a result of this change to avoid these connection errors for not just ES but for all of the services present.


kylevillegas93 (approved):

Apophenia:
Perfect :)

PR #323, opened 2024-08-05
NO-REF: Removing aggregation result print statement

kylevillegas93:
## Description
- This log resulted in quite a few print statements that looked like bucket []


Apophenia (approved):
Sounds good - if we decide this isn't transparent enough maybe we can add some more specific logging in the future.

mitri-slory (approved):
Looks good to me and I agree with Lyndsey's assessment.

PR #324, opened 2024-08-05
NO-REF: Remove index from adding new column migration

mitri-slory:
This PR removes the index from the migration to add the publisher_project_source field to the records and items table due to deadblocks occurring during database migration in the production environment.

### DIFF
--- a/migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
+++ b/migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
@@ -17,8 +17,8 @@ depends_on = None
 
 
 def upgrade():
-    op.add_column('records', sa.Column('publisher_project_source', sa.Unicode, index=True))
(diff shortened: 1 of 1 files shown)


migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
```diff
@@ -17,8 +17,8 @@
 
 
 def upgrade():
```
kylevillegas93, line 19: let's change the file name to `54e57fb2e1c6_add_publisher_project_source_to_records_and_items.py

kylevillegas93 (approved):

migrations/versions/54e57fb2e1c6_add_contract_source_to_records_and_items.py
mitri-slory: Ok I can do that

PR #326, opened 2024-08-06
NO-REF: Release v0.13.1

kylevillegas93:
## Description 
- Release v0.13.1


mitri-slory (approved):
Looks good and I think this makes sense to create a PR referencing a backend version we're releasing rather than directly pushing it to main. We just need to remember to make a tag after this is approved that will reference the version number which you said you will do so thank you for that.

PR #327, opened 2024-08-06
SFR-2080: Add auxiliary functions for query building

Apophenia:
This pull request adds auxiliary functions and tests for building a search query to pass to the OCLC search in the OCLC Catalog Manager. 

I have added these functions and tests to the Classify Manager because the logic that these functions will replace currently "lives in" the Classify process and is governed by the Classify manager. However, leaving the logic in these two separate places will result in the Classify process calling functionality from two different managers, and potentially also the managers calling functionality from each other.

In short, I don't know if this is the final home of this functionality, or if the functions should have a different signature and operate on the properties of the record the manager is instantiated with. I think as we build more functionality bridging the gap between the new and old OCLC functionality, the way to refactor the Classify manager and the Classify process will become more apparent, but I'm open to suggestions about improvements to be made at this stage.


managers/oclcClassify.py
```diff
@@ -67,6 +67,27 @@ def getClassifyResponse(self):
 
         return self.parseXMLResponse()
 
+    def generate_identifier_query(self):
+        identifier_map = { "isbn": "bn",
+                          "issn": "in",
+                          "oclc": "no"}
```
Apophenia, line 73: We could add other identifiers here, like lccn, which we store but don't currently use (we would need to add this 
"upstream" as well). However, I don't see a reason to do this unless we think it would improve the clustering process.

managers/oclcClassify.py
```diff
@@ -67,6 +67,27 @@ def getClassifyResponse(self):
...
+    def generate_search_query(self):
+        """Parses the received data and generates an OCLC search query based either
+        on an identifier (preferred) or an author/title combination.
+        """
+        if self.identifier and self.identifierType:
+            return self.generate_identifier_query()
+        elif self.title and self.author:
+            return self.generate_author_title_query()
+        else:
+            print('Cannot search for work without identifier or title/author')
```
Apophenia, line 88: The function I am adapting has a print statement like this, but I do wonder how much value it adds - maybe it would be better to call a logger and pass some information from the record? Maybe this is best done when refactoring the manager as a whole.

kylevillegas93: I think let's remove and ensure the function that handles the ClassifyError is properly logging the error. 

I think the error message log should include the record id or uuid so we can reference it.

Apophenia: I'm not 100% sure the ClassifyManager retains a reference to the record at this point because at this point I believe one record could have fired off potentially several different requests to Classify/OtherEditions based on the identifiers present. (For example, while we can find which edition was created from which record, I don't think we have a record of which DCDW record passed an OCLC number to rabbitmq to grab the OCLC single catalog record... but I could be wrong/hope I'm wrong! I removed the print statement and I'll double check that assumption.

mitri-slory (approved):
The functions look good to me and the rest of the code looks ready to go.

PR #328, opened 2024-08-08
SFR-2098: Database Migration In Docker Environment

mitri-slory:
This PR focuses on having the local database in the Docker containers always being the latest version of the PSQL database when the docker compose file is run.


kylevillegas93 (approved):
Nice!! Looks great!

PR #331, opened 2024-08-09
SFR-2119: Adding works API error handling

kylevillegas93:
## Description
- Adds error handling to works API
- Refactors both works and editions API to use snake case

## Testing
`make unit`


mitri-slory (approved):
The refactoring of the works API looks good here.

PR #343, opened 2024-08-15
SFR-2122: Adding collection GET API error handling

kylevillegas93:
## Description
- Adding error handling to the collection endpoints that are called from the frontend

## Testing
`make test`


mitri-slory (approved):
Good catch to add error handling to the GET request for collections.

PR #344, opened 2024-08-19
Draft: OCLC Refactor in progress

Apophenia:
**Changed:**
- Created utility file with some auxiliary functions from managers
- Changed utility file functions to use snake case
- Define exceptions and tighten error handling in OCLC manager class (some exceptions were previously referred to but never thrown or defined, I changed the code to match the documentation rather than the other way around)
- Remove OCLC manager, OCLC manager tests, OCLC mapping, and OCLC mapping tests

**TODOS: Refactoring the manager to not instantiate a full manager object for each dispatch to OCLC**
- [x] Create separate test file for newly created catalog utilities and move tests from Classify manager tests into it
- [x] To be persisted between auxiliary functions: identifier, identifier type (e.g. OCLC, ISBN, etc), title, author
- [ ] Find a place to persist "start" or remove entirely in refactor - this seems to be used only for offset when calling OCLC for "additional identifiers" (which is actually a function that finds more related OCLC numbers)
- [ ] Make sure Other Editions endpoint has error case for submitting a bad OCLC number vs an empty response (Classify distinguished between these and I think if the new Other Editions endpoint does too, we will want to deal with that)
- [ ] Since we are replacing 1 call with two, it's not clear where we should limit the number of edition records (on the initial search or on the second?). Quantity or where to cut off calls that return really large amounts of data feels like a common theme in this refactor. Not sure if "summary=false" is relevant on any of our calls either.
- [x] ~~To be investigated further: it looks like calls to Classify may not have counted against our daily limit of calls to OCLC's search API (which I think is capped at something like 300k or 500k). I am guessing that Other Editions calls will probably come out of the same "pool" and we should be incrementing any OCLC GET calls we make via the existing mechanism so that we do not end up with low estimates of our calls. (I haven't investigated this fully but this might also present a problem if that incrementing only happens in the Catalog process, but calls from the Classify/"editioning" process now count against our daily OCLC cap as well)~~ This is actually incremented when the processing loop in the Classify process calls "frbrizeRecord" and always has been, so Classify has always been incrementing this internal counter. What we actually need to worry about is whether all of our calls to Other Editions are incrementing it, too.
- [x] Finish structuring utility file so that it can be called from processes
- [ ] The old Classify process actually created DCDW (records table) entries based on its responses from Classify, which I did not know (I previously thought that it received the OCLC records returned from its rough equivalent of brief bibs, and then sent those records' OCLC numbers to the rabbitmq queue - but it looks instead like the metadata processing function actually directly created DCDW records directly. I am still figuring out how to best replicate this part of the process, but I went ahead and deleted the mapping which will no longer be useful to us.

**Functionality removed that was specific to Classify and does not have a 1:1 replacement:**
- The Classify manager had code for dealing with a single item or multiple item response. Is that possible in any of the 2.0 endpoints? (My impression is that thankfully even a 1 item response will return structured as a list?)


Apophenia:
This is unfortunately still in a pretty unfinished state as I leave for vacation. Most of the adjustments to the Classify process are pseudocode and comments. But the things that need to be taken care of, on a high level, are as follows:

1) The old Classify process worked by retrieving a set of possibly related books from the OCLC Classify endpoint, searching for more OCLC numbers possibly related to that book, and then creating a record based on the information retrieved from Classify. I didn't realize the degree to which this relied on a custom XML mapping until I more closely audited the code, and I think creating a new mapping for brief-bibs or bibs is out of scope for me to attempt tonight before I am OOO. It also looks like an "owi" is a consistent part of the data being passed around -- I'm not sure what this means, but it might be an OCLC Work Number, which we would have gotten from Classify but do not get back from brief-bibs. (It might be available in bibs but not sure!)

2) After any record is persisted from DCDW, the additional OCLC numbers retrieved should be passed to `fetch_oclc_catalog_records` to add them to the lookup queue. However, this function also takes an owi number.

3) I wasn't able to test the code changes I did make much, and I think it might be necessary to adjust the numbers of records we save (or don't save, etc) by an order of magnitude. We might need to keep tons of entries from the original bibs call and then cut off the other-editions results fairly early, or vice versa. 

I'm sorry to not be able to leave this in a more complete state or with passing tests, but I will be around early tomorrow if there are any big questions I can help answer.

Apophenia:
Closing this as I think all of the remaining concerns were addressed in pull requests made in between this PR & now -- changing things over to new OCLC APIs (e.g. catalog calls -> metadata bibs endpoint) remains, but attempts at that in this branch are largely irrelevant.

PR #345, opened 2024-08-19
NO-REF: Fix edition ID API Param

kylevillegas93:
## Description
- Flask param needs to match python function param name

## Testing
`http://localhost:5050/edition/1`


Apophenia (approved):
Looks good! Thanks for the fix!

PR #346, opened 2024-08-19
SFR-2133: Adding UUID validation to APIs

kylevillegas93:
## Description
- The database client does expect a valid uuid before querying, otherwise it will throw an error making it look like there's something wrong on the server
- This change adds validation to check the uuid is valid

## Testing
`make unit`


mitri-slory (approved):
Good catch to make sure the urls are valid before they're queried.

PR #349, opened 2024-08-21
SFR-2049: Implementing country-level analytics report

kylevillegas93:
## Description
- This change refactors a bit of how we aggregate the data
- We aggregate the data in one place and the reports then can use the aggregated data
- Change also implements the country-level report

## Testing 
`python3 analytics/upress_reporting/runner.py "2024-08-01 to 2024-08-02"`


mitri-slory (approved):
This implementation looks great. Good to see this update to the analytics.

PR #350, opened 2024-08-21
SFR-2135: Implementing total usage analytics report

kylevillegas93:
## Description
- Implements total usage analytics report

## Testing
`python3 analytics/upress_reporting/runner.py "2024-08-01 to 2024-08-02"`

### DIFF
--- a/analytics/upress_reporting/models/reports/total_usage.py
+++ b/analytics/upress_reporting/models/reports/total_usage.py
@@ -1,6 +1,41 @@
+import csv
+
 from models.reports.counter_5_report import Counter5Report
 
 
 class TotalUsageReport(Counter5Report):
-    def __init__(self, publisher):
-        super().__init__(publisher)
+    def __init__(self, *args):
+        super().__init__(*args)
+    
+    def build_report(self, events):
+        print("Building total usage report...")
+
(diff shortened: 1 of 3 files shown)


mitri-slory (approved):
This looks good and I do have a question about one part of this code, but otherwise this looks ready to go.

analytics/upress_reporting/models/reports/total_usage.py
```diff
@@ -1,6 +1,41 @@
...
+                for key, value in header.items():
+                    writer.writerow([key, value])
+                writer.writerow([])
+                writer.writerow(columns)
+                for title in final_data:
+                    writer.writerow(title.values())
+
+            print("Total usage report generation complete!")
+        else:
+            print("No events found in reporting period!")
```
mitri-slory, line 30: Why are print statements used here instead of logging statements? This question applies to your previous PR as well, but I didn't notice it until now.

kylevillegas93: Since this code is run locally, we just want to print to the console!

mitri-slory: Ok that makes sense

PR #351, opened 2024-08-21
SFR-2149: Implementing OCLC Classify Process with Editions Endpoints

kylevillegas93:
## Description
- Implements classify process using the OCLC editions endpoints

## Testing
Tested by calling classify_record_by_metadata_v2 - looks good!

### DIFF
--- a/managers/oclcClassify.py
+++ b/managers/oclcClassify.py
@@ -67,30 +67,6 @@ class ClassifyManager:
 
         return self.parseXMLResponse()
 
-    # TODO: can we query by everything?
-    def generate_identifier_query(self):
-        identifier_map = { 
-            "isbn": "bn",
-            "issn": "in",
-            "oclc": "no"
-        }
-
-        return f"{identifier_map[self.identifierType]}: {self.identifier}"
-
-    def generate_author_title_query(self):
-        return f"ti:{self.title} au:{self.author}"
-
-    def generate_search_query(self):
-        """Parses the received data and generates an OCLC search query based either
-        on an identifier (preferred) or an author/title combination.
-        """
-        if self.identifier and self.identifierType:
-            return self.generate_identifier_query()
-        elif self.title and self.author:
-            return self.generate_author_title_query()
-        else:
-            raise ClassifyError('Record lacks identifier or title/author pair')
-
     def generateQueryURL(self):
         """Parses the received data and generates a Classify query based either
         on an identifier (preferred) or an author/title combination.
--- a/managers/oclcClassify.py
+++ b/managers/oclcClassify.py
@@ -67,6 +67,7 @@ class ClassifyManager:
 
         return self.parseXMLResponse()
 
+    # TODO: can we query by everything?
     def generate_identifier_query(self):
         identifier_map = { 
             "isbn": "bn",
--- a/mappings/oclc_bib.py
+++ b/mappings/oclc_bib.py
@@ -24,7 +24,7 @@ def map_oclc_bib_to_record(oclc_bib, related_oclc_numbers=[]) -> Record:
         # TODO: should we include merged OCLC numbers? 
         identifiers=[f"{oclc_bib['oclcNumber']}|oclc"] + \
             map(lambda issn: f"{issn}|issn", oclc_bib['issns']) if oclc_bib['issns'] else [] + \
-            map(lambda isbn: f"{isbn}|issn", oclc_bib['isbns']) if oclc_bib['isbns'] else [] + \
+            map(lambda isbn: f"{isbn}|isbn", oclc_bib['isbns']) if oclc_bib['isbns'] else [] + \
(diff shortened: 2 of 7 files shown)


managers/oclcClassify.py
```diff
@@ -67,10 +67,14 @@ def getClassifyResponse(self):
 
         return self.parseXMLResponse()
 
+    # TODO: can we query by everything?
```
mitri-slory, line 70: I'm not sure. We could try it, but it's good to have these query methods separated so it would make it easier to error handle.

mappings/oclc_bib.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def map_oclc_bib_to_record(oclc_bib, related_oclc_numbers=[]) -> Record:
+    return Record(
+        uuid=uuid4(),
+        frbr_status='complete',
+        cluster_status=False,
+        source='oclcClassify',
+        # TODO: should this be an owi number? 
+        source_id=f"{oclc_bib['oclcNumber']}|oclc",
```
mitri-slory, line 14: I don't believe so.

mappings/oclc_bib.py
```diff
@@ -0,0 +1,32 @@
...
+        # TODO: split contributors from creator field
+        contributors=[],
+        languages=[oclc_bib['language']],
+        # TODO: previously we weren't pulling in dates but we could here
+        dates=[],
+        # TODO: should we include merged OCLC numbers? 
+        identifiers=[f"{oclc_bib['oclcNumber']}|oclc"] + \
+            map(lambda issn: f"{issn}|issn", oclc_bib['issns']) if oclc_bib['issns'] else [] + \
+            map(lambda isbn: f"{isbn}|isbn", oclc_bib['isbns']) if oclc_bib['isbns'] else [] + \
+            map(lambda oclc_number: f"{oclc_number}|oclc", related_oclc_numbers),
```
mitri-slory, line 28: What do you mean by 'merged' OCLC numbers exactly?

processes/oclcClassify.py
```diff
@@ -106,19 +107,38 @@ def frbrizeRecord(self, record):
...
+
+        # TODO: move generate_search_query to utils file
+        search_query = classify_manager.generate_search_query()
+
+        # TODO: return brief records directly 
+        # TODO: check to see we don't return the record's bib
+        related_oclc_bibs = catalog_manager.query_brief_bibs(search_query)
+
+        for related_oclc_bib in related_oclc_bibs.get('briefRecords', []):
+            # TODO: returns only 10 records, should we get all of the records?
```
mitri-slory, line 133: We could try doing that, but not sure how that will affect performance and optimization which are what I'm worried about.

processes/oclcClassify.py
```diff
@@ -106,19 +107,38 @@ def frbrizeRecord(self, record):
...
+            # TODO: not sure if actually have to send this to the oclc catalog process
+            related_oclc_numbers = catalog_manager.get_related_oclc_numbers(related_oclc_bib['oclcNumber'])
```
mitri-slory, line 135: I feel like we should, but we can test to see what happens if we don't vs. if we do send it to the oclc catalog process.

mitri-slory (commented):
This PR looks good and I commented on some of the TODO questions, so once I receive your responses to them, I will approve this request.

mappings/oclc_bib.py
kylevillegas93: Haha good question - theres a list of merged OCLC numbers that come back as part of the brief bib. 

Here's more info: https://help.oclc.org/Metadata_Services/WorldShare_Collection_Manager/Understand_record_processing/Criteria_for_new_update_delete_and_merged_records

processes/oclcClassify.py
kylevillegas93: Same - do we know how much this impacts frberization?

processes/oclcClassify.py
kylevillegas93: Agreed - I think we should send them over and then make a change later on if we decide we can simplify.

kylevillegas93:
Alright @mitri-slory should be ready for review! Created a couple more tickets to test and finalize the mapping/calls to the editions endpoints.

mappings/oclc_bib.py
mitri-slory: Oh ok I see now. I would say for right now we don't need to include the merged OCLC numbers since they are essentially two records put together and I'm not sure what the criteria for what makes two records merged according to OCLC.

processes/oclcClassify.py
mitri-slory: I'm not sure but the classify process usually didn't take as long as the Catalog or Cluster Process.

mitri-slory (approved):
These changes look great and I only have one more comment I added.

PR #354, opened 2024-08-29
NO-REF: Moving up db migration in dev setup

kylevillegas93:
## Description
This change moves up the db migration process in dev setup so that it runs before trying to ingest and cluster any records.


Apophenia (approved):
Hopefully this will help resolve some of the issues!

PR #355, opened 2024-08-29
SFR-2139: FulfillProcessCheckRights

mitri-slory:
This PR focuses on updating both the fulfill process and script to check the rights status of the records associated with the manifests before they are updated. Only works that are in copyright/limited access should have their manifests updated.


processes/fulfillURLManifest.py
```diff
@@ -62,31 +62,37 @@ def update_manifest(self, metadataObject, bucketName, currKey):
```
kylevillegas93, line 62: can we update this file to be snake case?

processes/fulfillURLManifest.py
```diff
@@ -148,6 +154,15 @@ def fulfillFlagUpdate(self, metadata):
...
+         for link in metadataJSON['links']:
+            if metadataJSON['type'] == 'application/webpub+json':
+                for link in self.session.query(Link) \
+                    .filter(Link.url == metadataJSON['href'].replace('https://', '')):   
+                        if 'fulfill_limited_access' not in link.flags.keys():
+                            checkRightsStatus = 'public domain'
+                        return checkRightsStatus
 
 class FulfillError(Exception):
     pass
```
kylevillegas93, line 168: let's add a newline

scripts/addFulfillManifest.py
```diff
@@ -43,7 +43,12 @@ def update_batch(metadataObject, bucketName, currKey):
     metadataJSON = json.loads(metadataObject['Body'].read().decode("utf-8"))
     metadataJSONCopy = copy.deepcopy(metadataJSON)
 
-    counter = 0
+    #Checking if record asscociated with manifest is public domain or in copyright
```
kylevillegas93, line 46: what's the difference between the script and the process? is the script just supposed to be run locally? 

If there is duplicate logic, then I think we should encapsulate it in one place.

processes/fulfillURLManifest.py
```diff
@@ -62,31 +62,37 @@ def update_manifest(self, metadataObject, bucketName, currKey):
...
+            counter = 0
+        
+            metadataJSON, counter = self.linkFulfill(metadataJSON, counter)
+            metadataJSON, counter = self.readingOrderFulfill(metadataJSON, counter)
+            metadataJSON, counter = self.resourceFulfill(metadataJSON, counter)
+            metadataJSON, counter = self.tocFulfill(metadataJSON, counter)
+
+            if counter >= 4: 
+                for link in metadataJSON['links']:
+                    self.fulfillFlagUpdate(link)
```
kylevillegas93, line 81: The counter logic is unclear. What are your thoughts on refactoring this and the methods are also unclear?

What does it mean to fulfill the link, reading order, resource, toc, etc? What does the counter mean? Why is it if the counter >=4 that we update the fulfill flag?

processes/fulfillURLManifest.py
```diff
@@ -62,31 +62,37 @@ def update_manifest(self, metadataObject, bucketName, currKey):
 
         metadataJSON = json.loads(metadataObject['Body'].read().decode("utf-8"))
         metadataJSONCopy = copy.deepcopy(metadataJSON)
+
+        #Checking if record asscociated with manifest is public domain or in copyright
+        checkRightsStatus = 'in copyright'
+        checkRightsStatus = self.checkMetaData(metadataJSON, checkRightsStatus)
+
+        if checkRightsStatus == 'in copyright':
```
kylevillegas93, line 70: instead of indenting all of this code under this if statement, we could do

```
if rights_status == 'in copyright':
  return

File: processes/fulfillURLManifest.py [hunk 6]
```diff
@@ -148,6 +154,15 @@ def fulfillFlagUpdate(self, metadata):
                                 newLinkFlag['fulfill_limited_access'] = True
                                 link.flags = newLinkFlag
                                 self.commitChanges()
+                        
+    def checkMetaData(self, checkRightsStatus, metadataJSON):
```
kylevillegas93, line 158: checkMetadata is not a clear function name. What are we checking? 

we are trying to see if the fulfill_limited_acess flag is in the link right? Can we simplify this logic?

processes/fulfillURLManifest.py
```diff
@@ -62,31 +62,37 @@ def update_manifest(self, metadataObject, bucketName, currKey):
 
         metadataJSON = json.loads(metadataObject['Body'].read().decode("utf-8"))
         metadataJSONCopy = copy.deepcopy(metadataJSON)
+
+        #Checking if record asscociated with manifest is public domain or in copyright
+        checkRightsStatus = 'in copyright'
```
kylevillegas93, line 67: the variable name is also not clear

processes/fulfillURLManifest.py
```diff
@@ -62,31 +62,37 @@ def update_manifest(self, metadataObject, bucketName, currKey):
...
+                try:
+                    fulfillManifest = json.dumps(metadataJSON, ensure_ascii = False)
+                    return self.s3Client.put_object(
+                        Bucket=bucketName, 
+                        Key=currKey, 
+                        Body=fulfillManifest, ACL= 'public-read', 
+                        ContentType = 'application/json'
+                    )
+                except ClientError as e:
+                    logging.error(e)
```
kylevillegas93, line 95: I think we should break this out into a separate function for readability.

processes/fulfillURLManifest.py
```diff
@@ -62,31 +62,37 @@ def update_manifest(self, metadataObject, bucketName, currKey):
 
         metadataJSON = json.loads(metadataObject['Body'].read().decode("utf-8"))
         metadataJSONCopy = copy.deepcopy(metadataJSON)
+
+        #Checking if record asscociated with manifest is public domain or in copyright
```
kylevillegas93, line 66: I think if we make this method clearer we can remove this comment. Your comment is explaining what the code is doing but your code is not explaining what it itself is doing. Lean into the latter.

processes/fulfillURLManifest.py
mitri-slory: Sure

processes/fulfillURLManifest.py
mitri-slory: Sure

scripts/addFulfillManifest.py
mitri-slory: The script was initially made as a manual process to add the fulfill manifests which was later used to create the fulfill process which is an automatic task that can be run manually. We don't need the script anymore since the process will serve both purposes.

processes/fulfillURLManifest.py
mitri-slory: Sure

processes/fulfillURLManifest.py
mitri-slory: I changed the name to check_rights_status which is what we are checking. The logic is as simplified as it can be given my understanding.

processes/fulfillURLManifest.py
mitri-slory: I changed the variable to rights_status to make it more clear.

processes/fulfillURLManifest.py
mitri-slory: I mentioned changing the name of the method to be more clear in a earlier comment so I'll delete this comment.

mitri-slory:
I'll make another commit to add comments to some of the methods and I'll respond to the rest of your comments @kylevillegas93 tomorrow.

scripts/addFulfillManifest.py
kylevillegas93: should we delete the script?

mitri-slory: Yes I believe we should and in my newest commit I did delete the script.

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+
+        self.fullImport = self.process == 'complete' 
+        self.startTimestamp = None
+
+        # Connect to database
+        self.generateEngine()
+        self.createSession()
+
+        # S3 Configuration
+        self.s3Bucket = os.environ['FILE_BUCKET']
```
kylevillegas93, line 27: s3_bucket

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+            startTimeStamp = datetime.now(timezone.utc) - timedelta(days=1)
+            self.get_manifests(startTimeStamp)
+        elif self.process == 'complete':
+            self.get_manifests()
+        elif self.process == 'custom':
+            time_stamp = self.ingestPeriod
+            startTimeStamp = datetime.strptime(time_stamp, '%Y-%m-%dT%H:%M:%S')
+            self.get_manifests(startTimeStamp)
+
+    def get_manifests(self, startTimeStamp=None):
```
kylevillegas93, line 43: startTimeStamp -> start_time_stamp

kylevillegas93: the comment below is explaining further what this function does. Can we rename this function in order to remove the comment?

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+from botocore.exceptions import ClientError
+
+from .core import CoreProcess
+from datetime import datetime, timedelta, timezone
+from model import Link
+from logger import createLog
+
+logger = createLog(__name__)
+
+class Fulfill_Process(CoreProcess):
```
kylevillegas93, line 14: class name can be FulfillProcess

kylevillegas93: FulfillProcess is also vague. Is there a better name that we can use to be explicit?

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __init__(self, *args):
+        super(Fulfill_Process, self).__init__(*args[:4])
+
+        self.fullImport = self.process == 'complete' 
+        self.startTimestamp = None
+
+        # Connect to database
```
kylevillegas93, line 22: let's remove this comment - I think generateEngine and createSession are clear enough that we are connecting to the database.

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+        self.generateEngine()
+        self.createSession()
+
+        # S3 Configuration
```
kylevillegas93, line 26: same here - I think this makes sense.

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+        self.s3Bucket = os.environ['FILE_BUCKET']
+        self.host = os.environ['DRB_API_HOST']
+        self.prefix = 'manifests/UofM/'
+        self.createS3Client()
+
+    def runProcess(self):
+        if self.process == 'daily':
+            startTimeStamp = datetime.now(timezone.utc) - timedelta(days=1)
```
kylevillegas93, line 34: start_timestamp

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+            self.get_manifests(startTimeStamp)
+        elif self.process == 'complete':
+            self.get_manifests()
+        elif self.process == 'custom':
+            time_stamp = self.ingestPeriod
+            startTimeStamp = datetime.strptime(time_stamp, '%Y-%m-%dT%H:%M:%S')
```
kylevillegas93, line 40: start_timestamp

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+        # S3 Configuration
+        self.s3Bucket = os.environ['FILE_BUCKET']
+        self.host = os.environ['DRB_API_HOST']
+        self.prefix = 'manifests/UofM/'
+        self.createS3Client()
+
+    def runProcess(self):
+        if self.process == 'daily':
+            startTimeStamp = datetime.now(timezone.utc) - timedelta(days=1)
+            self.get_manifests(startTimeStamp)
```
kylevillegas93, line 35: From reading this function from here, it just looks as though we are getting manifests.  But this function has two responsibilities: 

1. Get manifests
2. Update manifests

I like following the single responsibility principle per function.

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+                .filter(Link.url == metadata['href'].replace('https://', '')):   
+                        if 'fulfill_limited_access' in link.flags.keys():
+                            if link.flags['fulfill_limited_access'] == False:
+                                newLinkFlag = dict(link.flags)
+                                newLinkFlag['fulfill_limited_access'] = True
+                                link.flags = newLinkFlag
+                                self.commitChanges()
+                        
+    def check_rights_status(self, metadata_json, rights_status):
+        for metadata in metadata_json['links']:
```
kylevillegas93, line 157: are we iterating over the metadata or or the links

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+            for link in self.session.query(Link) \
+                .filter(Link.url == metadata['href'].replace('https://', '')):   
+                        if 'fulfill_limited_access' in link.flags.keys():
+                            if link.flags['fulfill_limited_access'] == False:
+                                newLinkFlag = dict(link.flags)
+                                newLinkFlag['fulfill_limited_access'] = True
+                                link.flags = newLinkFlag
+                                self.commitChanges()
+                        
+    def check_rights_status(self, metadata_json, rights_status):
```
kylevillegas93, line 156: it seems what we are doing is actually checking to see if the rights status is in copyright? should this return a boolean?

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+
+        self.closeConnection()
+
+        if metadata_json != metadata_json_copy:
+            try:
+                fulfill_manifest = json.dumps(metadata_json, ensure_ascii = False)
+                return self.s3Client.put_object(
+                    Bucket=bucket_name, 
+                    Key=curr_key, 
+                    Body=fulfill_manifest, ACL= 'public-read',
```
kylevillegas93, line 91: let's put ACL on a new line

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+            try:
+                fulfill_manifest = json.dumps(metadata_json, ensure_ascii = False)
+                return self.s3Client.put_object(
+                    Bucket=bucket_name, 
+                    Key=curr_key, 
+                    Body=fulfill_manifest, ACL= 'public-read', 
+                    ContentType = 'application/json'
+                )
+            except ClientError as e:
+                logging.error(e)
```
kylevillegas93, line 95: should this be logger?

processes/fulfillURLManifest.py
mitri-slory: The counter is there to keep track and make sure all the necessary pdf/epub links in the manifest JSON document are replaced with fulfill endpoints. The structure of our manifest documents always have a links, readingOrder, resource, and toc dictionary that include pdf/epub links. Therefore to ensure that the fulfill_limited_access flag doesn't become true while one of the links is not replaced with a fulfill endpoint for some reason or another the counter must be at least 4 to ensure all four dictionaries have been updated properly.

I can refactor the counter to instead be four separate flags that become true for each fulfill method to make this more clear. What do you think?

processes/fulfill_url_manifest.py
mitri-slory: Sure. I can rename it to get_manifest_batches

processes/fulfill_url_manifest.py
mitri-slory: I can rename it FulfillURLManifestProcess. What are your thoughts?

processes/fulfill_url_manifest.py
mitri-slory: I agree

processes/fulfill_url_manifest.py
mitri-slory: I agree

processes/fulfill_url_manifest.py
mitri-slory: Ok I can refactor the code to make `update_manifest `not be called in the `get_manifests` method but instead the runProcess method.

processes/fulfill_url_manifest.py
mitri-slory: We're iterating over the links. I just changed it because I had two 'link' variables in this method with one representing the manifest link and another for a database link object. I'll change this `metadata` variable to `link` and the database link object to `psql_link`

processes/fulfill_url_manifest.py
mitri-slory: Yeah we are checking if it's in copyright. I can change the method name to `check_copyright_status` and have a boolean value be returned as well.

processes/fulfill_url_manifest.py
mitri-slory: Yeah it should be. I'll change it

processes/fulfill_url_manifest.py
mitri-slory: From my perspective the method is mainly getting the manifests to provide for the  calls the update_manifests method to update the manifests. I think it's better to keep it like this for now and I'll keep what you said in mind

processes/fulfill_url_manifest.py
kylevillegas93: That works!

processes/fulfill_url_manifest.py
kylevillegas93: We are updating the manifests in this function though. I recommend either renaming it so it is clear that the function is updating or you split the function up - 1 get function that returns the batches and another that updates the batches.

processes/fulfill_url_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+        counter = 0
+    
+        metadata_json, counter = self.link_fulfill(metadata_json, counter)
+        metadata_json, counter = self.reading_order_fulfill(metadata_json, counter)
+        metadata_json, counter = self.resource_fulfill(metadata_json, counter)
+        metadata_json, counter = self.toc_fulfill(metadata_json, counter)
+
+        if counter >= 4: 
+            for link in metadata_json['links']:
+                self.fulfill_flag_update(link)
```
kylevillegas93, line 81: Let's refactor this part of the code. Can we do it without a counter?

processes/fulfillURLManifest.py
kylevillegas93: I think what you can do is try to use try/except and error handling for this. There are interesting edge cases here like what happens if there are 4 links in linkFulfill but no reading order, resource, nor toc links? 

I think we should try to do it without a flag and without a counter. One way to think about it is to simplify the functions that you are calling in this block. What if we get all the links ids for the links, reading order links, resource links, and toc links? 

We can use error handling to answer questions like what happens when there are no links found.

mitri-slory: For now I'll implement a try/except block to error handle those edge cases but I'll keep the counter since I have other tasks to focus on at the moment and simplifying the functions to one method isn't necessary at this time since the process is working at this moment.

processes/__init__.py
```diff
@@ -18,4 +18,4 @@
 from .UofSC import UofSCProcess
 from .loc import LOCProcess
 from .UofM import UofMProcess
-from .fulfillURLManifest import FulfillProcess
```
kylevillegas93, line 21: nit - new line!

scripts/__init__.py
```diff
@@ -13,5 +13,4 @@
 from .updatePubLocationAndLinks import main as updateLocationAndLinks
 from .countCABooks import main as countCA
 from .nyplLoginFlags import main as nyplFlags
-from .deleteUMPManifestLinks import main as deleteUMPManifests
-from .addFulfillManifest import main as fulfillManifest
\ No newline at end of file
+from .deleteUMPManifestLinks import main as deleteUMPManifests
```
kylevillegas93, line 16: nit: new line!

tests/unit/test_fulfillManifest_process.py
```diff
@@ -32,27 +32,26 @@ def __init__(self):
...
         )
 
         mockPrefix = mocker.MagicMock(prefix='testPrefix')
         mockTimeStamp = mocker.MagicMock(timeStamp='testTimeStamp')
         
-        testProcess.getManifests(mockTimeStamp)
-
+        testProcess.fetch_and_update_manifests(mockTimeStamp)
 
         processMocks['load_batches'].assert_called_once_with('testPrefix','test_aws_bucket')
```
kylevillegas93, line 57: nit: new line!

kylevillegas93 (approved):
Recommend just making sure we have added new lines to all the new files! Otherwise, looks good.

tests/unit/test_fulfillManifest_process.py
mitri-slory: There is a new line after the `processMocks` method call.

PR #356, opened 2024-09-05
SFR-2141: Delete Duplicate Work Identifiers

mitri-slory:
This PR focuses on finding a way to prevent duplicate work_identifier rows from being created which caused issues when clustering some works. The idea here is to first delete the current work_identifier duplicates from the QA and prod environment. Then, add a unique constraint to both the work_id and identifier_id columns through a migration.


processes/sfrCluster.py
```diff
@@ -101,7 +101,7 @@ def clusterRecord(self, rec):
 
         try:
             self.session.flush()
-        except (DataError, StaleDataError) as e:
+        except (IntegrityError, DataError, StaleDataError) as e:
```
kylevillegas93, line 104: Hmm - what would happen if one of these errors are not thrown? Should we generally handle all exceptions here?

managers/sfrRecord.py
```diff
@@ -39,7 +39,8 @@ def mergeRecords(self):
 
         matchedWorks.sort(key=lambda x: x[1])
 
-        allIdentifiers = self.work.identifiers
+        allIdentifiers = self.work.identifiers.copy()
```
kylevillegas93, line 42: Is my understanding correct here that by creating a shallow copy and deduping and extending allIdenfitiers, we also update self.work.idenfitiers as well?

kylevillegas93: Is there a case where self.work.identifiers is null?

mitri-slory: With a shallow copy the original list is only updated if the nested list elements are modified so adding a new element wouldn't change it. I will change this line to a deepcopy instead though so there's no chance of `self.work.identifiers` from being updated alongside allIdentifiers.

processes/sfrCluster.py
kylevillegas93: Awesome, that sounds good!

managers/sfrRecord.py
kylevillegas93: Cool - I think that sounds good!

kylevillegas93 (approved):

PR #358, opened 2024-09-09
NO-REF: Downloading and aggregating S3 access logs

kylevillegas93:
# Description
- POC of the short-term work to scale the report

# Testing
- Depending on the number of files, it's about a minute to run per day so for a quarter it would take 90 minutes to run the script.


fatimarahman:
I think overall this looks good! I'm curious, have you tested this out, i.e. integrated this into the existing code? I'm curious to see if after aggregating, the analysis runs any faster. If it does, maybe we can integrate this into the Counter5Controller code and remove the S3 calls from the rest of the code.

Also! I think if we end up using this code, we could move the log aggregator into a `helpers` folder. Lmk your thoughts!

fatimarahman:
I can approve after we test out integration! I can also do that if you don't have the bandwidth

fatimarahman:
Did some testing on my end, here are the results!:

- Original code took 15 min to generate report for 2 week period, new code took 3 min
- Original code took 25 min to generate report for month of August, new code took 7 min

fatimarahman (approved):

PR #360, opened 2024-09-10
NO-REF: Adding OCLC query bibs call

kylevillegas93:
## Description 
- Implements a call to query bibs from OCLC
- On of our hypotheses is that brief bibs does not provide enough information to cluster OCLC records

## Testing
`make integration`


Apophenia (approved):
Thanks for adding this! Looks good!

If the lambdas concerning the contributor role edge cases concern logic that is in the brief-bibs endpoint, we might want to backport that, but maybe we won't end up using brief-bibs at all for now.

PR #361, opened 2024-09-11
SFR-2090: Finalize OCLC implementation

kylevillegas93:
## Description
- Finalizes the OCLC implementation 
- Uses bib records to get subjects and full set of contributors 
- Uses work id as owi and sets that as the main source_id of oclc classify records
- Filters out item types we are not interested in 

## Testing
- Dev setup process clustered 100% of records from HathiTrust, 7 / 11 records from OCLC classify, and all OCLC catalog records.
- Deleted dev setup process unit tests - they are blocking dev more than helping


processes/oclcClassify.py
```diff
@@ -122,23 +122,24 @@ def frbrizeRecord(self, record):
     def classify_record_by_metadata_v2(self, identifier, identifier_type, author, title):
         search_query = self.oclc_catalog_manager.generate_search_query(identifier, identifier_type, title, author)
 
-        # TODO: SFR-2090: Finalize query to brief bibs
-        related_oclc_bibs = self.oclc_catalog_manager.query_brief_bibs(search_query)
+        related_oclc_bibs = self.oclc_catalog_manager.query_bibs(search_query)
```
Apophenia, line 125: query_bibs would recover gracefully from a 400, but I think we could just return the v2 classify function or otherwise short circuit the call to OCLC with a reasonable error if there's no query (as long as we trust the original logic, that it's not worth trying to query on title alone, e.g.).

managers/oclc_catalog.py
```diff
@@ -13,6 +13,10 @@
 class OCLCCatalogManager:
     CATALOG_URL = 'http://www.worldcat.org/webservices/catalog/content/{}?wskey={}'
     OCLC_SEARCH_URL = 'https://americas.discovery.api.oclc.org/worldcat/search/v2/'
+    ITEM_TYPES = ['archv', 'audiobook', 'book', 'encyc', 'jrnl']
+    LIMIT = 50
+    MAX_NUMBER_OF_RECORDS = 100
```
Apophenia, line 18: Is this an arbitrary/placeholder number or does this seem like it achieves parity with the old process? Either way, good to have these as constants

Apophenia (approved):
This looks good to me!

managers/oclc_catalog.py
kylevillegas93: It definitely will retrieve less than the current implementation. This just speeds its up for testing purposes. I think we can tweak this number accordingly!

PR #363, opened 2024-09-12
NO-REF: Switching to v2 classify record

kylevillegas93:
## Description
- Switching over to the v2 classify record implementation 

## Testing 
`make unit`


Apophenia (approved):
Wow, that really is a minimal change, but it makes sense since the functions more or less take the same shape. Thanks!

PR #365, opened 2024-09-13
NOREF-Better error handling & logging

Apophenia:
Todo:
- Check exception in query building function is handled correctly if query can't be formed
- Ensure graceful failure if OCLC json does not contain error type

### DIFF
--- a/managers/oclc_catalog.py
+++ b/managers/oclc_catalog.py
@@ -96,17 +96,20 @@ class OCLCCatalogManager:
             )
 
             if other_editions_response.status_code != 200:
-                logger.warning(f'OCLC other editions request failed with status {other_editions_response.status_code}')
+                logger.warning(
+                            f"Other editions query for no {oclc_number} failed "
+                            f"With status {other_editions_response.status_code} "
+                            f"and reason {other_editions_response.json()['type']}")
                 return None
-            
+
             return other_editions_response.json()
-        except Exception as e: 
-            logger.error(f'Failed to query other editions endpoint {other_editions_url}', e)
+        except Exception as e:
+            logger.error(f'Failed to query other editions endpoint {other_editions_url}: Exception: {e}')
             return None
-        
+
     def _get_oclc_number_from_bibs(self, oclc_number: int, oclc_bibs) -> int:
         return [int(edition['oclcNumber']) for edition in oclc_bibs if int(edition['oclcNumber']) != oclc_number]
-  
+
     def query_bibs(self, query: str):
         bibs = []
 
@@ -133,7 +136,7 @@ class OCLCCatalogManager:
 
             return bibs
         except Exception as e:
-            logger.error(f'Failed to query search bibs with query {query}', e)
+            logger.error(f'Failed to query search bibs with query {query}. Exception: {e}')
             return bibs
     
     def _search_bibs(self, query: str, offset: int=0):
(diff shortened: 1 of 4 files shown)


managers/oclc_catalog.py
```diff
@@ -96,17 +96,20 @@ def _get_other_editions(self, oclc_number: int, offset: int=0):
             )
 
             if other_editions_response.status_code != 200:
-                logger.warning(f'OCLC other editions request failed with status {other_editions_response.status_code}')
+                logger.warning(
+                            f"Other editions query for no {oclc_number} failed "
```
kylevillegas93, line 100: nit: let's use single quotation marks

managers/oclc_catalog.py
```diff
@@ -96,17 +96,20 @@ def _get_other_editions(self, oclc_number: int, offset: int=0):
...
+                            f"With status {other_editions_response.status_code} "
+                            f"and reason {other_editions_response.json()['type']}")
```
kylevillegas93, line 102: I think we should be careful with getting the json body. If it's a 500 error, there may not be a json body and we may error out here.

managers/oclc_catalog.py
```diff
@@ -96,17 +96,20 @@ def _get_other_editions(self, oclc_number: int, offset: int=0):
...
                 return None
-            
+
             return other_editions_response.json()
-        except Exception as e: 
-            logger.error(f'Failed to query other editions endpoint {other_editions_url}', e)
+        except Exception as e:
+            logger.error(f'Failed to query other editions endpoint {other_editions_url}: Exception: {e}')
```
kylevillegas93, line 107: We may not get the entire stack trace here, which may be helpful to debug.

managers/oclc_catalog.py
```diff
@@ -133,7 +136,7 @@ def query_bibs(self, query: str):
 
             return bibs
         except Exception as e:
-            logger.error(f'Failed to query search bibs with query {query}', e)
+            logger.error(f'Failed to query search bibs with query {query}. Exception: {e}')
```
kylevillegas93, line 139: Same as above

processes/oclcClassify.py
```diff
@@ -119,7 +119,11 @@ def frbrizeRecord(self, record):
                 logger.debug(err.message)
 
     def classify_record_by_metadata_v2(self, identifier, identifier_type, author, title):
-        search_query = self.oclc_catalog_manager.generate_search_query(identifier, identifier_type, title, author)
+
+        try:
+            search_query = self.oclc_catalog_manager.generate_search_query(identifier, identifier_type, title, author)
```
kylevillegas93, line 124: hmm are we erring out when we generate the search query?

processes/core.py
```diff
@@ -19,17 +24,17 @@ def addDCDWToUpdateList(self, rec):
         existing = self.session.query(Record)\
             .filter(Record.source_id == rec.record.source_id).first()
         if existing:
-            print('EXISTING', existing)
+            logger.debug('Existing record: ' + str(existing))
```
kylevillegas93, line 27: nice! this is much needed haha

processes/oclcClassify.py
Apophenia: In main, if "generate search query" returns None because a search query can't be built, the code still sends the (empty) search query to OCLC and then fails with a warning level log and continues. I'm open to how exceptions should be handled but I do think we should avoid sending empty queries to OCLC - very flexible about how or where we do that!

Apophenia: I'm going to try a different rework

managers/oclc_catalog.py
```diff
@@ -96,17 +96,24 @@ def _get_other_editions(self, oclc_number: int, offset: int=0):
             )
 
             if other_editions_response.status_code != 200:
-                logger.warning(f'OCLC other editions request failed with status {other_editions_response.status_code}')
+                logger.warning(
+                            f'OCLC search bibs request for OCLC no {oclc_number} failed '
+                            f'with status {other_editions_response.status_code}')
+                try:
+                            oclc_error_type = other_editions_response.json()["type"]
```
kylevillegas93, line 103: I think we can do get('type') here instead to just guard.

Apophenia: I think json() could still throw a JSONDecodeError before the get() is run on type, but it could be split out such that the exception is narrowed to not include both the JSONDecodeError and the KeyError here - do you think that would be better?

Apophenia: (The possibilities including: the response not having valid JSON, and the response having valid JSON but it might not contain a type)

kylevillegas93: But we aren't handling the TypeError right? I think using get would prevent the KeyError.

kylevillegas93 (approved):
Approving for now - let's definitely cleanup the error handling and make sure we've documented/covered all of the edge cases.

PR #367, opened 2024-09-17
SFR-2171: Fixing OCLC bib author mapping

kylevillegas93:
## Description
- FIxes OCLC bib author mapping
- Tries to get text first then fallbacks to romanized text
- Handles first name or second name not being present in the contributor

## Testing 
`make unit`


tests/unit/test_oclc_bib_mapping.py
```diff
@@ -0,0 +1,84 @@
...
+            'secondName': {
+                'romanizedText': 'Simpson'
+            },
+            'isPrimary': True
+        }]
+    }
+
+    oclc_bib_mapping = OCLCBibMapping(base_oclc_bib)
+
+    assert ['Simpson, Homer|||true'] == oclc_bib_mapping.record.authors
```
Apophenia, line 84: Thanks for adding these!

Apophenia (approved):
I think this takes care of the known edge cases. If we come across more, we have a framework for adding a test & fix. Thanks for putting this together!

PR #369, opened 2024-09-18
SFR-2173: Improving error handling and logging in OCLC classify process

kylevillegas93:
## Description
- Improves error handling and logging in the OCLC classify process
- Switches to using detail from OCLC error response, e.g:
```
{
  "type": "INVALID_QUERY_PARAMETER_VALUE",
  "title": "Validation Failure",
  "detail": "oclcNumber must be a positive whole number"
}
```

## Testing 
`make test`

### DIFF
--- a/managers/oclc_catalog.py
+++ b/managers/oclc_catalog.py
@@ -139,7 +139,7 @@ class OCLCCatalogManager:
 
             return bibs
         except Exception as e:
-            logger.error(f'Failed to query search bibs with query {query} due to {e.message}')
+            logger.error(f'Failed to query search bibs with query {query} due to {e}')
             return bibs
 
     def _search_bibs(self, query: str, offset: int=0):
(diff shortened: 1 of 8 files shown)


managers/oclc_catalog.py
```diff
@@ -140,7 +139,7 @@ def query_bibs(self, query: str):
 
             return bibs
         except Exception as e:
-            logger.error(f'Failed to query search bibs with query {query}', e)
```
Apophenia, line 143: Did you revert this because of an unexpected behavior with the exception printing? I ran into something strange but I wasn't able to isolate it, but I'm curious because I think this exception logging pattern exists elsewhere in the codebase.

Apophenia (approved):
Thanks, looks good!

managers/oclc_catalog.py
kylevillegas93: hmm lemme see

kylevillegas93: changed it back!

PR #371, opened 2024-09-19
NO-REF: Fixing OCLC catalog query attempts bug

kylevillegas93:
## Description
- We discovered that the ratio of OCLC catalog records in QA vs local did not make sense
- There was bug in the OCLC catalog manager where the attempts kept increasing since it was a class variable and we we would not actually call OCLC. 

## Testing
Ran ETL pipeline against subset of HathiTrust record to see the increase in OCLC catalog records:

Screenshot 2024-09-19 at 4 56 41 PM

### DIFF
--- a/mappings/oclcCatalog.py
+++ b/mappings/oclcCatalog.py
@@ -160,17 +160,10 @@ class CatalogMapping(XMLMapping):
         self.record.source = 'oclcCatalog'
         self.record.source_id = self.record.identifiers[0]
         self.record.frbr_status = 'complete'
-
-        logger.info(f'Formatting OCLC Catalog record with id: {self.record.source_id}')
-
-        # Parse language field
-        logger.debug('Parsing ISO lang code from 008 fixed field')
+        
         _, _, lang_3, *_ = tuple(self.record.languages[0].split('|'))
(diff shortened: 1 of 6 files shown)


mappings/oclcCatalog.py
```diff
@@ -160,17 +160,10 @@ def applyFormatting(self):
         self.record.source = 'oclcCatalog'
         self.record.source_id = self.record.identifiers[0]
         self.record.frbr_status = 'complete'
-
-        logger.info(f'Formatting OCLC Catalog record with id: {self.record.source_id}')
```
Apophenia, line 164: I definitely agree conceptually with removing most of the debug logging in this file, but many of the errors currently being thrown by the clustering process are difficult to identify because it's harder to get a handle on the original record (and which source is throwing the error) when the error passes through the mapping. I definitely think we should focus some energy on making it easier to pinpoint these issues before we re-introduce the ingest processes.

Apophenia (approved):
Thanks for identifying this!

mappings/oclcCatalog.py
kylevillegas93: makes sense!

PR #373, opened 2024-09-23
SFR-2179/get_collection_list

ayan1229:
- This PR validates the api response for a query that returns collection list by title. 
- it also verifies that when the API is hit, it doesn't return an empty JSON body in response


tests/rest_api/test_get_collection.py
```diff
@@ -19,4 +19,4 @@ def test_get_collection():
 
     response_json = response.json()
     
-    assert response_json is not None
+    assert response_json is not None
```
kylevillegas93, line 22: nit: new line at eof

tests/rest_api/test_get_list.py
```diff
@@ -0,0 +1,22 @@
...
+        f"API call failed.\n"
+        f"Expected status code: {200}\n"
+        f"Actual status code: {response.status_code}\n"
+        f"URL: {url}\n"
+        f"Response text: {response.text[:100]}..."
+    )
+
+    response_json = response.json()
+    
+    assert response_json is not None
```
kylevillegas93, line 22: nit: new line at eof

tests/rest_api/test_get_list.py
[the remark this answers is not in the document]
ayan1229: renaming to test_get_collection_list

tests/rest_api/test_get_list.py
ayan1229: do you always want to keep an empty line at the end? or no lines

kylevillegas93: always a new line that is empty at the end of the file!

PR #375, opened 2024-09-26
SFR-2105: Deprecate OCLC Classify Manager

Apophenia:
- Removes Classify manager code
- Moves one remaining Classify manager function into Classify process (logically this might be better off in a manager, but moving it into a manager file would have presented more work, so I just made it a utility function in the process and the unit tests could pass with only small changes to the mocks)
- Removes unused functions specific to Classify XML processing and mapping from the Classify process
- Removes ClassifyRecordByMetadata "v1" code

Some functions were unused, but represent logic that was recreated in the Classify process (e.g. the "v2" metadata fetch function) and would benefit from new tests. Other portions of the removed logic seem to already be covered by new mapping tests.


tests/unit/test_oclcClassify_process.py
```diff
@@ -349,12 +271,5 @@ def test_sendCatalogLookupMessage(self, testInstance, mocker):
...
-        testXML = testXMLResponse(2, '<work owi="1"/>')
-
-        mockCheckRedis = mocker.patch.object(ClassifyProcess, 'checkSetRedis')
-        mockCheckRedis.return_value = False
-
-        assert testInstance.checkIfClassifyWorkFetched(testXML) == False
-
-        mockCheckRedis.assert_called_once_with('classifyWork', '1', 'owi')
+    def test_get_queryable_identifiers(self):
+        assert ClassifyProcess._get_queryable_identifiers(['1|isbn', '2|test']) == ['1|isbn']
```
kylevillegas93, line 275: nit: newline at the end of this test!

kylevillegas93 (approved):
Looks good! Once you're happy with the e2e processes, let's merge!

processes/oclcClassify.py
```diff
@@ -212,15 +170,14 @@ def fetchOCLCCatalogRecords(self, identifiers):
...
 
-    def checkIfClassifyWorkFetched(self, classifyXML):
-        workOWI = classifyXML.find(
-            './/work', namespaces=ClassifyManager.NAMESPACE
-        ).attrib['owi']
-
-        return self.checkSetRedis('classifyWork', workOWI, 'owi')
+    def _get_queryable_identifiers(identifiers):
+        return list(filter(
+            lambda x: re.search(r'\|(?:isbn|issn|oclc)$', x) != None,
```
kylevillegas93, line 181: nit: x is very generic, we could use identifier or id to be specific.

PR #376, opened 2024-09-26
NO-REF: Changing print statements to New Relic logging

kylevillegas93:
# Description
- Changes print statements to log statements


Apophenia (approved):
Looks good, thanks for changing these!

PR #378, opened 2024-09-27
NO-REF: Remove changelog

kylevillegas93:
# Description
- Removes the changelog - we will rely on githubs commit history to summarize the changes going out in releases


Apophenia (approved):
Goodbye changelog!

PR #379, opened 2024-09-27
NO-REF: Creating one root logger

kylevillegas93:
## Description
- Creating one root logger 
- Removes redundant code 
- Filename is included in log: 
`file.name":"/src/api/blueprints/drbCollection.py"`

## Testing
`docker compose up; python main.py -p DevelopmentSetupProcess -e local -i complete `


app_logging.py
```diff
@@ -28,3 +28,8 @@ def createLog(module):
     logger.addHandler(consoleLog)
 
     return logger
+
+
+stage = os.getenv('STAGE', 'qa')
+
+logger = create_logger(f'drb-etl-pipeline-{stage}')
```
mitri-slory, line 35: What is the purpose of this stage variable?

kylevillegas93: Just to differentiate between the prod and qa logger in New Relic. They are differentiated by the application name but good to have different logger names as well.

mitri-slory: That makes sense and a nice clarity check.

PR #384, opened 2024-10-01
SFR-2181: get edition id

ayan1229:
This PR verifies the following

1. API response for /edition/{id}
2. Verifies the returned JSON body is not null


tests/rest_api/test_get_edition_id.py
kylevillegas93, line 1: `test_get_edition.py`

kylevillegas93 (changes requested):
Looks good - just a file name change.

PR #390, opened 2024-10-04
SFR-2216: Improving DOAB logging and error handling

kylevillegas93:
## Description
- Improves logging and error handling
- DOAB is not ingesting e2e as of right now


mitri-slory (approved):
Good catch seeing the typographic errors here.

PR #392, opened 2024-10-04
SFR-2216: Fixing LOC process ingestion

kylevillegas93:
## Description
- Fixing a couple minor bugs with mapping LOC records
- Skipping ingest of records where we can't map authors 

## Testing
`make test; python main.py -p LOCProcess -e local -i complete -l 10`


mitri-slory (approved):
I appreciate the changes made to the LOC process and mapping since this code hasn't been touched for a while.

PR #393, opened 2024-10-04
SFR-2216: Fixing NYPL ingest process locally

kylevillegas93:
# Description
- SQLAlchemy is returning a tuple and not a JSON as per the original code. 
- I added a _map_bib function that will try to unpack a tuple first and then default to the bib. 
- Streamlined the query to only return fields we are interested in

# Testing
`python main.py -p NYPLProcess -e local -i complete -l 10`


mitri-slory (approved):
Really good changes here!

PR #394, opened 2024-10-07
SFR-2234 SFR-2235: Fix bugs in DOAB ingestion process

kylevillegas93:
# Description
- DOAB is returning 200 from the browser but not from Python code
- Adding a user agent header to mimic the header that the browser sends 
- Fixes bug where book-access element does not exist
- Fixes bug where finding final URI had too many redirects and hit the recursion call stack limit of 1000

# Testing 
`python main.py -p DOABProcess -e local -i complete -l 10`

### DIFF
--- a/managers/doabParser.py
+++ b/managers/doabParser.py
@@ -72,7 +72,12 @@ class DOABLinkManager:
         self.record.has_part = parsedLinks
     
     @staticmethod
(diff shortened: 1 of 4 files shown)


Apophenia (approved):
Looks great!

PR #396, opened 2024-10-10
SFR-2245: Refactor dev setup process and fix infinite cluster loop

kylevillegas93:
# Description
- Refactors dev setup process
- Removes the dev setup process docker container as it doesn't need to run every time you run docker compose up
- Fixes infinite cluster loop bug where the cluster_status would always remain false if we failed to cluster

# Testing 
`python main.py -p DevelopmentSetupProcess -e local`

### DIFF
--- a/processes/developmentSetup.py
+++ b/processes/developmentSetup.py
@@ -11,136 +11,169 @@ from time import sleep
 
 from managers.db import DBManager
 from .core import CoreProcess
+from logger import createLog
 from mappings.hathitrust import HathiMapping
 from .oclcClassify import ClassifyProcess
 from .oclcCatalog import CatalogProcess
 from .sfrCluster import ClusterProcess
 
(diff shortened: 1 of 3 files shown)


processes/developmentSetup.py
```diff
@@ -11,136 +11,169 @@
...
+                sa.text(
+                    f"GRANT ALL PRIVILEGES ON DATABASE {os.environ['POSTGRES_NAME']} "
+                    f"TO {os.environ['POSTGRES_USER']}",
+                ),
+            )
+        except ProgrammingError:
+            pass
+        except Exception as e:
+            logger.exception('Failed to create database user')
+            raise e
```
mitri-slory, line 110: Good idea to separate those lines of code into two separate methods for clarity.

mitri-slory (approved):
These changes looks great.

PR #397, opened 2024-10-11
SFR-2247: Fixing deep copy work identifiers bug

mitri-slory:
This hotfix PR is meant to fix a bug when making a deep copy of the workIdentifiers array since there are Instance objects and deep copy isn't naturally designed to make copies of Instance objects. In this case, it would be better to make a new array that references the Instance objects instead with a shallow copy.


Apophenia:
After running this locally with the changes in place, I no longer see the errors in the developer setup process. It makes sense to me that this might fix the error, since SQLAlchemy manages both the original custom data structure and the Instance objects in it.

PR #408, opened 2024-10-22
SFR-2277: Renaming process files and refiling ingest processes

kylevillegas93:
## Description
- Renames process files to snake case
- Moves ingest processes under an ingest folder 
- Moves processes unit tests under same folder structure 

## Testing
`make unit; python main.py -p NYPLProcess -e local -i complete -l 10`


tests/unit/processes/ingest/test_u_of_m_process.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 
 from mappings.core import MappingError
-from processes.UofM import UofMProcess
+from processes import UofMProcess
```
mitri-slory, line 4: Shouldn't this line be `from processes.ingest.u_of_m import UofMProcess.`

mitri-slory (approved):
This PR looks good and I just have one typo change that should be addressed before this gets merged in.

tests/unit/processes/ingest/test_u_of_m_process.py
kylevillegas93: ah because the __init__.py file in the processes folder imports the classes from the ingest file, this still works.

mitri-slory: Ah ok. I was just wondering since the other tests for these processes have that code line structure.

kylevillegas93: ya some of the errors classes are not imported in the processes module, so it needs to be the full path

mitri-slory: Oh appreciate the clarification

PR #410, opened 2024-10-22
Classify logging improvements

Apophenia:
- Adds a persistent count of records that have been updated during the process
- Adds a summary of this count before process exists
- Debug-level statement previously only logged per "chunk", aka when a full window of records was saved. Moved this debug statement to the function that actually persists these so that all occurrences of this are logged, since ingests of 5-10 records e.g. resulted in no debug statements


Apophenia:
Oops, need to fix that mock!

PR #411, opened 2024-10-22
SFR-2265: Refactored Chicago ISAC mapping and processes

mitri-slory:
This PR removes comments, uses composition instead of inheritance for the __init__ method, and transfers some code into a new method.

### DIFF
--- a/processes/ingest/chicago_isac.py
+++ b/processes/ingest/chicago_isac.py
@@ -13,8 +13,6 @@ class ChicagoISACProcess(CoreProcess):
     def __init__(self, *args):
         super(ChicagoISACProcess, self).__init__(*args[:4])
 
-        self.full_import = self.process == 'complete' 
-
         self.generateEngine()
         self.createSession()
 
--- a/processes/ingest/chicago_isac.py
+++ b/processes/ingest/chicago_isac.py
@@ -3,7 +3,7 @@ import os
 
 from ..core import CoreProcess
 from mappings.chicagoISAC import ChicagoISACMapping
-from managers import DBManager, ElasticsearchManager, RedisManager, S3Manager, WebpubManifest
+from managers import WebpubManifest
 from logger import createLog
 
 logger = createLog(__name__)
@@ -11,52 +11,40 @@ logger = createLog(__name__)
 class ChicagoISACProcess(CoreProcess):
 
     def __init__(self, *args):
+        super(ChicagoISACProcess, self).__init__(*args[:4])
 
-        self.ingestOffset = int(args[5] or 0)
-        self.ingestLimit = (int(args[4]) + self.ingestOffset) if args[4] else 5000
         self.fullImport = self.process == 'complete' 
-
-        self.db_manager = DBManager()
-        self.elastic_search_manager = ElasticsearchManager()
-        self.redis_manager = RedisManager()
 
         self.generateEngine()
         self.createSession()
 
         self.s3Bucket = os.environ['FILE_BUCKET']
-        self.s3_manager = S3Manager()
+        self.createS3Client()
 
-    def runProcess(self):
-        
-        logger.info('Starting ISAC process...')
-
-        self.db_manager.generateEngine()
-        self.redis_manager.createRedisClient()
-        self.elastic_search_manager.createElasticConnection()
-        self.s3_manager.createS3Client()
-        
+    def run_process(self):    
         with open('ingestJSONFiles/chicagoISAC_metadata.json') as f:
-                chicagoISACData = json.load(f)
+            chicagoISACData = json.load(f)
 
         for metaDict in chicagoISACData:
-            self.processChicagoISACRecord(metaDict)
+            self.process_chicago_isac_record(metaDict)
 
         self.saveRecords()
         self.commitChanges()
 
-    def processChicagoISACRecord(self, record):
+        logger.info(f'Ingested {len(self.records)} ISAC records')
+
+    def process_chicago_isac_record(self, record):
         try:
             chicagoISACRec = ChicagoISACMapping(record)
             chicagoISACRec.applyMapping()
-            self.storePDFManifest(chicagoISACRec.record)
+            self.store_pdf_manifest(chicagoISACRec.record)
             self.addDCDWToUpdateList(chicagoISACRec)
             
-        except Exception as e:
-            logger.exception(e)
-            logger.warning(ChicagoISACError('Unable to process ISAC record'))
+        except Exception:
+            logger.exception(ChicagoISACError('Unable to process ISAC record'))
             
 
-    def storePDFManifest(self, record):
+    def store_pdf_manifest(self, record):
         for link in record.has_part:
             itemNo, uri, source, mediaType, flags = link[0].split('|')
 
@@ -68,7 +56,7 @@ class ChicagoISACProcess(CoreProcess):
                     self.s3Bucket, manifestPath
                 )
 
-                manifestJSON = self.generateManifest(record, uri, manifestURI)
+                manifestJSON = self.generate_manifest(record, uri, manifestURI)
 
                 self.createManifestInS3(manifestPath, manifestJSON)
 
@@ -81,12 +69,12 @@ class ChicagoISACProcess(CoreProcess):
                 ])
 
                 record.has_part.insert(0, linkString)
-                self.hasPartArrayElemToStringElem(record)
+                self.change_has_part_url_array_to_string(record)
 
                 break
     
     @staticmethod
-    def hasPartArrayElemToStringElem(record):
+    def change_has_part_url_array_to_string(record):
         for i in range(1, len(record.has_part)):
             if len(record.has_part[i]) > 1:
                 urlArray = record.has_part[i]
@@ -96,13 +84,8 @@ class ChicagoISACProcess(CoreProcess):
             else:
                 record.has_part[i] = ''.join(record.has_part[i])
 
-    def createManifestInS3(self, manifestPath, manifestJSON):
-        self.putObjectInBucket(
-            manifestJSON.encode('utf-8'), manifestPath, self.s3Bucket
-        )
-
     @staticmethod
-    def generateManifest(record, sourceURI, manifestURI):
+    def generate_manifest(record, sourceURI, manifestURI):
         manifest = WebpubManifest(sourceURI, 'application/pdf')
 
         manifest.addMetadata(
--- a/processes/ingest/chicago_isac.py
+++ b/processes/ingest/chicago_isac.py
@@ -1,11 +1,9 @@
 import json
 import os
-from requests.exceptions import HTTPError, ConnectionError
 
 from ..core import CoreProcess
-from mappings.core import MappingError
 from mappings.chicagoISAC import ChicagoISACMapping
-from managers import WebpubManifest
+from managers import DBManager, ElasticsearchManager, RedisManager, S3Manager, WebpubManifest
 from logger import createLog
 
 logger = createLog(__name__)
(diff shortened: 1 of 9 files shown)


processes/ingest/chicago_isac.py
```diff
@@ -44,9 +51,9 @@ def processChicagoISACRecord(self, record):
             self.storePDFManifest(chicagoISACRec.record)
             self.addDCDWToUpdateList(chicagoISACRec)
             
-        except (MappingError, HTTPError, ConnectionError, IndexError, TypeError) as e:
+        except Exception as e:
             logger.exception(e)
```
kylevillegas93, line 55: The nice thing with logger.exception is that it will log all of the error details for us, so I suggest adding a descriptive message here like unable to process ISAC record.

processes/ingest/chicago_isac.py
```diff
@@ -44,9 +51,9 @@ def processChicagoISACRecord(self, record):
...
-            logger.warn(ChicagoISACError('Unable to process ISAC record'))
+            logger.warning(ChicagoISACError('Unable to process ISAC record'))
```
kylevillegas93, line 56: I think we can remove this line. We log the exception above.

processes/ingest/chicago_isac.py
kylevillegas93, line 1: Suggest moving to snake case in this class as well as another refactor.

processes/ingest/chicago_isac.py
```diff
@@ -72,18 +79,22 @@ def storePDFManifest(self, record):
...
-                        record.has_part.pop(i)
-                        for elem in urlArray:
-                            record.has_part.append(elem)
-                    else:
-                        record.has_part[i] = ''.join(record.has_part[i])
+                self.hasPartArrayElemToStringElem(record)
+
                 break
+    
+    @staticmethod
```
kylevillegas93, line 88: Hmm does this have to be a static method?

processes/ingest/chicago_isac.py
```diff
@@ -72,18 +79,22 @@ def storePDFManifest(self, record):
...
+    def hasPartArrayElemToStringElem(record):
```
kylevillegas93, line 89: this method name is a bit convoluted. could we change this name to be clearer?

processes/ingest/chicago_isac.py
```diff
@@ -72,18 +79,22 @@ def storePDFManifest(self, record):
...
+        for i in range(1, len(record.has_part)):
+            if len(record.has_part[i]) > 1:
+                urlArray = record.has_part[i]
+                record.has_part.pop(i)
+                for elem in urlArray:
+                    record.has_part.append(elem)
+            else:
+                record.has_part[i] = ''.join(record.has_part[i])
 
     def createManifestInS3(self, manifestPath, manifestJSON):
```
kylevillegas93, line 99: There are other processes that use this method. Could we abstract it out?

processes/ingest/chicago_isac.py
```diff
@@ -1,33 +1,40 @@
...
 
 logger = createLog(__name__)
 
 class ChicagoISACProcess(CoreProcess):
 
     def __init__(self, *args):
-        super(ChicagoISACProcess, self).__init__(*args[:4])
 
         self.ingestOffset = int(args[5] or 0)
         self.ingestLimit = (int(args[4]) + self.ingestOffset) if args[4] else 5000
```
kylevillegas93, line 16: Are ingest limit and offset being properly used in this class?

processes/ingest/chicago_isac.py
```diff
@@ -1,33 +1,40 @@
...
+
         self.generateEngine()
         self.createSession()
 
-        # S3 Configuration
         self.s3Bucket = os.environ['FILE_BUCKET']
-        self.createS3Client()
+        self.s3_manager = S3Manager()
 
     def runProcess(self):
```
kylevillegas93, line 29: Similar to the other refactors, but we could add a log statement after the process has ran to log out the number of records ingested.

kylevillegas93:
Recommend also migrating the unit tests to snake case and refactoring accordingly.

processes/ingest/chicago_isac.py
mitri-slory: The ingest limit and offset isn't needed since this process utilizes a static file and is manual compared to the automatic processes.

processes/ingest/chicago_isac.py
```diff
@@ -1,33 +1,40 @@
...
-        super(ChicagoISACProcess, self).__init__(*args[:4])
 
         self.ingestOffset = int(args[5] or 0)
         self.ingestLimit = (int(args[4]) + self.ingestOffset) if args[4] else 5000
         self.fullImport = self.process == 'complete' 
 
-        # Connect to database
+        self.db_manager = DBManager()
+        self.elastic_search_manager = ElasticsearchManager()
+        self.redis_manager = RedisManager()
```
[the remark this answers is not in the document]
mitri-slory: I decided to stick with inheritance for now since it would make sense to switch to composition when we decided to make a full transition with the other processes

processes/ingest/chicago_isac.py
mitri-slory: Sure I can add a log statement.

processes/ingest/chicago_isac.py
```diff
@@ -1,33 +1,40 @@
...
         self.createSession()
 
-        # S3 Configuration
         self.s3Bucket = os.environ['FILE_BUCKET']
-        self.createS3Client()
+        self.s3_manager = S3Manager()
 
     def runProcess(self):
+        
+        logger.info('Starting ISAC process...')
```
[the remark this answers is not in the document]
mitri-slory: Agreed. I'll remove this log statement.

processes/ingest/chicago_isac.py
mitri-slory: Ok I can add a more descriptive message.

processes/ingest/chicago_isac.py
mitri-slory: Agreed

processes/ingest/chicago_isac.py
```diff
@@ -44,9 +51,9 @@ def processChicagoISACRecord(self, record):
...
             self.addDCDWToUpdateList(chicagoISACRec)
             
-        except (MappingError, HTTPError, ConnectionError, IndexError, TypeError) as e:
+        except Exception as e:
             logger.exception(e)
-            logger.warn(ChicagoISACError('Unable to process ISAC record'))
+            logger.warning(ChicagoISACError('Unable to process ISAC record'))
             
 
     def storePDFManifest(self, record):
```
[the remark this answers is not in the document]
mitri-slory: The methods appear similar but they have different logic in them that pertain to each type of process. It's best to keep them separate for now.

processes/ingest/chicago_isac.py
mitri-slory: Not necessarily but the object itself isn't changing so there isn't a reason for it to be an object class method.

processes/ingest/chicago_isac.py
mitri-slory: I can change the name to `change_has_part_url_array_to_string`

processes/ingest/u_of_m.py
```diff
@@ -166,11 +166,6 @@ def storePDFManifest(self, record):
                     record.has_part.insert(0, linkString)
                     break
 
-    def createManifestInS3(self, manifestPath, manifestJSON):
-        self.putObjectInBucket(
-            manifestJSON.encode('utf-8'), manifestPath, self.s3Bucket
-        )
-
```
kylevillegas93, line 173: woo!

processes/ingest/chicago_isac.py
kylevillegas93, line 1: Overall looks good - let's go one step further and modify the variables to be snake case as well

tests/unit/processes/ingest/test_chicago_isac_process.py
kylevillegas93, line 1: Same as above let's change the method names and variable names to snake case

kylevillegas93 (changes requested):
Looking good! removing the duplicate code from the ingests is awesome. 

Couple more minor things and i think we will be good to go.

processes/ingest/chicago_isac.py
```diff
@@ -15,41 +13,38 @@ class ChicagoISACProcess(CoreProcess):
...
-        # Connect to database
         self.generateEngine()
         self.createSession()
 
-        # S3 Configuration
         self.s3Bucket = os.environ['FILE_BUCKET']
         self.createS3Client()
 
-    def runProcess(self):
+    def run_process(self):
```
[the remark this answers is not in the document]
mitri-slory: Ok I'll change it back to camelcase.

processes/ingest/chicago_isac.py
mitri-slory: Sure I can do that.

tests/unit/processes/ingest/test_chicago_isac_process.py
mitri-slory: Sure I can do that

tests/unit/processes/ingest/test_chicago_isac_process.py
```diff
@@ -14,91 +14,91 @@ def teardown_class(cls):
...
-        assert testManifest == 'testJSON'
-        assert mockManifest.links[0] == {'rel': 'self', 'href': 'manifestURI', 'type': 'application/webpub+json'}
+        assert test_manifest == 'test_json'
+        assert mock_manifest.links[0] == {'rel': 'self', 'href': 'manifest_url', 'type': 'application/webpub+json'}
 
-        mockManifest.addMetadata.assert_called_once_with(mockRecord, conformsTo='test_profile_uri')
-        mockManifest.addChapter.assert_called_once_with('sourceURI', 'testTitle')
\ No newline at end of file
+        mock_manifest.addMetadata.assert_called_once_with(mock_record, conformsTo='test_profile_uri')
+        mock_manifest.addChapter.assert_called_once_with('source_url', 'test_title')
```
kylevillegas93, line 104: nit: newline at the end of the file

tests/unit/processes/ingest/test_chicago_isac_process.py
```diff
@@ -14,91 +14,91 @@ def teardown_class(cls):
...
-    def test_createManifestInS3(self, testProcess, mocker):
-        mockPut = mocker.patch.object(ChicagoISACProcess, 'putObjectInBucket')
+    def test_create_manifest_in_S3(self, test_process, mocker):
+        mock_put = mocker.patch.object(ChicagoISACProcess, 'putObjectInBucket')
         
-        testProcess.createManifestInS3('testPath', '{"data": "testJSON"}')
+        test_process.createManifestInS3('test_path', '{"data": "test_json"}')
 
-        mockPut.assert_called_once_with(b'{"data": "testJSON"}', 'testPath', 'test_aws_bucket')
+        mock_put.assert_called_once_with(b'{"data": "test_json"}', 'test_path', 'test_aws_bucket')
```
kylevillegas93, line 89: I think we can remove this test as this method is no longer part of the CHicagoISACProcess class

processes/ingest/chicago_isac.py
```diff
@@ -15,95 +13,91 @@ class ChicagoISACProcess(CoreProcess):
     def __init__(self, *args):
         super(ChicagoISACProcess, self).__init__(*args[:4])
 
-        self.ingestOffset = int(args[5] or 0)
-        self.ingestLimit = (int(args[4]) + self.ingestOffset) if args[4] else 5000
-        self.fullImport = self.process == 'complete' 
+        self.full_import = self.process == 'complete'
```
kylevillegas93, line 16: Is this full_import field being used. If not, feel free to remove!

kylevillegas93 (approved):
Couple minor changes before merging in. 

Nice work!

tests/unit/processes/ingest/test_chicago_isac_process.py
mitri-slory: That's true. I'll remove it.

PR #413, opened 2024-10-23
SFR-2284: Improve Hathi Trust logging

kylevillegas93:
## Description
- Noticed that there were a lot of Read Row logs
- Improves logging and exception handling in Hathi Trust logging

## Testing
`make unit`


mitri-slory (approved):
This is a good change

PR #414, opened 2024-10-24
SFR-2280: Limit Hathi Records Ingested

mitri-slory:
This PR focuses on limiting the number of Hathi records ingested to 200 a day for now from 1000 a day to address current issues with the unfrbrized and unclustered records in QA. I also modified the `readHathiFile` method to match the method equivalent in the `seed_local_data` process because the language used there is more clear as to what is occurring in the method. This is step one of attempting to limit the amount of Hathi records ingested daily and the next step will be to address the backfill of Hathi files in QA that take priority over more recent records from other sources in the frbrization and cluster process.


processes/ingest/hathi_trust.py
```diff
@@ -16,7 +16,7 @@ class HathiTrustProcess(CoreProcess):
     HATHI_RIGHTS_SKIPS = ['ic', 'icus', 'ic-world', 'und']
 
     def __init__(self, *args):
-        super(HathiTrustProcess, self).__init__(*args[:4], batchSize=1000)
+        super(HathiTrustProcess, self).__init__(*args[:4], batchSize=200)
```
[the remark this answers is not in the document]
kylevillegas93: +1

mitri-slory:
The most recent commit I made implemented the feedback you both gave me and I updated the code to refelct that. The book[14] element represents the date of last update field of the Hathi data rows which is described here: [https://www.hathitrust.org/member-libraries/resources-for-librarians/data-resources/hathifiles/hathifiles-description/#:~:text=of%20Reason%20Codes.-,Date%20of%20last%20update,-rights_timestamp](url)

I will be fixing the formatting and cleaning up the code in a future commit.

processes/ingest/hathi_trust.py
```diff
@@ -72,7 +72,15 @@ def importFromHathiTrustDataFile(self, fullDump=False):
         )
 
         for hathiFile in fileJSON:
+            print(hathiFile)
```
kylevillegas93, line 75: nit: just remember to remove this print statement before we merge in

processes/ingest/hathi_trust.py
```diff
@@ -99,23 +99,12 @@ def importFromHathiFile(self, hathiURL):
             self.readHathiFile(hathiTSV)
 
     def readHathiFile(self, hathiTSV):
-        processed_record_count = 0
 
-        while True:
-            if self.ingest_limit and processed_record_count >= self.ingest_limit:
+        for number_of_books_ingested, book in enumerate(hathiTSV):
```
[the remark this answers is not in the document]
kylevillegas93: +1 this looks great!

processes/ingest/hathi_trust.py
```diff
@@ -99,23 +107,14 @@ def importFromHathiFile(self, hathiURL):
...
-            except csv.Error as e:
-                logger.warning('Unable to read TSV row')
-                logger.debug(e)
-                continue
-            except StopIteration:
-                logger.info('Reached end of TSV file')
-                break
+            book_right = book[2]
+            book_date_updated = book[14]
+            start_date_time = datetime.now(timezone.utc).replace(tzinfo=None) - timedelta(days=1)
```
kylevillegas93, line 116: There could be some edge cases here where the start_date_time slightly changes as we are enumerating over the TSV.  I recommend setting it once above and passing it through the functions as a param or setting it as a class variable.

processes/ingest/hathi_trust.py
```diff
@@ -99,23 +107,14 @@ def importFromHathiFile(self, hathiURL):
...
-                logger.warning('Unable to read TSV row')
-                logger.debug(e)
-                continue
-            except StopIteration:
-                logger.info('Reached end of TSV file')
-                break
+            book_right = book[2]
+            book_date_updated = book[14]
+            start_date_time = datetime.now(timezone.utc).replace(tzinfo=None) - timedelta(days=1) 
+            hathi_date_modified = datetime.strptime(book[14],'%Y-%m-%d %H:%M:%S')
```
kylevillegas93, line 117: Should we use our function here? 

`returnHathiDateFormat`

processes/ingest/hathi_trust.py
kylevillegas93: Also we should only set the start_date_time and compare each book updated at if it's a daily run.

processes/ingest/hathi_trust.py
```diff
@@ -72,7 +72,15 @@ def importFromHathiTrustDataFile(self, fullDump=False):
         )
 
         for hathiFile in fileJSON:
+            print(hathiFile)
             if hathiFile['full'] == fullDump:
+                if hathiFile['full'] == False:
```
kylevillegas93, line 77: Hmm - I think we could simplify this boolean logic a bit. Basically, we are saying if we are not doing a full dump and the file is not full then ingest records using a start date time.  If we are doing a full dump and the hathiFile is full, then import the whole file. 

This reads nested if statement reads if the hathiFile full field is equal to our fullDump param and the hathiFile full field is false, then ingest using start date time. 

I also recommending using .get('full') just to guard against missing fields.

processes/ingest/hathi_trust.py
```diff
@@ -99,23 +107,14 @@ def importFromHathiFile(self, hathiURL):
...
-                logger.info('Reading row')
-            except csv.Error as e:
-                logger.warning('Unable to read TSV row')
-                logger.debug(e)
-                continue
-            except StopIteration:
-                logger.info('Reached end of TSV file')
-                break
+            book_right = book[2]
+            book_date_updated = book[14]
```
kylevillegas93, line 115: Recommend guarding against index out of bounds exceptions by checking the length of the array first.

processes/ingest/hathi_trust.py
```diff
@@ -99,23 +107,14 @@ def importFromHathiFile(self, hathiURL):
...
+            start_date_time = datetime.now(timezone.utc).replace(tzinfo=None) - timedelta(days=1) 
+            hathi_date_modified = datetime.strptime(book[14],'%Y-%m-%d %H:%M:%S')
 
-            if row is not None and row[2] not in self.HATHI_RIGHTS_SKIPS:
-                self.parseHathiDataRow(row)
-                processed_record_count += 1
+            if book_right is not None and book_right not in self.HATHI_RIGHTS_SKIPS and hathi_date_modified > start_date_time:
```
kylevillegas93, line 119: you can simplify this to say book_right an book_right not in self.HATHI_RIGHTS_SKIPS

processes/ingest/hathi_trust.py
mitri-slory: Will do

processes/ingest/hathi_trust.py
mitri-slory: The 'updated dates' field for the hathi data rows usually have timestamps in this format 2017-02-16 05:58:14 which is a timestamp format of '%Y-%m-%d %H:%M:%S' that doesn't match any format in the current `returnHathiDateFormatMethod`. This is why I manually placed the timestamp format in the datetime object.

processes/ingest/hathi_trust.py
mitri-slory: That makes sense to to consider that case with the start_data_time changing so I will add it as a variable that's passed on to other functions. For your second comment, does this code line in the importFromHathiTrustDataFile not accomplish this comparison?

```
if hathi_date_modified > start_date_time:
                        self.importFromHathiFile(hathiFile['url'])
                        break
```

processes/ingest/hathi_trust.py
mitri-slory: Good suggestion Kyle.

processes/ingest/hathi_trust.py
mitri-slory: That's a good check to make to avoid that exception.

processes/ingest/hathi_trust.py
mitri-slory: These are all good suggestions Kyle. Thank you and I'll implement them in my next commit.

processes/ingest/hathi_trust.py
kylevillegas93: It does for sure but we can pass that start_date_time into importFromHathiFile

processes/ingest/hathi_trust.py
kylevillegas93: Copy!

processes/ingest/hathi_trust.py
```diff
@@ -36,44 +37,50 @@ def runProcess(self):
 
         logger.info(f'Ingested {len(self.records)} Hathi Trust records')
 
-    def importRemoteRecords(self, fullOrPartial=False):
```
kylevillegas93, line 39: nit: fullOrPartial is somewhat confusing as a variable name with a boolean type. When this value is true, does that mean we are doing a full or partial import? 

Recommend changing this variable name to full like the other ingest processes.

processes/ingest/hathi_trust.py
```diff
@@ -36,44 +37,50 @@ def runProcess(self):
...
-        self.importFromHathiTrustDataFile(fullDump=fullOrPartial)
+    def importRemoteRecords(self, start_date_time=None, full_or_partial=False):
+        self.importFromHathiTrustDataFile(start_date_time, full_dump=full_or_partial)
```
kylevillegas93, line 41: Also this is just a pass through to another function in this class. Can we just call importFromHathiTrustDataFile?

processes/ingest/hathi_trust.py
```diff
@@ -36,44 +37,50 @@ def runProcess(self):
...
-            if hathiFile['full'] == fullDump:
-                self.importFromHathiFile(hathiFile['url'])
+        for hathi_file in file_json:
+            if hathi_file.get('full') == full_dump and hathi_file.get('full') == False:
+                hathi_date_format = self.returnHathiDateFormat(hathi_file.get('modified'))
+                hathi_date_modified = datetime.strptime(hathi_file.get('modified'), hathi_date_format).replace(tzinfo=None)
+                if hathi_date_modified > start_date_time:
+                    self.importFromHathiFile(hathi_file.get('url'), start_date_time)
+                    break       
+            if hathi_file.get('full') == full_dump and hathi_file.get('full') == False:
```
kylevillegas93, line 82: Hmm looks like this is the same if statement as above? Is that correct? 

`if hathi_file.get('full') == full_dump and hathi_file.get('full') == False:`

processes/ingest/hathi_trust.py
```diff
@@ -36,44 +37,50 @@ def runProcess(self):
...
                 self.returnHathiDateFormat(x['created'])
             ).timestamp(),
             reverse=True
         )
 
-        for hathiFile in fileJSON:
-            if hathiFile['full'] == fullDump:
-                self.importFromHathiFile(hathiFile['url'])
+        for hathi_file in file_json:
+            if hathi_file.get('full') == full_dump and hathi_file.get('full') == False:
```
kylevillegas93, line 76: can you simplify this to`not hathi_file.get('full') && not full_dump`?

processes/ingest/hathi_trust.py
kylevillegas93: Should this be `elif hathi_file.get('full') and full_dump`

processes/ingest/hathi_trust.py
```diff
@@ -36,44 +37,50 @@ def runProcess(self):
...
+                hathi_date_format = self.returnHathiDateFormat(hathi_file.get('modified'))
+                hathi_date_modified = datetime.strptime(hathi_file.get('modified'), hathi_date_format).replace(tzinfo=None)
+                if hathi_date_modified > start_date_time:
+                    self.importFromHathiFile(hathi_file.get('url'), start_date_time)
+                    break       
+            if hathi_file.get('full') == full_dump and hathi_file.get('full') == False:
+                self.importFromHathiFile(hathi_file.get('url'), start_date_time)
```
kylevillegas93, line 83: If this is the full ingest case, should we pass start_date_time here?

processes/ingest/hathi_trust.py
```diff
@@ -85,37 +92,38 @@ def returnHathiDateFormat(strDate):
...
-            try:
-                row = next(hathiTSV)
-                logger.info('Reading row')
-            except csv.Error as e:
-                logger.warning('Unable to read TSV row')
-                logger.debug(e)
+            
+            if len(book) >= 2:
+                book_right = book[2]
+                if len(book) >= 15:
```
kylevillegas93, line 115: if start_date_time is None should we continue if we aren't able to get the book_date_updated?

processes/ingest/hathi_trust.py
```diff
@@ -85,37 +92,38 @@ def returnHathiDateFormat(strDate):
...
+                    book_date_updated = book[14]
+                else:
+                    continue
```
kylevillegas93, line 118: You can simplify this to: 

```
book_right = (len(book) > 2 and book[2]) or None
book_date_updated = (len(book) > 14 and book[14]) or None
```

processes/ingest/hathi_trust.py
```diff
@@ -85,37 +92,38 @@ def returnHathiDateFormat(strDate):
...
                 break
-
-            try:
-                row = next(hathiTSV)
-                logger.info('Reading row')
-            except csv.Error as e:
-                logger.warning('Unable to read TSV row')
-                logger.debug(e)
+            
+            if len(book) >= 2:
```
kylevillegas93, line 113: There's a bug here if the length of a book array is 2 but we try to access book[2]

processes/ingest/hathi_trust.py
```diff
@@ -85,37 +92,38 @@ def returnHathiDateFormat(strDate):
...
-                self.parseHathiDataRow(row)
-                processed_record_count += 1
+            hathi_date_modified = datetime.strptime(book_date_updated, '%Y-%m-%d %H:%M:%S').replace(tzinfo=None)
+
+            if start_date_time:
+                if book_right and book_right not in self.HATHI_RIGHTS_SKIPS and hathi_date_modified > start_date_time:
+                    self.parseHathiDataRow(book)
+            else:
+                if book_right and book_right not in self.HATHI_RIGHTS_SKIPS:
+                    self.parseHathiDataRow(book)
```
kylevillegas93, line 129: I think since there are two common conditions with the book rights, you can simplify to: 

```
if book_right and book_right not in self.HATHI_RIGHTS_SKIPS:
    if not start_date_time or hathi_date_modified > start_date_time:
        self.parseHathiDataRow(book)
```

Does that look right? Recommend reducing duplicate code.

processes/ingest/hathi_trust.py
mitri-slory: It is a confusing variable anytime I've looked at it and I believe it should be just full since for the complete process we are only ingesting hathifiles with a full field of true and a full field of false for the daily process, so the partial portion is unnecessary. I agree to rename the variable.

processes/ingest/hathi_trust.py
mitri-slory: I agree that the method here is unnecessary so I'll remove it and call `importFromHathiTrustDataFile` directly in the `runProcess` method.

processes/ingest/hathi_trust.py
mitri-slory: Oh thanks for cataching that. That conditional should have a True boolean value for second half of the and statement.

`elif hathi_file.get('full') == full_dump and hathi_file.get('full') == True:`

I'll add an else statement after that one to continue the for loop to catch the edge case where a hathi file doesn't have a boolean value or doesn't exist.

processes/ingest/hathi_trust.py
mitri-slory: Oh I get what you mean. That's a cleaner way to explain this same logic.

processes/ingest/hathi_trust.py
mitri-slory: You're right since the importFromHathiFile method already sets start_date_time to None if no parameter is inherited from another method.

processes/ingest/hathi_trust.py
mitri-slory: The else continue conditional for the `book_date_updated` variable is unnecessary since we we should only check if the variable exists before we set the `hathi_date_modified` variable. I'll remove the nested else conditional and make a new if conditional for the `hathi_date_modified` variable like this:

```
if book_date_updated:
                hathi_date_modified = datetime.strptime(book_date_updated, '%Y-%m-%d %H:%M:%S').replace(tzinfo=None)
```

processes/ingest/hathi_trust.py
mitri-slory: That's a good way of simplifying this code.

processes/ingest/hathi_trust.py
mitri-slory: Thanks for catching the typo. It should be `len(book) >= 3`

processes/ingest/hathi_trust.py
mitri-slory: I will remove the duplicate code. Good way to simplify the conditionals here. Definitely makes it more clear to read.

processes/ingest/hathi_trust.py
mitri-slory: Looking at it again I believe `True` represented full and `False` represented partial.

processes/ingest/hathi_trust.py
```diff
@@ -36,45 +37,50 @@ def runProcess(self):
...
+        for hathi_file in file_json:
+            if not hathi_file.get('full') and not full_dump:
+                hathi_date_format = self.returnHathiDateFormat(hathi_file.get('modified'))
+                hathi_date_modified = datetime.strptime(hathi_file.get('modified'), hathi_date_format).replace(tzinfo=None)
+                if start_date_time and hathi_date_modified > start_date_time:
+                    self.importFromHathiFile(hathi_file.get('url'), start_date_time)
+                    break       
+            elif hathi_file.get('full') and full_dump:
+                self.importFromHathiFile(hathi_file.get('url'))
                 break
```
kylevillegas93, line 81: Do we want to break here or do we want to import the file and move on to the next file?

mitri-slory: I kept the break because the original code here had a break after we called importFromHathiFile which I'm only assuming was put here so only one hathifile would be imported from all the hathifiles since they could be so massive. I think it's fine to remove them since we will only ingest hathi data rows that were updated in the past day so there won't be as much hathi rows ingested in general. What are your thoughts?

```
if hathiFile['full'] == fullDump:
                        self.importFromHathiFile(hathiFile['url'])
                        break
```

kylevillegas93 (approved):

PR #418, opened 2024-10-25
SFR-2283: Refactoring catalog process

kylevillegas93:
## Description
- Refactors catalog process
- Moves away from while true loop 
- Converts to snake case

## Testing
`make test`


processes/catalog.py
```diff
@@ -25,71 +25,52 @@ def __init__(self, *args):
...
-    def parseCatalogRecord(self, catalogXML, oclcNo, owiNo):
+    def parse_catalog_record(self, catalog_record, oclc_number, owi_number):
         try:
-            parseMARC = etree.fromstring(catalogXML.encode('utf-8'))
-        except etree.XMLSyntaxError as e:
-            logger.error(f'OCLC catalog MARC xml is invalid for OCLC number: {oclcNo}')
-            logger.debug(e)
+            parsed_marc_xml = etree.fromstring(catalog_record.encode('utf-8'))
+        except Exception:
+            logger.exception(f'Unable to parse OCLC catalog MARC XML for OCLC number: {oclc_number}')
```
Apophenia, line 65: Is it worth leaving in the debug level error so that we can see the way etree failed, if it fails? I suppose it's probably a rare case

Apophenia (approved):
Thanks, I think these changes look good!

processes/catalog.py
kylevillegas93: We will see the error in the exception log! However, I think if we are debugging locally we can add more debug logs. Since we set the log level as INFO for both QA and prod, these logs never get logged out.

PR #420, opened 2024-10-25
SFR-2285: Fix has part item build

kylevillegas93:
## Description
- Extends the number of items by the maximum number in the has_part list of the record. 

## Testing
- Tested against a record in QA that failed to cluster due to an index out of range error 

https://drb-qa.nypl.org/work/b21b52f1-e6f1-4363-8737-c708616b71b3

### DIFF
--- a/managers/sfrRecord.py
+++ b/managers/sfrRecord.py
@@ -258,10 +258,13 @@ class SFRRecordManager:
         editionData['dcdw_uuids'].append(rec.uuid.hex)
 
     def buildItems(self, editionData, rec, itemContributors):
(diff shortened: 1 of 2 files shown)


mitri-slory (approved):
This is a good catch to fixing this method.

kylevillegas93:
@mitri-slory great question. There are cases where the indices of the has_part array do not align with the number of parts. So for instance, there may be two parts but the indices are 2 and 3. In this case we would only be extending the array by 2 but we get an IndexOutOfBounds exception because of the has_part indices are greater than what we extended the array by.

PR #422, opened 2024-10-29
SFR-2291: Throw error if we fail to generate webpub 

kylevillegas93:
## Description
- We don't want to silently fail to generate a webpub 
- If we can't, we should not acknowledge the message and retry the processing of the message

## Testing
`make unit`


mitri-slory (approved):
Good catch to make sure we see this error right away.

PR #424, opened 2024-10-30
SFR-2294: Only cluster records that have a title

kylevillegas93:
## Description
- We are still seeing some NYPL records that we fail to cluster because the title is None
- Adding a filter to filter out records where the title is None in the first place.


mitri-slory (approved):
Good catch and hopefully this fixes records with no titles being clustered.

Apophenia (approved):

PR #428, opened 2024-10-31
NO-REF: Only run unit tests if requirements or source code changes

kylevillegas93:
## Description
- We should not run unit tests on every change to the repo
- This change triggers the unit tests GH action workflow only if files change in certain folders
- We should see a reduction hopefully in cost


ayan1229 (approved):
🚀🚀

PR #433, opened 2024-11-06
NO-REF: Refactor NYPL Process

kylevillegas93:
## Description
- Abstracts NYPL bib service and simplifies process to get and save records 

## Testing
`make unit; make integration`


mitri-slory (approved):
These changes look good. Nice use of abstraction.

PR #437, opened 2024-11-08
NO-REF: add error handling error process init

kylevillegas93:
## Description
- Sometimes I see that a process is starting from main.py but no other logs from the process files
- Adding some error handling to see if initiatlization fails. 
- We have a lot of code in our constructors, so this is an outer try/except block that will catch errors that occur initializing our processes.

### DIFF
--- a/main.py
+++ b/main.py
@@ -36,10 +36,14 @@ def main(args):
 
     availableProcesses = registerProcesses()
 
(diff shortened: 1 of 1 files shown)


jackiequach (approved):
Looks good

Apophenia (commented):
Thank you, looks good!

PR #440, opened 2024-11-08
NO-REF: Rename base mapping classes

kylevillegas93:
## Description
- Core and AbstractMapping were not clear names
- Renames Core to BaseMapping
- Renames AbstractMapping to RecordMapping

## Testing
`make unit`


mitri-slory (approved):
Great name changes for these classes

PR #443, opened 2024-11-12
NO-REF: Remove RedisManager as an ancestor of CoreProcess

kylevillegas93:
## Description
- Removes RedisManager as an ancestor of CoreProcess
- Uses composition for the RedisManager instances with the processes that rely on this manager

## Testing
`make unit`


mitri-slory (approved):
Looks good.

PR #449, opened 2024-11-15
SFR-2306: Integrating with Airtable to get backlist records

kylevillegas93:
## Description
- Integrates with Airtable to get backlist records

## Testing 
`make integration` 
`pytest -k test_pub_backlist_process`


mitri-slory:
This recent commit only adds the offset to the conditional for when a complete process is run for right now. My next commit will be focused on using the filterByFormula parameter in the Airtbale API to retrieve records based on the Last Modified field I recently added to the UofMichigan view.

processes/ingest/publisher_backlist.py
```diff
@@ -10,24 +10,40 @@ class PublisherBacklistProcess(CoreProcess):
...
+            self.commitChanges()
+
+            logger.info(f'Ingested {len(self.records)} UofM records')
 
         except Exception as e:
             logger.exception('Failed to run Pub Backlist process')
-            raise e
+            raise e   
+        finally:
+            self.close_connection()
```
kylevillegas93, line 49: nit: new line!

processes/ingest/publisher_backlist.py
```diff
@@ -10,24 +10,40 @@ class PublisherBacklistProcess(CoreProcess):
...
-        self.ingest_limit = (int(args[4]) + self.ingestOffset) if args[4] else 5000
-        self.full_import = self.process == 'complete' 
+        self.limit = (len(args) >= 5 and args[4] and args(4) <= 100) or None
+        self.offset = (len(args) >= 6 and args[5]) or None
 
-        self.generateEngine()
-        self.createSession()
-
-        self.s3_bucket = os.environ['FILE_BUCKET']
+        self.s3Bucket = os.environ['FILE_BUCKET']
```
kylevillegas93, line 16: nit: s3_bucket for snake case

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,65 @@
...
+        self.airtable_auth_token = os.environ.get('AIRTABLE_KEY', None)
+
+        pass
+
+    def get_records(
+        self,
+        full_import: bool=False, # get all records from Airtable
+        start_timestamp: datetime=None,
+        offset: Optional[int]=None, # start at specific position in paginated results 
+        limit: Optional[int]=None
```
kylevillegas93, line 26: I think we can remove the comments here!

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,65 @@
...
+    ) -> list[UofMMapping]:
+        
+        json_data = self.get_publisher_backlist_records(full_import, start_timestamp, offset, limit)
```
kylevillegas93, line 29: recommend being more explicit in the variable naming here.

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,65 @@
...
+    def get_publisher_backlist_records(self, full_import=False, start_timestamp=None, offset=None, limit=None):
+
+        if offset == None:
+            limit = 100
+
+        headers = {"Authorization": f"Bearer {self.airtable_auth_token}"}
+
+        if not full_import:
+            if start_timestamp:
+                response = requests.get(f'https://api.airtable.com/v0/appBoLf4lMofecGPU/Publisher%20Backlists%20%26%20Collections%20%F0%9F%93%96?view=UofMichigan%20Press&pageSize={limit}', headers=headers)
```
kylevillegas93, line 47: I think there's a common base URL for all of these requests. let's pull it out and set it as a constant so we can do `requests.get(f'{BASE_URL}&pageSize={limit}')`

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,65 @@
...
+                return response.json()
+            else:
+                start_date_time = datetime.now(timezone.utc).replace(tzinfo=None) - timedelta(hours=24)
```
kylevillegas93, line 50: Recommend adding a TODO here to use the start_date_time

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,65 @@
...
+                response = requests.get(f'https://api.airtable.com/v0/appBoLf4lMofecGPU/Publisher%20Backlists%20%26%20Collections%20%F0%9F%93%96?view=UofMichigan%20Press&pageSize={limit}', headers=headers)
+                return response.json
+        else:
+            response = requests.get(f'https://api.airtable.com/v0/appBoLf4lMofecGPU/Publisher%20Backlists%20%26%20Collections%20%F0%9F%93%96?view=UofMichigan%20Press', headers=headers)
+            response_json = response.json()
+            array_json = [response_json]
```
kylevillegas93, line 56: Would this be a list of lists?

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,65 @@
...
+        offset: Optional[int]=None, # start at specific position in paginated results 
+        limit: Optional[int]=None
+    ) -> list[UofMMapping]:
+        
+        json_data = self.get_publisher_backlist_records(full_import, start_timestamp, offset, limit)
+
+        array_index = 0
+        while array_index != len(json_data):
+            for dict_index in range(0, len(json_data['records'])):
+                metaDict = json_data['records'][dict_index]
```
kylevillegas93, line 34: I recommend actually putting this down in the other method so that this function only gets a flat list of json records and then can map those records.

services/__init__.py
```diff
@@ -1 +1,2 @@
 from .sources.nypl_bib_service import NYPLBibService
+from .sources.publisher_backlist_service import PublisherBacklistService
```
kylevillegas93, line 2: nit: newline here too

processes/ingest/publisher_backlist.py
```diff
@@ -10,24 +10,40 @@ class PublisherBacklistProcess(CoreProcess):
...
+            self.createSession()
+
+            if self.process == 'daily':
+                records = self.publisher_backlist_service.get_records(offset=self.offset, limit=self.limit)
+            elif self.process == 'complete':
+                records = self.publisher_backlist_service.get_records(full_import=True)
+            elif self.process == 'custom':
+                records = self.publisher_backlist_service.get_records(start_timestamp=self.ingestPeriod, offset=self.offset, limit=self.limit)
+            else: 
+                logger.warning(f'Unknown NYPL ingestion process type {self.process}')
```
kylevillegas93, line 33: nit: Update ingestion process name

services/sources/publisher_backlist_service.py
mitri-slory: I agree

services/sources/publisher_backlist_service.py
mitri-slory: I agree and I was planning to clean up the response urls in future commits as they were too long and shared a common base.

services/sources/publisher_backlist_service.py
mitri-slory: I'm gonna use the start_date_time in my next commit so a TODO won't be necessary anymore.

services/sources/publisher_backlist_service.py
mitri-slory: I'm kinda confused here since if I move this code down to the other method wouldn't I have to return the `metaDict `to `get_records()` and therefore the conditional would end after only returning the first record json file in the array?

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,92 @@
...
+        limit: Optional[int]=None
+    ) -> list[UofMMapping]:
+        
+        array_json_records = self.get_records_json(full_import, start_timestamp, offset, limit)
+
+        array_index = 0
+        while array_index != len(array_json_records):
+            for dict_index in range(0, len(array_json_records[array_index]['records'])):
+                meta_dict = array_json_records[array_index]['records'][dict_index]
+                raise Exception
```
Apophenia, line 36: I would recommend adding exception details when raising an exception.

Apophenia (commented):
Hi Dmitri, thanks for opening this PR. I'm really glad to see work on the Airtable side of things.

In general, I'm wondering if we can refine `get_records` and `get_records_json` to be a little more concise and readable. In `get_records_json` I see a lot of control flow blocks that have similar code content, but slightly different configurations of variables. Is it possible to generalize some of this functionality, and to pull out and separately declare repeated variable declarations (e.g. is it possible to name the formulas and not repeat them)?

In `get_records` I see a type hint, which is great (I think it would be good to add type hints to `get_records_json` too. I think the updated variable name in line 30 is helpful, but it's still difficult for me to follow the logic in this function. And although the type hint is there, I don't see anything returned.

At a high level, I think it's hard for me to understand the data structure manipulation happening here when I don't have an example of the Airtable json response to hand. I think giving things like `meta_dict` some more specific names that describe not just the data structure, but what's contained in those structures, could help add some clarity for someone who is just glancing at the file.

services/sources/publisher_backlist_service.py
mitri-slory: I mainly added that raise Exception for testing purposes when running the process to make sure it didn't run through all the records and to make sure it got to that point in the code at least. It won't be there in the final version of this ingestion purpose.

processes/ingest/publisher_backlist.py
```diff
@@ -10,24 +10,43 @@ class PublisherBacklistProcess(CoreProcess):
...
+            logger.info(f'Ingested {len(self.records)} Publisher Backlist records')
 
         except Exception as e:
-            logger.exception('Failed to run Pub Backlist process')
-            raise e
+            logger.exception('Failed to run Publisher Backlist process')
+            raise e   
+        finally:
+            self.close_connection()
+
```
kylevillegas93, line 52: There may be a \t character here - recommend removing it so Github stops complaining.

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,115 @@
...
+from typing import Optional, Dict
+
+from logger import create_log
+from mappings.UofM import UofMMapping
+from .source_service import SourceService
+
+
+logger = create_log(__name__)
+
+BASE_URL = "https://api.airtable.com/v0/appBoLf4lMofecGPU/Publisher%20Backlists%20%26%20Collections%20%F0%9F%93%96?view=UofMichigan%20Press"
```
kylevillegas93, line 15: Based on our conversation on Wednesday - do we want to update this to look at the all lists view?

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,115 @@
...
+        if not start_timestamp:
+            start_timestamp = datetime.now(timezone.utc).replace(tzinfo=None) - timedelta(hours=24)
+
+        start_date_time_str = start_timestamp.strftime("%Y-%m-%d %H:%M:%S.%f")
+        start_date_time_encoded = urllib.parse.quote(start_date_time_str)
+        is_same_date_time_filter = f"IS_SAME(%7BLast%20Modified%7D,%20%22{start_date_time_encoded}%22,%20%22second%22"
+        is_after_date_time_filter = f"%20IS_AFTER(%7BLast%20Modified%7D,%20%22{start_date_time_encoded}%22"
+        
+        filter_by_formula = f"OR({is_same_date_time_filter}),{is_after_date_time_filter}))"
+        return filter_by_formula
```
kylevillegas93, line 90: Nice - this looks good. Do the Airtable last modified dates have seconds to them as well? Could we simplify this to do is after without the seconds or even minutes? 

Nit: recommend adding a newline between the body of this function and the return statement for readability.

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,115 @@
...
+    def get_records_array(self, 
+        headers: Dict[str, str], 
+        limit: Optional[int]=None, 
+        filter_by_formula: str=None,
+    ) -> list[dict]:
+        
+        if filter_by_formula: 
+            url_format = f'{BASE_URL}&filterByFormula={filter_by_formula}&pageSize={limit}'
+        else: 
+            url_format = f'{BASE_URL}&pageSize={limit}'
```
kylevillegas93, line 101: looks like you can simplify this because we are adding the pageSize in both cases. 

```
url = f'{BASE_URL}&pageSize={limit}

if filter_by_formula:
  url += f'&filterByFormula{filter_by_formula}'

File: services/sources/publisher_backlist_service.py [hunk 16]
```diff
@@ -0,0 +1,115 @@
...
+
+        response = requests.get(url_format, headers=headers)
+        response_json = response.json()
+        array_json = [response_json]
+        
+        while 'offset' in response_json:
+            offset_parameter = f"&offset={response_json['offset']}"
+            complete_url = url_format + offset_parameter
```
kylevillegas93, line 109: Recommend changing complete_url to something like next_page_url to be clear that we are getting the next page of results. 

Could we do this concatenation in 1-line without the offset_parameter variable?

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,115 @@
...
+        limit: Optional[int]=None, 
+        filter_by_formula: str=None,
+    ) -> list[dict]:
+        
+        if filter_by_formula: 
+            url_format = f'{BASE_URL}&filterByFormula={filter_by_formula}&pageSize={limit}'
+        else: 
+            url_format = f'{BASE_URL}&pageSize={limit}'
+
+        response = requests.get(url_format, headers=headers)
```
kylevillegas93, line 103: you can be more explicit here like - publisher_backlist_records_response

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,115 @@
...
+        response_json = response.json()
+        array_json = [response_json]
```
kylevillegas93, line 105: Is the response_json a list itself? Could we just append the next page responses to this list?

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,115 @@
...
+        full_import: bool=False, 
+        start_timestamp: datetime=None,
+        offset: Optional[int]=None,
+        limit: Optional[int]=None
+    ) -> list[UofMMapping]:
+        
+        array_json_records = self.get_records_json(full_import, start_timestamp, offset, limit)
+
+        array_index = 0
+        while array_index != len(array_json_records):
```
kylevillegas93, line 32: could we use a for loop here instead to remove the array_index increment line?

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,115 @@
...
+            for dict_index in range(0, len(array_json_records[array_index]['records'])):
```
kylevillegas93, line 33: Could we enumerate here to get the index/key/item and make this easier?

services/sources/publisher_backlist_service.py
mitri-slory: I'm fine with changing the view to All lists view since we clarified which view we will be using for the process in the future.

services/sources/publisher_backlist_service.py
mitri-slory: The Airtable last modified dates do have seconds in their response like this: `2024-06-20T17:22:51.000Z` I just tested out the IS_SAME filter in Insomnia and unit of time like seconds aren't required in the filter parameter which isn't stated in the Airtable formulas sheet on their website unlike other filters that detail when a parameter is optional. The unit of time details how we should match the dates up to a certain unit like hours, minutes, and seconds. Therefore, I can get rid of the unit of time parameter in that filter to simplify the filter.

services/sources/publisher_backlist_service.py
mitri-slory: I agree with the name change to make the purpose of the variable more clear. I can condense these lines into just one line too.

services/sources/publisher_backlist_service.py
mitri-slory: Sure I can do that

services/sources/publisher_backlist_service.py
mitri-slory: Sure

services/sources/publisher_backlist_service.py
mitri-slory: I don't think we even need enumerate here. We just need the elements/values for both loops.

services/sources/publisher_backlist_service.py
kylevillegas93: nice!

tests/integration/services/sources/test_publisher_backlist_service.py
```diff
@@ -0,0 +1,19 @@
+from datetime import datetime, timezone, timedelta
+import pytest
+
+from load_env import load_env_file
+from services import PublisherBacklistService
+
+class TestPublisherBacklistService:
```
kylevillegas93, line 7: woo!

tests/unit/processes/ingest/test_pub_backlist_process.py
```diff
@@ -0,0 +1,106 @@
+import pytest
+
+from tests.helper import TestHelpers
+from processes import PublisherBacklistProcess
+
+
+class TestPublisherBacklistProcess:
```
kylevillegas93, line 7: this looks good to me nice work!

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,110 @@
...
+        self.airtable_auth_token = os.environ.get('AIRTABLE_KEY', None)
+
+    def get_records(
+        self,
+        full_import: bool=False, 
+        start_timestamp: datetime=None,
+        offset: Optional[int]=None,
+        limit: Optional[int]=None
+    ) -> list[UofMMapping]:
+
```
kylevillegas93, line 28: nit: for all of these functions, I recommend removing the new lines

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,110 @@
...
+        full_import: bool=False, 
+        start_timestamp: datetime=None,
+        offset: Optional[int]=None,
+        limit: Optional[int]=None
+    ) -> list[dict]:
+
+        if offset == None:
+            limit = 100
+
+        headers = {"Authorization": f"Bearer {self.airtable_auth_token}"}
```
kylevillegas93, line 52: I think let's move the headers down to the function that uses it.

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,110 @@
...
+
+            else:
+
+                filter_by_formula = self.build_filter_by_formula_parameter(start_timestamp)
+
+                array_json_records = self.get_records_array(headers, limit, filter_by_formula)
+
+                return array_json_records
+            
+        else:
```
kylevillegas93, line 71: since you are returning in the above if/else you can remove this else clause.

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,110 @@
...
+        start_timestamp: datetime=None,
+        offset: Optional[int]=None,
+        limit: Optional[int]=None
+    ) -> list[UofMMapping]:
+        
+        array_json_records = self.get_records_json(full_import, start_timestamp, offset, limit)
+
+
+        for json_dict in array_json_records:
+            for records_value in json_dict['records']:
```
kylevillegas93, line 33: I recommend that get_records_array does this logic to return the records. that way this function is only getting back a list of json records and has to simply map it.

services/sources/publisher_backlist_service.py
```diff
@@ -0,0 +1,110 @@
...
+                try:
+                    record_metadata_dict = records_value
+                    record = UofMMapping(record_metadata_dict)
+                    record.applyMapping()
+                except Exception as e:
+                    logger.exception(f'Failed to process Publisher Backlist record due to {e}')
```
kylevillegas93, line 39: you can remove the due to {e} from this exception log because exception will also log the error message by default.

services/sources/publisher_backlist_service.py
mitri-slory: Makes sense

services/sources/publisher_backlist_service.py
mitri-slory: I think I need to clarify what array_json_records consists of. It seems from what you're proposing that array_json_records looks like this:

`[{"records": [values from all records from airtable API]}]`

but in actuality it looks like this:

`[{"records": [value from first 100 records from airtable]}, {"records":[value from next 100 records from airtable]}]`

That's why I can't just return the values from the records key one time since there are several records due to their being multiple dictionaries in the array.

mitri-slory (approved):

PR #450, opened 2024-11-15
SFR-2316: Create a local backend for frontend CI

kylevillegas93:
## Description
- This change sets up a local backend for frontend CI
- It connects to elastic search via the VPC endpoint which will require some authorization/authentication via AWS sig4.
- It uses a combination of connecting to the QA database and elastic search index and a local redis. 

## Testing
`docker compose up redis; python main.py -p APIProcess -e frontend-ci`


jackiequach (approved):
This makes sense to me

PR #458, opened 2024-11-21
SFR-2352: Script to upload digital assets to Google Drive

kylevillegas93:
## Description
- Migrates files from S3 to google drive 

## Testing
`python scripts/migrate_digital_assets.py`

### DIFF
--- a/scripts/migrate_digital_assets.py
+++ b/scripts/migrate_digital_assets.py
@@ -5,46 +5,58 @@ from googleapiclient.discovery import build
 from googleapiclient.http import MediaIoBaseUpload
 from google.oauth2.service_account import Credentials
 
+# Please install the requirements with this command: pip install boto3 google-api-python-client google-auth
+
 s3 = boto3.client('s3')
+
+S3_BUCKET_NAME = 'ump-pdf-repository'
+SERVICE_ACCOUNT_FILE = 'service-account.json'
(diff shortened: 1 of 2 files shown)


scripts/migrate_digital_assets.py
```diff
@@ -0,0 +1,62 @@
...
+        print(f'Processing file: {file_key}')
+        
+        s3_object = s3.get_object(Bucket=s3_bucket, Key=file_key)
+        file_data = s3_object['Body'].read()
+        
+        try:
+            drive_file_id = upload_to_google_drive(drive_service, file_key, file_data)
+            print(f'Uploaded {file_key} to Google Drive with ID: {drive_file_id}')
+        except Exception as e:
+            print(f'Failed to upload {file_key} to Google Drive: {e}')
```
mitri-slory, line 55: Do we need the "e" variable here if Exception here will detail the reasons behind the error?

mitri-slory (approved):
Looks good and I only have one question on a try/except block.

scripts/migrate_digital_assets.py
kylevillegas93: yes! the print statement here won't know about the exception message by default.

mitri-slory: Oh that makes sense. A logger would show the exception message by default not a print statement.

kylevillegas93: exactly!

PR #459, opened 2024-11-25
SFR-2364: Remove DBManager as an ancestor of CoreProcess

kylevillegas93:
# Description
- Removes the last inheritance piece of the puzzle from the CoreProcess
- Adds a RecordBuffer class to handle bulk updating records in postgres

# Testing
- Tested ingest, classify, catalog, and cluster processes locally


tests/unit/processes/file/test_cover_process.py
```diff
@@ -11,8 +11,11 @@ class TestCoverProcess:
     def testProcess(self, mocker):
         class TestCoverProcess(CoverProcess):
             def __init__(self, *args):
+                self.editions = set()
+                self.db_manager = mocker.MagicMock()
                 self.fileBucket = 'test_aws_bucket'
-                self.batchSize = 3
+                self.batch_size = 3
+                self.ingestLimit = None
```
jackiequach, line 18: how do we decide if we should use snake case or camel case?

tests/unit/processes/ingest/test_hathi_trust_process.py
```diff
@@ -60,40 +62,31 @@ def start_date_time(self):
...
-        mockSave = mocker.patch.object(HathiTrustProcess, 'saveRecords')
-        mockCommit = mocker.patch.object(HathiTrustProcess, 'commitChanges')
 
         testInstance.process = 'complete'
         testInstance.runProcess()
 
         mockImport.assert_called_once
-        mockSave.assert_called_once
-        mockCommit.assert_called_once
+        testInstance.record_buffer.flush.assert_called_once
```
jackiequach, line 79: missing ()

tests/unit/processes/ingest/test_doab_process.py
```diff
@@ -61,53 +63,41 @@ def mockOAIQuery(self, mocker):
...
-        mockSave = mocker.patch.object(DOABProcess, 'saveRecords')
-        mockCommit = mocker.patch.object(DOABProcess, 'commitChanges')
 
         testProcess.process = 'daily'
         testProcess.runProcess()
 
         mockImport.assert_called_once
-        mockSave.assert_called_once
-        mockCommit.assert_called_once
+        testProcess.record_buffer.flush.assert_called_once
```
jackiequach, line 71: I assume the () is optional, but there should be consistency across files.

tests/unit/processes/ingest/test_nypl_process.py
```diff
@@ -18,89 +18,69 @@ def test_instance(self, mocker) -> NYPLProcess:
...
         test_instance.nypl_bib_service.get_records.return_value = record_mappings
 
         test_instance.process = 'daily'
         test_instance.runProcess()
 
         test_instance.nypl_bib_service.get_records.assert_called_once_with(offset=None, limit=None)
-        self.assert_common_expectations(mocks)
-        assert mocks['add_record'].call_count == len(record_mappings)
+        assert test_instance.record_buffer.flush.call_count == len(record_mappings)
+        test_instance.record_buffer.flush.assert_called_once
```
jackiequach, line 42: Same as above, a few instances in this file

tests/unit/processes/test_cluster_process.py
```diff
@@ -230,38 +206,31 @@ def test_cluster_record_error(self, testInstance, testRecord, mocker):
...
-        mockSession.query().filter().all.return_value = ['rec1', 'rec2', 'rec3']
-        testInstance.session = mockSession
+        testInstance.db_manager.session.query().filter().all.return_value = ['rec1', 'rec2', 'rec3']
         testEditions, testRecords = testInstance.cluster_matched_records([1, 2, 3]) 
 
         assert testEditions == ['ed1', 'ed2']
         assert testRecords == ['rec1', 'rec2', 'rec3']
 
-        mockSession.query.filter.all.assert_called_once
+        testInstance.db_manager.session.query.filter.all.assert_called_once
```
jackiequach, line 233: Same here

jackiequach (commented):
Overall looks good, just a few minor comments. Also think it would be beneficial for a backend dev to review this.

tests/unit/processes/file/test_cover_process.py
kylevillegas93: That's a good question. We are in the midst of transitioning over to snake case which is a common practice. You will cases where there is a mix of both. I will double check my snake case here tho!

tests/unit/processes/ingest/test_doab_process.py
kylevillegas93: I think it is but it's a good callout - I'll make those changes!

jackiequach (approved):
missed one in test_cluster_process.py, but otherwise looks good

PR #460, opened 2024-11-25
SFR-2308_PubBacklistMapping

mitri-slory:
This PR is focused on creating a unified mapping for the Publisher Backlist Project records. One issue I'm running into is that the 'Projects' field when retrieved from Airtable comes back as a code like this ['recaXuQYVJMiFs6D8'] instead of ['MIT Press']. I would like to receive some help on how to access the actual name of the project instead of the code so I can properly format the source field of these records in the DRB database.

### DIFF
--- a/mappings/publisher_backlist.py
+++ b/mappings/publisher_backlist.py
@@ -22,12 +22,10 @@ class PublisherBacklistMapping(JSONMapping):
             'publisher_project_source': ('Publisher (from Projects)', '{0}')
         }
 
-    def apply_formatting(self):
+    def applyFormatting(self):
         self.record.has_part = []
-
         if self.record.source:
             source_list = self.record.source.split(' ')
-            print(source_list)
             self.record.source = source_list[0]
 
         if self.record.publisher_project_source:
--- a/mappings/publisher_backlist.py
+++ b/mappings/publisher_backlist.py
@@ -18,14 +18,15 @@ class PublisherBacklistMapping(JSONMapping):
             'rights': ('DRB Rights Classification', '{0}||||'),
             'contributors': [('Contributors', '{0}|||contributor')],
             'subjects': ('Subject 1', '{0}'),
-            'source': ('Projects', '{0}'),
+            'source': ('Project Name (from Projects)', '{0}'),
+            'source_id': ('DRB Record_ID', '{0}'),
             'publisher_project_source': ('Publisher (from Projects)', '{0}')
         }
 
     def applyFormatting(self):
         self.record.has_part = []
         if self.record.source:
-            source_list = self.record.source.split(' ')
+            source_list = self.record.source[0].split(' ')
             self.record.source = source_list[0]
 
         if self.record.publisher_project_source:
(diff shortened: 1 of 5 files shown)


mitri-slory:
I realize that the code displayed is actually the record id of where the value comes from instead of the value itself. This is what happens when a value is coming from a linked table in the Airtable API.

services/sources/publisher_backlist_service.py
```diff
@@ -96,4 +96,5 @@ def get_records_array(self,
             pub_backlist_records_response_json = pub_backlist_records_response.json()
             array_json.append(pub_backlist_records_response_json)
 
+        print(len(array_json))
```
kylevillegas93, line 99: nit: let's remove the array_json

mappings/publisher_backlist.py
kylevillegas93, line 1: I know some functions need to be in the old case style, but let's make sure we are consistent with the snake case where we can.

mappings/publisher_backlist.py
```diff
@@ -0,0 +1,103 @@
...
+                if len(self.record.identifiers) > 1 and 'oclc' in self.record.identifiers[1]:
+                    newISBNList.append(f'{self.record.identifiers[1]}')
+                    return newISBNList
+                else:
+                    return newISBNList
+                
+        return self.record.identifiers
+    
+    def format_subjects(self):
+        subjectList = []
```
kylevillegas93, line 80: e.g. - this can be subject_list

mappings/publisher_backlist.py
```diff
@@ -0,0 +1,103 @@
...
+                self.record.identifiers = self.format_identifiers()
+
+        self.record.rights = self.format_rights()
+
+    def format_authors(self):
+        authorList = []
+
+        if ';' in self.record.authors:
+            authorList = self.record.authors.split('; ')
+            newAuthorList = [f'{author}|||true' for author in authorList]
```
kylevillegas93, line 58: Should there be 2 or 3 pipes. I thought I've seen other author lists have only 2 pipes

mappings/publisher_backlist.py
```diff
@@ -0,0 +1,103 @@
...
+            publisher_source = self.record.publisher_project_source[0]
+            self.record.publisher_project_source = publisher_source
+
+        if self.record.authors:
+            self.record.authors = self.format_authors()
+
+        if self.record.subjects:
+            self.record.subjects = self.format_subjects()
+
+        if self.record.identifiers:
```
kylevillegas93, line 41: @Apophenia @mitri-slory - from my understanding, I think just the Airtable record ID can be used as the source_id and we don't need the prefix. What are y'all's thoughts?

mitri-slory: It can be used for that I agree.

mappings/publisher_backlist.py
```diff
@@ -0,0 +1,97 @@
...
+        else:
+            author_list.append(f'{self.record.authors}|||true)')
+            return author_list
+        
+        
+    def format_identifiers(self):
+        if 'isbn' in self.record.identifiers[0]:
+            isbn_string = self.record.identifiers[0].split('|')[0]
+            if ';' in isbn_string:
+                isbnList = isbn_string.split('; ')
```
kylevillegas93, line 63: nit: isbn_list or isbns

tests/unit/test_pub_backlist_mapping.py
```diff
@@ -0,0 +1,51 @@
+import pytest
+
+from mappings.publisher_backlist import PublisherBacklistMapping
+
+class TestPublisherBacklistMapping:
+    @pytest.fixture
+    def testMapping(self):
```
kylevillegas93, line 7: test_mapping

mappings/publisher_backlist.py
```diff
@@ -0,0 +1,97 @@
...
+                
+        return self.record.identifiers
+    
+    def format_subjects(self):
+        subject_list = []
+
+        if '|' in self.record.subjects:
+            subject_list = self.record.subjects.split('|')
+            new_subject_list = [f'{subject}||' for subject in subject_list] 
+            return new_subject_list
```
kylevillegas93, line 79: can just do `return [f'{subject}||' for subject in subject_list]`

mappings/publisher_backlist.py
```diff
@@ -0,0 +1,97 @@
...
+            author_list.append(f'{self.record.authors}|||true)')
+            return author_list
+        
+        
+    def format_identifiers(self):
+        if 'isbn' in self.record.identifiers[0]:
+            isbn_string = self.record.identifiers[0].split('|')[0]
+            if ';' in isbn_string:
+                isbnList = isbn_string.split('; ')
+                new_isbn_list = [f'{isbn}|isbn' for isbn in isbnList]
```
kylevillegas93, line 64: formatted_isbns

mappings/publisher_backlist.py
mitri-slory: Appreciate the suggestion

kylevillegas93 (approved):

Apophenia (approved):

PR #462, opened 2024-11-27
Add drive download utility functions and test

Apophenia:
- Adds utility functions for use in moving drive files to s3


tests/integration/test_google_integration.py
```diff
@@ -0,0 +1,15 @@
+import os
+from load_env import load_env_file
+
+load_env_file('local-compose', file_string='config/local-compose.yaml')
+
+from processes.util.google_integration import get_drive_file
+
+def test_get_drive_file():
+    test_id = os.environ['EXAMPLE_FILE_ID']
```
Apophenia, line 9: I'm using a test file ID in my local-compose. Could we upload a small file to test with, assuming the UUID of a drive file is otherwise anonymous (e.g. doesn't expose anything else about our account info or structure)?

processes/util/google_integration.py
```diff
@@ -0,0 +1,44 @@
...
+        file = None
+    except Exception as err:
+        logger.warning(f"Unexpected {type(err)=} occurred when downloading drive file {file_id}: {error}")
+
+    return file
+
+
+
+
+
```
kylevillegas93, line 43: nit: only need one new line at the end of the file!

processes/util/google_integration.py
```diff
@@ -0,0 +1,44 @@
...
+        done = False
+        while done is False:
+            status, done = downloader.next_chunk()
+
+    except HttpError as error:
+        logger.warning(f"HTTP error occurred when downloading Drive file {file_id}: {error}")
+        file = None
+    except Exception as err:
+        logger.warning(f"Unexpected {type(err)=} occurred when downloading drive file {file_id}: {error}")
```
kylevillegas93, line 36: logger.exception works well in this case to capture the error type and message

processes/util/google_integration.py
```diff
@@ -0,0 +1,44 @@
...
+logger = create_log(__name__)
+
+def create_drive_service(service_account_info):
+    scopes = ['https://www.googleapis.com/auth/drive']
+    credentials = Credentials.from_service_account_info(service_account_info, scopes=scopes)
+
+    return build('drive', 'v3', credentials=credentials)
+
+def get_drive_file(file_id: str) -> BytesIO:
+    drive_service = create_drive_service(service_account_info=json.loads(SERVICE_ACCOUNT_FILE))
```
kylevillegas93, line 21: I like the use of functions rather than a class here. Could we create the drive_service at the top level of this file so it's used more like a singleton?

processes/util/google_integration.py
```diff
@@ -0,0 +1,44 @@
...
+    try:
+        downloader = MediaIoBaseDownload(file, request)
+
+        done = False
+        while done is False:
+            status, done = downloader.next_chunk()
+
+    except HttpError as error:
+        logger.warning(f"HTTP error occurred when downloading Drive file {file_id}: {error}")
+        file = None
```
kylevillegas93, line 34: we can return early here and just have `return None`

processes/util/google_integration.py
```diff
@@ -0,0 +1,44 @@
...
+def get_drive_file(file_id: str) -> BytesIO:
+    drive_service = create_drive_service(service_account_info=json.loads(SERVICE_ACCOUNT_FILE))
+    request = drive_service.files().get_media(fileId=file_id)
+    file = BytesIO()
+
+    try:
+        downloader = MediaIoBaseDownload(file, request)
+
+        done = False
+        while done is False:
```
kylevillegas93, line 29: `while not done` I think may read better here!

processes/util/google_integration.py
kylevillegas93, line 1: I think putting this in the services folder instead is more explicit - e.g. `services/google_drive_service.py`

kylevillegas93 (changes requested):
Overall, looks good - nice work! just a few minor comments.

processes/util/google_integration.py
Apophenia: yeah, I think it's probably better this to refactor this into a service structure and move it there

processes/util/google_integration.py
Apophenia: oops, thanks, vs code was no longer linting on save!

services/google_drive_service.py
```diff
@@ -0,0 +1,38 @@
...
+
+from logger import create_log
+
+logger = create_log(__name__)
+
+service_account_info=json.loads(SERVICE_ACCOUNT_FILE)
+scopes = ['https://www.googleapis.com/auth/drive']
+credentials = Credentials.from_service_account_info(service_account_info, scopes=scopes)
+
+drive_service = build('drive', 'v3', credentials=credentials)
```
Apophenia, line 19: I left this as functions (instead of encapsulating it in a class) based on your comments -- this breaks from the other files in /services but I can see an argument for doing it either way. Let me know if this is what you were envisioning.

tests/integration/test_google_integration.py
Apophenia: This is hard-coded now, I added a test folder & a small PDF.

services/google_drive_service.py
```diff
@@ -0,0 +1,38 @@
...
+
+def get_drive_file(file_id: str) -> BytesIO:
```
kylevillegas93, line 21: nit: `Optional[BytesIO]`

tests/integration/test_google_integration.py
```diff
@@ -0,0 +1,14 @@
+import os
+from load_env import load_env_file
+
+load_env_file('local-compose', file_string='config/local-compose.yaml')
+
+from services.google_drive_service import get_drive_file
+
+TEST_ID = '1GBskazIv6j2BjOolNYqJKKIJC8iYhTcs'
```
kylevillegas93, line 8: TEST_FILE_ID is more explicit

services/google_drive_service.py
```diff
@@ -0,0 +1,38 @@
...
+from googleapiclient.errors import HttpError
+from google.oauth2.service_account import Credentials
+
+SERVICE_ACCOUNT_FILE = os.environ['GOOGLE_SERVICE_ACCOUNT_FILE']
+
+from logger import create_log
+
+logger = create_log(__name__)
+
+service_account_info=json.loads(SERVICE_ACCOUNT_FILE)
```
kylevillegas93, line 15: nit: spaces around equal sign

services/google_drive_service.py
kylevillegas93: Looks good!

services/google_drive_service.py
```diff
@@ -0,0 +1,38 @@
...
+
+        done = False
+        while not done:
+            status, done = downloader.next_chunk()
+
+    except HttpError as error:
+        logger.warning(f"HTTP error occurred when downloading Drive file {file_id}: {error}")
+        return None
+    except Exception as err:
+        logger.exception(f"Error occurred when downloading Drive file {file_id}: {error}")
```
kylevillegas93, line 36: logger.exception already logs the error message so there's no need to include the error in the message string

kylevillegas93: should we return None here?

kylevillegas93 (changes requested):
A few more small changes!

services/google_drive_service.py
Apophenia: Done!

Apophenia:
I think all of the suggestions should now be addressed, including the variable name change.

PR #464, opened 2024-12-03
SFR-2288 Querying by OCLC Number uses metadata API MARCXML endpoint

Apophenia:
- Removes search v1 endpoint and instead uses metadata v2
- Requests key with reduced scope (this is now required, since key no longer has access to full read/write API actions)
- Strips OCLC prefixes from OCLC control numbers when applying mapping: there are a few thousand records in QA that are prefixed, from various sources. However, the great majority of OCLC numbers are not prefixed and I created a function to remove the prefixes, as we do for e.g. Project MUSE

Tested with the SeedLocalData process and fields were manually compared -- after removing OCLC control number prefixes, they now look comparable at a glance.

### DIFF
--- a/managers/oclc_auth.py
+++ b/managers/oclc_auth.py
@@ -16,7 +16,7 @@ class OCLCAuthManager:
     _metadata_token = None
     _metadata_token_expires_at = None
     OCLC_SEARCH_AUTH_URL = 'https://oauth.oclc.org/token?scope=wcapi&grant_type=client_credentials'
-    OCLC_METADATA_AUTH_URL = 'https://oauth.oclc.org/token?scope=WorldCatMetadataAPI&grant_type=client_credentials'
+    OCLC_METADATA_AUTH_URL = 'https://oauth.oclc.org/token?scope=WorldCatMetadataAPI:view_marc_bib&grant_type=client_credentials'
 
     TIME_TO_REFRESH_IN_SECONDS = 60
 
--- a/mappings/oclcCatalog.py
+++ b/mappings/oclcCatalog.py
@@ -20,6 +20,9 @@ class CatalogMapping(XMLMapping):
         super(CatalogMapping, self).__init__(source, namespace, constants)
         self.mapping = self.createMapping()
 
+    def remove_oclc_prefixes(self, oclc_id):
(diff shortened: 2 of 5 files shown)


managers/oclc_auth.py
```diff
@@ -16,7 +16,7 @@ class OCLCAuthManager:
     _metadata_token = None
     _metadata_token_expires_at = None
     OCLC_SEARCH_AUTH_URL = 'https://oauth.oclc.org/token?scope=wcapi&grant_type=client_credentials'
-    OCLC_METADATA_AUTH_URL = 'https://oauth.oclc.org/token?scope=WorldCatMetadataAPI&grant_type=client_credentials'
+    OCLC_METADATA_AUTH_URL = 'https://oauth.oclc.org/token?scope=WorldCatMetadataAPI:view_marc_bib&grant_type=client_credentials'
```
kylevillegas93, line 19: Nice!

mappings/oclcCatalog.py
```diff
@@ -20,6 +20,9 @@ def __init__(self, source, namespace, constants):
         super(CatalogMapping, self).__init__(source, namespace, constants)
         self.mapping = self.createMapping()
 
+    def remove_oclc_prefixes(self, oclc_id):
+        return oclc_id.removeprefix('(OCoLC)').removeprefix('on').removeprefix('ocn').removeprefix('ocm')
```
kylevillegas93, line 24: I recommend pulling up prefixes into a set or list above as a constant, e.g. OCLC_PREFIXES = ['(OCoLC)', 'on', 'ocn', 'ocm']. 

That way we can just do: 

```
for prefix in OCLC_PREFIXES:
  oclc_id = oclc_id.removeprefix(prefix)
```

In theory ha

kylevillegas93 (changes requested):
Nice! Overall looks great - just 1 minor function improvement

mappings/oclcCatalog.py
Apophenia: Updated!

kylevillegas93 (commented):
Looks great!

kylevillegas93 (approved):

PR #466, opened 2024-12-05
NO-REF: Delete UofM & UofSC Processes

mitri-slory:
This hotfix PR deletes the UofM and UofSC processes, mappings, and tests due to the Publisher Backlist process encompassing both in its implementation.


kylevillegas93 (approved):
😃

PR #467, opened 2024-12-06
SFR-2102 Update declarative base import

Apophenia:
It looks like the SQLAlchemy mappings have already been moved to the "new style" 2.x declarative_base() but is using it under the old/deprecated import structure:

"The sqlalchemy.ext.declarative package is mostly, with some exceptions, moved to the sqlalchemy.orm package. The [declarative_base()](https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.declarative_base) and [declared_attr()](https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.declared_attr) functions are present without any behavioral changes."

From https://docs.sqlalchemy.org/en/20/changelog/migration_20.html

Did not try rebuilding the tables from scratch since the documentation states that this is just moving the function laterally to another package; I did run all of the integration and unit tests which are passing. This should fix the warning that occurs when tests are run.


kylevillegas93 (approved):
Nice - this looks good to me! Hopefully that'll get rid of the warnings !

PR #468, opened 2024-12-06
SFR-2386 Remove empty related bib warning log

kylevillegas93:
## Description
- This change removes the empty related bib warning log which is quite noisy and also normal in the majority of cases


Apophenia:
We could reintroduce something at the debug level in the future if it makes sense, but I agree this doesn't merit a warning

Apophenia (approved):

PR #470, opened 2024-12-06
SFR-2388: Adding ingest process functional tests

kylevillegas93:
## Description
- Adds ingest process functional tests

## Testing
pytest tests/functional

* Note that some tests are not passing locally - will follow up on those.


mitri-slory (approved):
This looks great. Once the local tests that are failing are fixed then I would say this looks good to go.

PR #471, opened 2024-12-09
SFR-2384: Fulfill Manifest Pub Backlist S3 Bucket

mitri-slory:
With this PR I modified the `store_pdf_manifest` method and the prefix attribute of the `FulfillManifestProcess` object to incorporate a new S3 folder called pubBacklist that will include all the manifests from the publisher backlist project such as UofM and UofSC. [https://us-east-1.console.aws.amazon.com/s3/buckets/drb-files-qa?prefix=manifests/&region=us-east-1&bucketType=general](url)

### DIFF
--- a/processes/file/fulfill_url_manifest.py
+++ b/processes/file/fulfill_url_manifest.py
@@ -16,15 +16,15 @@ class FulfillURLManifestProcess(CoreProcess):
     def __init__(self, *args):
         super(FulfillURLManifestProcess, self).__init__(*args[:4])
 
-        self.fullImport = self.process == 'complete' 
+        self.full_import = self.process == 'complete' 
         self.start_timestamp = None
 
         self.generateEngine()
         self.createSession()
 
-        self.s3Bucket = os.environ['FILE_BUCKET']
+        self.s3_bucket = os.environ['FILE_BUCKET']
         self.host = os.environ['DRB_API_HOST']
-        self.prefix = 'manifests/UofM/'
+        self.prefix = 'manifests/pubBacklist/'
         self.s3_manager = S3Manager()
         self.s3_manager.createS3Client()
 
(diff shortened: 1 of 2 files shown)


processes/file/fulfill_url_manifest.py
```diff
@@ -16,15 +16,15 @@ class FulfillURLManifestProcess(CoreProcess):
...
         self.start_timestamp = None
 
         self.generateEngine()
         self.createSession()
 
-        self.s3Bucket = os.environ['FILE_BUCKET']
+        self.s3_bucket = os.environ['FILE_BUCKET']
         self.host = os.environ['DRB_API_HOST']
-        self.prefix = 'manifests/UofM/'
+        self.prefix = 'manifests/pubBacklist/'
```
kylevillegas93, line 27: I recommend actually doing something like publisher_backlist to be explicit with the prefix. @Apophenia what are your thoughts?

processes/file/fulfill_url_manifest.py
```diff
@@ -73,10 +73,8 @@ def update_metadata_object(self, metadata_object, bucket_name, curr_key):
             metadata_json, counter = self.reading_order_fulfill(metadata_json, counter)
             metadata_json, counter = self.resource_fulfill(metadata_json, counter)
             metadata_json, counter = self.toc_fulfill(metadata_json, counter)
-        except (Exception, IndexError) as e:
-            logger.error(e)
-        except:
-            logger.error('One of the Link fulfill methods failed')  
+        except Exception as e:
```
kylevillegas93, line 76: Nice!

services/sources/publisher_backlist_service.py
```diff
@@ -133,9 +133,9 @@ def store_pdf_manifest(self, record):
 
             if media_type == 'application/pdf':
                 record_id = record.identifiers[0].split('|')[0]
-                manifest_path = 'manifests/{}/{}.json'.format(source, record_id)
+                manifest_path = 'manifests/pubBacklist/{}/{}.json'.format(source, record_id)
```
kylevillegas93, line 136: Could you use self.prefix here so that you don't have to update this prefix in two places?

mitri-slory: Sure I can do that

kylevillegas93 (approved):

processes/file/fulfill_url_manifest.py
Apophenia: Yes, I think this looks great!

services/sources/publisher_backlist_service.py
```diff
@@ -133,9 +134,9 @@ def store_pdf_manifest(self, record):
 
             if media_type == 'application/pdf':
                 record_id = record.identifiers[0].split('|')[0]
-                manifest_path = 'manifests/{}/{}.json'.format(source, record_id)
+                manifest_path = f'{self.prefix}/{source}/{record_id}.json'
```
Apophenia, line 137: I think this looks good!

PR #473, opened 2024-12-11
SFR-2391: Delete Publisher Backlist Record Manifests

mitri-slory:
This PR focuses on deleting Publisher Backlist record manifests and digital asset files from Google Drive. Currently there is no code for the delete_digital_asset_file method since I'm not sure where the digital assets are located on Google Drive and how to delete them with the API. Maybe @Apophenia could assist me on that?

### DIFF
--- a/services/sources/publisher_backlist_service.py
+++ b/services/sources/publisher_backlist_service.py
@@ -4,6 +4,7 @@ import requests
 import json
 import urllib.parse
 from typing import Optional
+from model import Record
 
 from logger import create_log
 from mappings.publisher_backlist import PublisherBacklistMapping
@@ -23,6 +24,48 @@ class PublisherBacklistService(SourceService):
         
         self.airtable_auth_token = os.environ.get('AIRTABLE_KEY', None)
 
+    def delete_records(
+        self,
+        offset: Optional[int]=None,
+        limit: Optional[int]=None
+    ):
+        filter_by_formula = self.build_filter_by_formula_parameter(deleted=True, full_import=False, start_timestamp=None)
+        
+        array_json_records = self.get_records_array(limit, filter_by_formula)
+
+        for json_dict in array_json_records:
+            if json_dict['records']:
+                for records_value in json_dict['records']:
+                    try:
+                        record_metadata_dict = records_value['fields']
+                        pub_backlist_record = PublisherBacklistMapping(record_metadata_dict)
+                        pub_backlist_record.applyMapping()
+                        self.delete_manifest(record_metadata_dict, pub_backlist_record.record)
+                        self.delete_digital_asset_file()
+                    except Exception:
+                        logger.exception(f'Record has not been created {records_value}')
+        
+    def delete_manifest(self, record_metadata_dict, mapped_record):
+        try:
+            for record in self.session.query(Record) \
+                .filter(Record.source_id == mapped_record.source_id):  
+                    key_format = f"{self.prefix}{record.source}"
+
+                    if record_metadata_dict['File ID 1']:
+                        file_title = record_metadata_dict['File ID 1']
+                        self.s3_manager.s3Client.delete_object(Bucket= self.s3_bucket, Key= f'{key_format}{file_title}.json')
+                    elif record_metadata_dict['File ID 2']:
+                        file_title = record_metadata_dict['File ID 2']
+                        self.s3_manager.s3Client.delete_object(Bucket= self.s3_bucket, Key= f'{key_format}{file_title}.json')
+                    elif record_metadata_dict['Hathi ID']:
+                        file_title = record_metadata_dict['Hathi ID']
+                        self.s3_manager.s3Client.delete_object(Bucket= self.s3_bucket, Key= f'{key_format}{file_title}.json')
+        except Exception:
+            logger.exception('Key does not exist')
+            
+    def delete_digital_asset_file(self):
+        pass
+
     def get_records(
         self,
         full_import: bool=False, 
@@ -38,7 +81,7 @@ class PublisherBacklistService(SourceService):
                     record_metadata_dict = records_value['fields']
                     pub_backlist_record = PublisherBacklistMapping(record_metadata_dict)
                     pub_backlist_record.applyMapping()
-                    self.add_has_part_mapping(pub_backlist_record, pub_backlist_record.record)
+                    self.add_has_part_mapping(pub_backlist_record.record)
                     self.store_pdf_manifest(pub_backlist_record.record)
                     complete_records.append(pub_backlist_record)
                 except Exception:
@@ -54,13 +97,20 @@ class PublisherBacklistService(SourceService):
         if offset == None:
             limit = 100
         
-        filter_by_formula = self.build_filter_by_formula_parameter(full_import, start_timestamp)
+        filter_by_formula = self.build_filter_by_formula_parameter(deleted=False, full_import=None, start_timestamp=None)
                 
         array_json_records = self.get_records_array(limit, filter_by_formula)
         
         return array_json_records
         
-    def build_filter_by_formula_parameter(self, full_import: bool=False, start_timestamp: datetime=None) -> str:
+    def build_filter_by_formula_parameter(self, deleted=None, full_import: bool=False, start_timestamp: datetime=None) -> str:
+        if deleted == True:
+            deleted_filter = f"IF(%7BDRB_Deleted%7D%20%3D%20TRUE(),%20TRUE(),%20FALSE())"
+            filter_by_formula = f"&filterByFormula={deleted_filter}"
+            return filter_by_formula
+        else:
+            deleted_filter = f"IF(%7BDRB_Deleted%7D%20!%3D%20TRUE(),%20TRUE(),%20FALSE())"
+
         if not start_timestamp:
             start_timestamp = datetime.now(timezone.utc).replace(tzinfo=None) - timedelta(hours=24)
 
@@ -75,7 +125,7 @@ class PublisherBacklistService(SourceService):
             is_same_date_time_filter = f"IS_SAME(%7BLast%20Modified%7D,%20%22{start_date_time_encoded}%22"
             is_after_date_time_filter = f"%20IS_AFTER(%7BLast%20Modified%7D,%20%22{start_date_time_encoded}%22"
 
-            filter_by_formula = f"&filterByFormula=AND(OR({is_same_date_time_filter}),{is_after_date_time_filter})),{if_ready_to_ingest_is_true_filter})"
+            filter_by_formula = f"&filterByFormula=AND(OR({is_same_date_time_filter}),{is_after_date_time_filter})),AND({if_ready_to_ingest_is_true_filter},{deleted_filter}))"
 
             return filter_by_formula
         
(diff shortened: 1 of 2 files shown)


services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,48 @@ def __init__(self):
...
+                        file_title = record_metadata_dict['File ID 1']
+                        self.s3_manager.s3Client.delete_object(Bucket= self.s3_bucket, Key= f'{key_format}{file_title}.json')
+                    elif record_metadata_dict['File ID 2']:
+                        file_title = record_metadata_dict['File ID 2']
+                        self.s3_manager.s3Client.delete_object(Bucket= self.s3_bucket, Key= f'{key_format}{file_title}.json')
+                    elif record_metadata_dict['Hathi ID']:
+                        file_title = record_metadata_dict['Hathi ID']
+                        self.s3_manager.s3Client.delete_object(Bucket= self.s3_bucket, Key= f'{key_format}{file_title}.json')
+        except Exception:
+            logger.exception('Key does not exist')
```
kylevillegas93, line 64: I think since we are catching all Exceptions here we may want to change this error message to something more generic, e.g. Failed to delete manifest for record: {record}

services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,48 @@ def __init__(self):
...
+            
+    def delete_digital_asset_file(self):
```
kylevillegas93, line 66: I think we can leave the digital asset file in Google Drive alone for now.

services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,48 @@ def __init__(self):
...
+        filter_by_formula = self.build_filter_by_formula_parameter(deleted=True, full_import=False, start_timestamp=None)
+        
+        array_json_records = self.get_records_array(limit, filter_by_formula)
+
+        for json_dict in array_json_records:
+            if json_dict['records']:
+                for records_value in json_dict['records']:
+                    try:
+                        record_metadata_dict = records_value['fields']
+                        pub_backlist_record = PublisherBacklistMapping(record_metadata_dict)
```
kylevillegas93, line 41: do we need to create the publisher back list mapping?

services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,48 @@ def __init__(self):
...
+                        pub_backlist_record.applyMapping()
+                        self.delete_manifest(record_metadata_dict, pub_backlist_record.record)
```
kylevillegas93, line 43: Could we just use the source id?

services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,48 @@ def __init__(self):
...
+                        self.delete_digital_asset_file()
+                    except Exception:
+                        logger.exception(f'Record has not been created {records_value}')
+        
+    def delete_manifest(self, record_metadata_dict, mapped_record):
+        try:
+            for record in self.session.query(Record) \
```
kylevillegas93, line 50: There should be only one record here so we can do .first() and not loop through

services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,48 @@ def __init__(self):
...
+                .filter(Record.source_id == mapped_record.source_id):  
+                    key_format = f"{self.prefix}{record.source}"
+
+                    if record_metadata_dict['File ID 1']:
+                        file_title = record_metadata_dict['File ID 1']
```
kylevillegas93, line 55: Hmm is there an easier way to do get the manifest file name? Is this anywhere in the record?

services/sources/publisher_backlist_service.py
```diff
@@ -54,13 +97,20 @@ def get_records_json(self,
...
-        filter_by_formula = self.build_filter_by_formula_parameter(full_import, start_timestamp)
+        filter_by_formula = self.build_filter_by_formula_parameter(deleted=False, full_import=None, start_timestamp=None)
                 
         array_json_records = self.get_records_array(limit, filter_by_formula)
         
         return array_json_records
         
-    def build_filter_by_formula_parameter(self, full_import: bool=False, start_timestamp: datetime=None) -> str:
+    def build_filter_by_formula_parameter(self, deleted=None, full_import: bool=False, start_timestamp: datetime=None) -> str:
+        if deleted == True:
```
kylevillegas93, line 107: we can simplify here to if deleted

services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,48 @@ def __init__(self):
         
         self.airtable_auth_token = os.environ.get('AIRTABLE_KEY', None)
 
+    def delete_records(
+        self,
+        offset: Optional[int]=None,
+        limit: Optional[int]=None
+    ):
+        filter_by_formula = self.build_filter_by_formula_parameter(deleted=True, full_import=False, start_timestamp=None)
```
kylevillegas93, line 32: since there are default params - you could just do build_filter_by_formula_pamaeter(deleted=True)

services/sources/publisher_backlist_service.py
```diff
@@ -54,13 +97,20 @@ def get_records_json(self,
...
         array_json_records = self.get_records_array(limit, filter_by_formula)
         
         return array_json_records
         
-    def build_filter_by_formula_parameter(self, full_import: bool=False, start_timestamp: datetime=None) -> str:
+    def build_filter_by_formula_parameter(self, deleted=None, full_import: bool=False, start_timestamp: datetime=None) -> str:
+        if deleted == True:
+            deleted_filter = f"IF(%7BDRB_Deleted%7D%20%3D%20TRUE(),%20TRUE(),%20FALSE())"
+            filter_by_formula = f"&filterByFormula={deleted_filter}"
+            return filter_by_formula
```
kylevillegas93, line 110: since you are returning early here we don't need the else

services/sources/publisher_backlist_service.py
```diff
@@ -54,13 +97,20 @@ def get_records_json(self,
...
+        else:
+            deleted_filter = f"IF(%7BDRB_Deleted%7D%20!%3D%20TRUE(),%20TRUE(),%20FALSE())"
```
kylevillegas93, line 112: we could be explicit like - `is_not_deleted`

services/sources/publisher_backlist_service.py
kylevillegas93, line 110: You could just do `return "&filterByFormula=IF(%7BDRB_Deleted%7D%20%3D%20TRUE(),%20TRUE(),%20FALSE())"`

does this mean that it will only return records that are marked as deleted?

services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,48 @@ def __init__(self):
...
+        for json_dict in array_json_records:
+            if json_dict['records']:
+                for records_value in json_dict['records']:
+                    try:
+                        record_metadata_dict = records_value['fields']
+                        pub_backlist_record = PublisherBacklistMapping(record_metadata_dict)
+                        pub_backlist_record.applyMapping()
+                        self.delete_manifest(record_metadata_dict, pub_backlist_record.record)
+                        self.delete_digital_asset_file()
+                    except Exception:
```
kylevillegas93, line 45: Do we know for sure that the record has not been created if an exception has been thrown? could there be other exceptions that may happen?

services/sources/publisher_backlist_service.py
Apophenia: I wasn't planning on writing any functionality to handle Drive deletion this sprint - I think we could handle this in future sprints. I think we could consider that part of a large project of deciding how we want to manage the Drive assets that have already been ingested.

services/sources/publisher_backlist_service.py
mitri-slory: That's a good suggestion

services/sources/publisher_backlist_service.py
mitri-slory: I initially used the mapping to get access to the `source_id` directly but then realized that the record_metadata_dict has the Airtable field `DRB Record_ID` which is used as the `source_id` in the DRB record table. Therefore, we don't need to create the mapping.

services/sources/publisher_backlist_service.py
mitri-slory: We can just use the DRB_Record ID from the record_metadata_dict since that is the source id in the mapping.

services/sources/publisher_backlist_service.py
mitri-slory: The only alternative I could think of is using the record.has_part attribute, splitting the list element that contains the link to the webpub manifest (application/webpub+json type) and splicing the link string to retrieve the file id from that link.

services/sources/publisher_backlist_service.py
mitri-slory: I like the clarity of specifying the type and purpose of the filter that I'm adding to the formula. Also, it will only return records that are marked as deleted with a check mark which is equivalent to "true" in the JSON response from the Airtable API.

services/sources/publisher_backlist_service.py
mitri-slory: With the mapping creation being deleted, I think a better exception would be stating that 'Delete manifest method has failed for {records_value}'

services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,41 @@ def __init__(self):
         
         self.airtable_auth_token = os.environ.get('AIRTABLE_KEY', None)
 
+    def delete_records(
+        self,
+        offset: Optional[int]=None,
+        limit: Optional[int]=None
```
kylevillegas93, line 30: do we need the offset here?

mitri-slory: Yeah we don't need it

services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,40 @@ def __init__(self):
...
+    ):
+        filter_by_formula = self.build_filter_by_formula_parameter(deleted=True)
+        
+        array_json_records = self.get_records_array(limit, filter_by_formula)
+
+        for json_dict in array_json_records:
+            if json_dict['records']:
+                for records_value in json_dict['records']:
+                    try:
```
kylevillegas93, line 38: I think because we have the try/except login below in delete_manifeset - we can remove the try/except block here. You can add some extra guards against trying to access dict fields that don't exist

services/sources/publisher_backlist_service.py
```diff
@@ -23,6 +24,40 @@ def __init__(self):
...
+                        record_metadata_dict = records_value['fields']
+                        self.delete_manifest(record_metadata_dict)
+                    except Exception:
+                        logger.exception(f'Delete manifest method has failed for {records_value}')
+        
+    def delete_manifest(self, record_metadata_dict):
+        try:
+            record = self.session.query(Record).filter(Record.source_id == record_metadata_dict['DRB Record_ID']).first()
```
kylevillegas93, line 46: Can record be None here? we should handle that case

services/sources/publisher_backlist_service.py
mitri-slory: That makes sense and I agree with that suggestion.

services/sources/publisher_backlist_service.py
mitri-slory: Agreed

kylevillegas93 (approved):
Looks good to me - nice work!

PR #474, opened 2024-12-11
SFR-2397: Deploy to the new terraform stacks in production

kylevillegas93:
## Description
- Deploys to the new terraform stacks in production


Apophenia (approved):
Looks good, thanks for putting this together!

PR #477, opened 2024-12-13
SFR-2392_Delete Publisher Backlist Works

mitri-slory:
This PR focuses on deleting Publisher Backlist works from the PSQL and ES databases when the Airtable Deleted field's value is set to the boolean true.

### DIFF
--- a/services/sources/publisher_backlist_service.py
+++ b/services/sources/publisher_backlist_service.py
@@ -131,6 +131,8 @@ class PublisherBacklistService(SourceService):
     ) -> list[dict]:
         if offset == None:
             limit = 100
+            
+        limit = offset
         
         filter_by_formula = self.build_filter_by_formula_parameter(deleted=False, full_import=None, start_timestamp=None)
                 
--- a/services/sources/publisher_backlist_service.py
+++ b/services/sources/publisher_backlist_service.py
@@ -24,6 +24,16 @@ class PublisherBacklistService(SourceService):
         self.s3_manager.createS3Client()
         self.s3_bucket = os.environ['FILE_BUCKET']
         self.prefix = 'manifests/publisher_backlist'
+        self.db_manager = DBManager(
+            user= os.environ.get('POSTGRES_USER', None),
+            pswd= os.environ.get('POSTGRES_PSWD', None),
+            host= os.environ.get('POSTGRES_HOST', None),
+            port= os.environ.get('POSTGRES_PORT', None),
+            db= os.environ.get('POSTGRES_NAME', None)
+        )
+        self.db_manager.generateEngine()
+        self.es_manager = ElasticsearchManager()
+        self.es_manager.createElasticConnection()
         
         self.airtable_auth_token = os.environ.get('AIRTABLE_KEY', None)
 
@@ -35,30 +45,16 @@ class PublisherBacklistService(SourceService):
         
         array_json_records = self.get_records_array(limit, filter_by_formula)
 
-        dbManager = DBManager(
-        user= os.environ.get('POSTGRES_USER', None),
-        pswd= os.environ.get('POSTGRES_PSWD', None),
-        host= os.environ.get('POSTGRES_HOST', None),
-        port= os.environ.get('POSTGRES_PORT', None),
-        db= os.environ.get('POSTGRES_NAME', None)
-    )
-
-        dbManager.generateEngine()
-
-        dbManager.createSession()
-
-        esManager = ElasticsearchManager()
-        esManager.createElasticConnection()
-
         for json_dict in array_json_records:
             if json_dict['records']:
                 for records_value in json_dict['records']:
                     if records_value['fields']:
                         record_metadata_dict = records_value['fields']
-                        self.delete_manifest(record_metadata_dict)
-                        self.delete_work(dbManager, record_metadata_dict)
+                        self.delete_manifest(self.db_manager, record_metadata_dict)
+                        self.delete_work(record_metadata_dict)
         
     def delete_manifest(self, record_metadata_dict):
+        self.db_manager.createSession()
         try:
             record = self.session.query(Record).filter(Record.source_id == record_metadata_dict['DRB Record_ID']).first()
             if record:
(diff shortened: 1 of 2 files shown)


services/sources/publisher_backlist_service.py
```diff
@@ -4,25 +4,103 @@
...
+        
+        array_json_records = self.get_records_array(limit, filter_by_formula)
+
+        dbManager = DBManager(
+        user= os.environ.get('POSTGRES_USER', None),
+        pswd= os.environ.get('POSTGRES_PSWD', None),
+        host= os.environ.get('POSTGRES_HOST', None),
+        port= os.environ.get('POSTGRES_PORT', None),
+        db= os.environ.get('POSTGRES_NAME', None)
+    )
```
kylevillegas93, line 44: let's fix the alignment here!

services/sources/publisher_backlist_service.py
```diff
@@ -4,25 +4,103 @@
...
 
+    def delete_records(
+        self,
+        limit: Optional[int]=None
+    ):
+        filter_by_formula = self.build_filter_by_formula_parameter(deleted=True)
+        
+        array_json_records = self.get_records_array(limit, filter_by_formula)
+
+        dbManager = DBManager(
```
kylevillegas93, line 38: db_manager

services/sources/publisher_backlist_service.py
```diff
@@ -4,25 +4,103 @@
...
+        host= os.environ.get('POSTGRES_HOST', None),
+        port= os.environ.get('POSTGRES_PORT', None),
+        db= os.environ.get('POSTGRES_NAME', None)
+    )
+
+        dbManager.generateEngine()
+
+        dbManager.createSession()
+
+        esManager = ElasticsearchManager()
```
kylevillegas93, line 50: es_manager

services/sources/publisher_backlist_service.py
kylevillegas93: you can pull the db_manager and es_manager up into the constructor

services/sources/publisher_backlist_service.py
```diff
@@ -4,25 +4,103 @@
...
+                work = dbManager.session.query(Work).filter(Work.id == edition.work_id).first()
+                work_uuid_str = str(work.uuid)
+                es_work_resp = Search(index=os.environ['ELASTICSEARCH_INDEX']).query('match', uuid=work_uuid_str)
+                es_work_resp.delete()
+                work.delete()
+        except Exception:
+            logger.exception('Failed to delete work')
+
+        dbManager.session.commit()
+        dbManager.session.close()
```
kylevillegas93, line 86: you can wrap the db_manager.session.close() in a finally block

services/sources/publisher_backlist_service.py
```diff
@@ -4,25 +4,103 @@
...
+                edition =  dbManager.session.query(Edition).filter(Edition.dcdw_uuids.contains([record_uuid_str])).first()
+                work = dbManager.session.query(Work).filter(Work.id == edition.work_id).first()
+                work_uuid_str = str(work.uuid)
+                es_work_resp = Search(index=os.environ['ELASTICSEARCH_INDEX']).query('match', uuid=work_uuid_str)
+                es_work_resp.delete()
+                work.delete()
+        except Exception:
+            logger.exception('Failed to delete work')
+
+        dbManager.session.commit()
```
kylevillegas93, line 85: I would move the commit call up to the try block in case we error there

services/sources/publisher_backlist_service.py
```diff
@@ -4,25 +4,103 @@
...
+            if record:
+                record_uuid_str = str(record.uuid)
+                edition =  dbManager.session.query(Edition).filter(Edition.dcdw_uuids.contains([record_uuid_str])).first()
+                work = dbManager.session.query(Work).filter(Work.id == edition.work_id).first()
+                work_uuid_str = str(work.uuid)
+                es_work_resp = Search(index=os.environ['ELASTICSEARCH_INDEX']).query('match', uuid=work_uuid_str)
+                es_work_resp.delete()
+                work.delete()
+        except Exception:
+            logger.exception('Failed to delete work')
```
kylevillegas93, line 83: Let's add some more detail to this exception message

services/sources/publisher_backlist_service.py
```diff
@@ -4,25 +4,103 @@
...
+            logger.exception(f'Failed to delete manifest for record: {record.source_id}')
+
+    def delete_work(self, dbManager, record_metadata_dict):
+        record = dbManager.session.query(Record).filter(Record.source_id == record_metadata_dict['DRB Record_ID']).first()
+
+        try:
+            if record:
+                record_uuid_str = str(record.uuid)
+                edition =  dbManager.session.query(Edition).filter(Edition.dcdw_uuids.contains([record_uuid_str])).first()
+                work = dbManager.session.query(Work).filter(Work.id == edition.work_id).first()
```
kylevillegas93, line 77: Hmm - we should be able to get the work uuid from the edition right? If we can, we could reduce this to just 1 call on the db instead of 2!

kylevillegas93 (changes requested):
Looking good!

services/sources/publisher_backlist_service.py
mitri-slory: Nice catch

services/sources/publisher_backlist_service.py
mitri-slory: We can't currently get the work uuid from the edition since the uuids in the dcdw_uuids field are record uuids, but this can be a talking point to make the relation between works and records more clear in the database.

services/sources/publisher_backlist_service.py
```diff
@@ -4,25 +4,102 @@
...
         self.s3_manager.createS3Client()
         self.s3_bucket = os.environ['FILE_BUCKET']
         self.prefix = 'manifests/publisher_backlist'
+        self.db_manager = DBManager(
+            user= os.environ.get('POSTGRES_USER', None),
+            pswd= os.environ.get('POSTGRES_PSWD', None),
+            host= os.environ.get('POSTGRES_HOST', None),
+            port= os.environ.get('POSTGRES_PORT', None),
+            db= os.environ.get('POSTGRES_NAME', None)
+        )
```
kylevillegas93, line 33: ah last thing is that you can just do DBManager() and it'll init with those default values

PR #478, opened 2024-12-17
SFR-2406: Deploy only to the terraform stacks

kylevillegas93:
## Description
- Deploys only to the terraform stacks
- Stops deploying to the CF stacks


Apophenia (approved):
Looks good!

PR #479, opened 2024-12-17
SFR-2393: Delete Publisher Backlist Editions

mitri-slory:
This PR focuses on the automated deleted edge case for when a Publisher Backlist record produces a new edition with an already existing work that has other editions during the ingestion process. The PSQL deletion has been enacted but the deletion of the same edition in ES is commented out since the fields for editions are not as clear with how to search for it compared finding the same work in the PSQL and ES databases. That will be my focus with future commits.

### DIFF
--- a/services/sources/publisher_backlist_service.py
+++ b/services/sources/publisher_backlist_service.py
@@ -74,18 +74,28 @@ class PublisherBacklistService(SourceService):
                     self.db_manager.session.query(Work).filter(Work.id == edition.work_id).delete()
                     es_work_resp.delete()
                     self.db_manager.session.commit()
-
+                else:
+                    self.delete_pub_backlist_edition_only(record.uuid_str, work)
         except Exception:
             logger.exception('Work/Edition does not exist or failed to delete work: {work.id}')
         finally:
             self.db_manager.session.close()
 
     def checkAllEditionsRelatedToRecord(self, record_uuid_str, work):
-        for edition in self.db_manager.session.query(Edition).filter(Edition.work_id == work.id):
+        for edition in self.db_manager.session.query(Edition).filter(Edition.work_id == work.id).all():
             if record_uuid_str not in edition.dcdw_uuids:
                 return False
         return True
-
+            
+    def delete_pub_backlist_edition_only(self, record_uuid_str, work):
+        for edition in self.db_manager.session.query(Edition) \
+            .filter(Edition.work_id == work.id) \
+            .filter(Edition.dcdw_uuids.contains([record_uuid_str])) \
+            .all():
+                self.db_manager.session.delete(edition)
+                # es_edition_resp = Search(index=os.environ['ELASTICSEARCH_INDEX']).query('match', uuid=uuid)
+                # es_edition_esp.delete()
+            
     def get_metadata_file_name(self, record, record_metadata_dict):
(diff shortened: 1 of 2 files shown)


services/sources/publisher_backlist_service.py
```diff
@@ -74,18 +74,28 @@ def delete_work(self, record_metadata_dict):
...
                     self.db_manager.session.commit()
-
+                else:
+                    self.delete_pub_backlist_edition_only(record.uuid_str, work)
         except Exception:
             logger.exception('Work/Edition does not exist or failed to delete work: {work.id}')
         finally:
             self.db_manager.session.close()
 
     def checkAllEditionsRelatedToRecord(self, record_uuid_str, work):
```
kylevillegas93, line 84: oops missed this - let's do snake case for this method name

services/sources/publisher_backlist_service.py
```diff
@@ -74,18 +74,28 @@ def delete_work(self, record_metadata_dict):
                     self.db_manager.session.query(Work).filter(Work.id == edition.work_id).delete()
                     es_work_resp.delete()
                     self.db_manager.session.commit()
-
+                else:
```
kylevillegas93, line 77: The above if can be just `if self.checkAllEditionsRelatedToRecord:`

services/sources/publisher_backlist_service.py
```diff
@@ -74,18 +74,28 @@ def delete_work(self, record_metadata_dict):
...
             logger.exception('Work/Edition does not exist or failed to delete work: {work.id}')
         finally:
             self.db_manager.session.close()
 
     def checkAllEditionsRelatedToRecord(self, record_uuid_str, work):
-        for edition in self.db_manager.session.query(Edition).filter(Edition.work_id == work.id):
+        for edition in self.db_manager.session.query(Edition).filter(Edition.work_id == work.id).all():
             if record_uuid_str not in edition.dcdw_uuids:
                 return False
         return True
```
kylevillegas93, line 88: Based on the above query you're getting the first edition associated with the record uuid then grabbing the work. I think we can delete the work if that's the only edition. 

Would it be easier to just check if the length of editions is just equal to 1?

services/sources/publisher_backlist_service.py
```diff
@@ -74,18 +74,28 @@ def delete_work(self, record_metadata_dict):
...
-
+            
+    def delete_pub_backlist_edition_only(self, record_uuid_str, work):
+        for edition in self.db_manager.session.query(Edition) \
+            .filter(Edition.work_id == work.id) \
+            .filter(Edition.dcdw_uuids.contains([record_uuid_str])) \
+            .all():
+                self.db_manager.session.delete(edition)
+                # es_edition_resp = Search(index=os.environ['ELASTICSEARCH_INDEX']).query('match', uuid=uuid)
+                # es_edition_esp.delete()
```
kylevillegas93, line 97: Here you may need to get the work from elastic search and remove the edition based on the edition id from the nested editions list. Then we can update the work in elastic search.

services/sources/publisher_backlist_service.py
mitri-slory: You're right that checking the length of the editions would be easier

kylevillegas93 (approved):
Looks good - just 1 minor suggestion that's non-blocking!

PR #481, opened 2024-12-18
SFR-2413: Migrate edition endpoint to use plural noun

kylevillegas93:
## Description
- Migrates edition endpoint to use plural noun

## Testing
`docker compose up and hit the endpoint /editions`

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -9,7 +9,7 @@ from waitress import serve
 
 from logger import create_log
 from .blueprints import (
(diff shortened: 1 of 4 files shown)


mitri-slory (approved):
This looks good. I would like to ask what is the motivation for adding this alternative edition endpoint? Is it to allow users and patrons to use either endpoint based on how they interpret the endpoint?

PR #483, opened 2024-12-18
SFR-2412: Migrate collection endpoint to use plural noun

jackiequach:
# Description
- Migrates collection endpoint to use plural noun

# Testing
`docker compose up` and hit the endpoint /collections/list

### DIFF
--- a/api/blueprints/drbCollection.py
+++ b/api/blueprints/drbCollection.py
@@ -362,7 +362,7 @@ def collectionDeleteWorkEdition(uuid, user=None):
 
 
 @collection.route('/list', methods=['GET'])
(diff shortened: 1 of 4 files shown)


kylevillegas93 (approved):
nice!! looking good :)

api/blueprints/drbCollection.py
```diff
@@ -355,6 +362,7 @@ def collectionDeleteWorkEdition(uuid, user=None):
 
 
 @collection.route('/list', methods=['GET'])
+@collections.route('/list', methods=['GET'])
```
kylevillegas93, line 365: ohhh this actually should be just `''`

PR #484, opened 2024-12-19
SFR-2414: Migrate work endpoint to use plural noun

jackiequach:
# Description
- Migrates work endpoint to use plural noun
- Moves /collections/list GET method to /collections in swagger
# Testing
`docker compose up` and hit the endpoint /works


kylevillegas93 (approved):
looks good! nice job :)

PR #485, opened 2024-12-19
SFR-2415: Migrate link endpoint to use plural noun

jackiequach:
# Description
- Migrates link endpoint to use plural noun
# Testing
`docker compose up` and hit the endpoint /links


kylevillegas93 (approved):
woot woot!

PR #487, opened 2024-12-20
NOREF: Retrieve & use Airtable parameters directly

Apophenia:
Changing the SSM service to return the parameter value only - I don't think there's a lot of value in what comes back with the rest of the object and my instinct is that it should more closely resemble the API for grabbing the parameter as an environment variable.

### DIFF
--- a/services/sources/publisher_backlist_service.py
+++ b/services/sources/publisher_backlist_service.py
@@ -24,8 +24,7 @@ class PublisherBacklistService(SourceService):
         self.prefix = 'manifests/publisher_backlist'
 
         if os.environ['ENVIRONMENT'] == 'production':
-            # self.airtable_auth_token = get_parameter
-            pass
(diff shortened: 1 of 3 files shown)


services/sources/publisher_backlist_service.py
```diff
@@ -29,7 +30,10 @@ def __init__(self):
         self.es_manager = ElasticsearchManager()
         self.es_manager.createElasticConnection()
         
-        self.airtable_auth_token = os.environ.get('AIRTABLE_KEY', None)
+        if os.environ['ENVIRONMENT'] == 'production':
+            self.airtable_auth_token = get_parameter('arn:aws:ssm:us-east-1:946183545209:parameter/drb/production/airtable/pub-backlist/api-key')
+        else:
+            self.airtable_auth_token = get_parameter('arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/airtable/pub-backlist/api-key')
```
kylevillegas93, line 36: I recommend just passing in the path `/airtable/pub-backlist/api-key` and the get_parameter function can figure out the environment and add the prefix!

PR #488, opened 2024-12-20
NOREF: Drive Service Refactor

Apophenia:
- Retrieve parameters directly based on environment
- Lazily instantiate service with helper function if service does not already exist when called
- Add helper functions for Google-specific metadata parsing used by backlist processing functions


Apophenia:
I don't think I can get this to work how I want within the context of the unit tests without expanding the scope in a way that feels bad in Python so I'm just going to make it a class

PR #489, opened 2024-12-20
Refactor Drive service into a class

Apophenia:
Refactor drive service into a class & only return param from SSM service

### DIFF
--- a/services/google_drive_service.py
+++ b/services/google_drive_service.py
@@ -2,39 +2,57 @@ from io import BytesIO
 import json
 import os
 from typing import Optional
+from urllib.parse import urlparse, parse_qs
 from googleapiclient.discovery import build
 from googleapiclient.http import MediaIoBaseDownload
 from googleapiclient.errors import HttpError
 from google.oauth2.service_account import Credentials
+from services.ssm_service import get_parameter
 
-SERVICE_ACCOUNT_FILE = os.environ['GOOGLE_SERVICE_ACCOUNT_FILE']
 
 from logger import create_log
 
 logger = create_log(__name__)
 
-service_account_info = json.loads(SERVICE_ACCOUNT_FILE)
-scopes = ['https://www.googleapis.com/auth/drive']
-credentials = Credentials.from_service_account_info(service_account_info, scopes=scopes)
+class GoogleDriveService:
+    def __init__(self):
+        if os.environ['ENVIRONMENT'] == 'production':
+            SERVICE_ACCOUNT_FILE = get_parameter('arn:aws:ssm:us-east-1:946183545209:parameter/drb/production/google-drive-service-key')
+        else:
+            SERVICE_ACCOUNT_FILE = get_parameter('arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/google-drive-service-key')
+        self.service_account_info = json.loads(SERVICE_ACCOUNT_FILE)
+        self.scopes = ['https://www.googleapis.com/auth/drive',
+        'https://www.googleapis.com/auth/drive.file',
+        'https://www.googleapis.com/auth/drive.metadata']
+        self.credentials = Credentials.from_service_account_info(self.service_account_info, scopes=self.scopes)
+        self.drive_service = build('drive', 'v3', credentials=self.credentials)
 
-drive_service = build('drive', 'v3', credentials=credentials)
+    def get_drive_file(self, file_id: str) -> Optional[BytesIO]:
+        request = self.drive_service.files().get_media(fileId=file_id)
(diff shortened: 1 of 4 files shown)


tests/unit/test_opds_publication.py
```diff
@@ -265,7 +274,7 @@ def test_parseEditionToPublication_ReaderFlagTrue(self, testPubEls, mocker):
             mocker.call('rights', {'source': 'testSource', 'license': 'testLicense', 'rightsStatement': 'testStatement'})
         ])
 
-        pubMocks['addLink'].assert_called_with({'href': 'https://digital-research-books-beta.nypl.org/read/testID', 'type': 'testType', 'rel': 'http://opds-spec.org/acquisition/open-access', 'identifier': 'readable'})
+        pubMocks['addLink'].assert_called_with({'href': 'https://drb-qa.nypl.org/read/testID', 'type': 'testType', 'rel': 'http://opds-spec.org/acquisition/open-access', 'identifier': 'readable'})
```
Apophenia, line 277: The environment variable used by this function seemed to be different when I ran the tests individually vs as a set, and then seemed to also behave different on my machine vs QA. I wondered if it was being set by TestFlaskAPI in some cases but cleared to an empty string in others. In any case, adding the helper setup and teardown seemed to normalize it to a "test" environment.

services/google_drive_service.py
```diff
@@ -2,39 +2,58 @@
...
-        return None
-
-    return file
+class GoogleDriveService:
+    def __init__(self):
+        ssm_service = SSMService()
+        if os.environ['ENVIRONMENT'] == 'production':
+            SERVICE_ACCOUNT_FILE = ssm_service.get_parameter('arn:aws:ssm:us-east-1:946183545209:parameter/drb/production/google-drive-service-key')
+        else:
+            SERVICE_ACCOUNT_FILE = ssm_service.get_parameter('arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/google-drive-service-key')
```
Apophenia, line 23: Would it be better to encapsulate the ARNs in e.g. a dictionary within the SSM service itself?

kylevillegas93: could you do`ssm_service.get_parameter(f'arn.../{os.environ.get('ENVIRONMENT', 'qa')}...` to make this easier?

services/google_drive_service.py
```diff
@@ -2,39 +2,58 @@
...
+        service_account_info = json.loads(SERVICE_ACCOUNT_FILE)
+        scopes = ['https://www.googleapis.com/auth/drive',
+        'https://www.googleapis.com/auth/drive.file',
+        'https://www.googleapis.com/auth/drive.metadata']
```
kylevillegas93, line 27: just a nit on formatting - might be best to do:
```
= [
   'https://www.googleapis.com/auth/drive',
   ...
]
```

services/google_drive_service.py
```diff
@@ -2,39 +2,58 @@
...
+            logger.warning(f"HTTP error occurred when downloading Drive file {file_id}: {error}")
+            return None
+        except Exception as err:
+            logger.exception(f"Error occurred when downloading Drive file {file_id}")
+            return None
+
+        return file
+
+    def get_file_metadata(self, file_id: str) -> Optional[str]:
+        # supportsAllDrives=True required as of 12/24 despite deprecation
```
kylevillegas93, line 52: is this still needed?

services/google_drive_service.py
```diff
@@ -2,39 +2,58 @@
...
+        request = self.drive_service.files().get(fileId=file_id, supportsAllDrives=True)
+        meta = request.execute()
```
kylevillegas93, line 54: what is meta? the variable name isn't quite clear to me - sorry!

kylevillegas93 (approved):
Overall, looks good - just a few minor comments!

services/google_drive_service.py
Apophenia: I wanted to leave this here in case anyone tried to refactor this code and referred to the Google documentation, which is [in some places incorrect](https://developers.google.com/resources/api-libraries/documentation/drive/v3/python/latest/drive_v3.changes.html). It looks like they added warnings to the codebase that the supportsAllDrives would be considered true by default after 2020, and planned to support Team drives by default, but this isn't the case at least for get_media().

Apophenia: I lost a lot of time to this before I just tried adding it anyway - I didn't want the same to happen to anyone else.

services/google_drive_service.py
Apophenia: updated

PR #490, opened 2024-12-23
[SFR-2433] Adding record id to items table

kylevillegas93:
## Description
- Adds record id to items table
- Adds cascade to delete items if record is deleted
- Does not index!

## Testing
`python -m main.py -p SeedLocalDataProcess -e local`

Screenshot 2024-12-23 at 12 48 15


jackiequach (approved):
Looks good

PR #491, opened 2024-12-23
[SFR-2434] Simplifying delete record implementation

kylevillegas93:
## Description
- Fixes a few bugs to get the backlist project working as expected
- Updates the delete logic to delete by items then editions and then works
- Still needs some refactoring - I think that I can get it to be cleaner

## Testing
- Tested ingesting and clustering a record
- Tested deleting a record

### DIFF
--- a/services/sources/publisher_backlist_service.py
+++ b/services/sources/publisher_backlist_service.py
@@ -6,6 +6,7 @@ import urllib.parse
 from typing import Optional
 import traceback
 from model import Record, Work, Edition, Item
+from sqlalchemy.orm import joinedload
 from urllib.parse import urlparse
 
 from logger import create_log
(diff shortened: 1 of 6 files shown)


Apophenia:
looking good so far - I think i'll have to shuffle some things around on my upload branch but it should be ok, but i'll wait to open it until this is merged.

services/sources/publisher_backlist_service.py
```diff
@@ -1,10 +1,12 @@
...
+import json
 import os
 import requests
-import json
 import urllib.parse
 from typing import Optional
-from model import Record, Work, Edition, Item, Link
+from model import Record, Work, Edition, Item
+from sqlalchemy.orm import joinedload
+from urllib.parse import urlparse
```
Apophenia, line 9: probably possible to just change this instance to be the same access pattern as the other urllib.parse usages

PR #492, opened 2024-12-24
Sfr 2390 download files from Drive and upload them to s3 as part of publishers project process

Apophenia:


services/sources/publisher_backlist_service.py
```diff
@@ -121,9 +131,22 @@ def get_records(
...
+                    if not file:
+                        logger.warn(f'Could not retrieve file for {record_metadata_dict["id"]} from Drive, skipping')
+                        continue
+                    bucket = self.file_bucket # TODO: if record is limited access, upload to limited access bucket
+                    s3_path = f'{self.title_prefix}/{record_metadata_dict["Publisher (from Projects)"][0]}/{file_name}'
+                    s3_response = self.s3_manager.putObjectInBucket(file.getvalue(), s3_path, bucket)
+                    if not s3_response.get('ResponseMetadata').get('HTTPStatusCode') == 200:
+                        logger.warn(f'Could not upload file for {record_metadata_dict["id"]} to s3, skipping')
+                        continue
+                    s3_url = f'https://{bucket}.s3.amazonaws.com/{s3_path}'
```
Apophenia, line 144: We could build this dynamically from the bucket host if we don't want to hard-code s3 here

kylevillegas93 (approved):

PR #493, opened 2024-12-26
SFR-2419: Deprecate singular noun endpoints

jackiequach:
# Description
- Adds `deprecated` custom decorator to add `Warning` header to responses
- Adds deprecation warning to singular noun endpoints

# Testing
- docker compose up, hit the /edition, /link, /collection, /work endpoints and verify the warning header is present

Note: we could also update the status code (i.e. 299) for the response to flag it to users of the API


kylevillegas93 (approved):
These changes look good! Will the deprecated annotation also add the warning for the new routes or just the old?

jackiequach:
> These changes look good! Will the deprecated annotation also add the warning for the new routes or just the old?

Based on my testing, they should only be added to the old routes.

PR #494, opened 2024-12-30
SFR-2430: Set flags based on limited access permissions

Apophenia:
Still lacking full testing for the flag logic but ready for re-review!


services/sources/publisher_backlist_service.py
```diff
@@ -283,3 +284,20 @@ def generate_manifest(record, source_url, manifest_url):
...
+    @staticmethod
+    def parse_permissions(permissions: str) -> dict:
+        if permissions == 'Full access':
+            return {'is_downloadable': True, 'is_login_limited': False}
+        if permissions == 'Partial access/read only/no download/no login':
+            return {'is_downloadable': False, 'is_login_limited': False}
+        if permissions == 'Limited Access/login for read/no download':
+            return {'is_downloadable': False, 'is_login_limited': True}
+        else:
+            return {'is_downloadable': False, 'is_login_limited': True}
```
kylevillegas93, line 297: This looks good! Does this method need to be static?

services/sources/publisher_backlist_service.py
```diff
@@ -26,9 +26,10 @@ def __init__(self):
         self.s3_manager.createS3Client()
         self.title_prefix = 'titles/publisher_backlist'
         self.file_bucket = os.environ['FILE_BUCKET']
-        
+        self.limited_file_bucket = self.build_limited_bucket(self.file_bucket)
```
kylevillegas93, line 29: what are your thoughts on just using the environment config variable?  

```
f"drb-files-limited-${os.environment.get('ENVIRONMENT', 'qa')}"
```

Apophenia: I think that works fine, but I was (I guess arbitrarily) trying to avoid it because it felt like hard-coding the variable - like, why have os.environ['FILE_BUCKET'] if the limited bucket is just going to be concatenated from a string - but maybe it's not worth worrying about.

Apophenia: It also looks like Amazon has changed their pricing structure such that revealing bucket names is no longer a DDOS vector via 400-level errors, so I am less concerned about the bucket names being publicly discoverable: https://docs.aws.amazon.com/AmazonS3/latest/userguide/ErrorCodeBilling.html

services/sources/publisher_backlist_service.py
```diff
@@ -149,17 +150,24 @@ def get_records(
...
                     logger.error(f'Failed to retrieve file for {record_metadata.get("DRB_Record ID")} from Google Drive')
                     continue
 
-                bucket = self.file_bucket # TODO: if record is limited access, upload to limited access bucket
-                s3_path = f'{self.title_prefix}/{record_metadata["Publisher (from Projects)"][0]}/{file_name}'
+                record_permissions = self.parse_permissions(record_metadata.get('Access type in DRB (from Access types)')[0])
+                if not record_permissions['is_login_limited']:
+                    bucket = self.file_bucket
+                else:
+                    bucket = self.limited_file_bucket
```
kylevillegas93, line 169: You could do ternary here to make it 1 line but no worries.

services/sources/publisher_backlist_service.py
```diff
@@ -215,31 +223,31 @@ def get_publisher_backlist_records(self,
...
             records_response_json = records_response.json()
-            
+
             publisher_backlist_records.extend(records_response_json.get('records', []))
 
         return publisher_backlist_records
-    
-    def add_has_part_mapping(self, s3_url: str, record: Record):        
+
+    def add_has_part_mapping(self, s3_url: str, record: Record, is_downloadable: bool=False, is_login_limited: bool=True):
```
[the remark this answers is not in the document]
Apophenia: I think requires_login is way better, I'll change it!

Apophenia:
> Do we need to change the has_part flags for the manifest as well based on the parsed permissions?

I'll cross-check the original Michigan ingest process, but I don't think so. We discussed different approaches, but I believe the way it was implemented was that all manifests are stored in the public bucket, and the manifests contain fulfill links, but those fulfillment links are what require authentication (and they're stored in the private buckets).

services/sources/publisher_backlist_service.py
```diff
@@ -215,33 +228,32 @@ def get_publisher_backlist_records(self,
...
             'embed': False,
-            **({'nypl_login': True} if 'in_copyright' in record.rights else {})
+            'nypl_login': requires_login,
         }
 
-        record.has_part.append('|'.join([item_no, s3_url, record.source, media_tpye, json.dumps(flags)]))
+        record.has_part.append('|'.join([item_no, s3_url, record.source, media_type, json.dumps(flags)]))
 
-    def store_pdf_manifest(self, record: Record):
+    def store_pdf_manifest(self, record: Record, requires_login: bool=True):
```
kylevillegas93, line 256: Just confirming but do we want have a default value here and does it make sense for it to be True? Seems like in all cases this function will be called with that param set.

kylevillegas93 (approved):
Looks good!

services/sources/publisher_backlist_service.py
Apophenia: I'll remove the default here - I wanted to consistently fall back to the least permissive option to reduce the likelihood of something being set to open-access because of malformed input data, but I agree that this is not such a case.

PR #495, opened 2025-01-02
SFR-2438: Fix Fulfill URL Manifest Process

kylevillegas93:
## Description
- The Fulfill URL Manifest process was using different incorrect class variables
- This change fixes typos!

## Testing
`python main.py -p FulfillURLManifestProcess -e local -i complete`
`python main.py -p FulfillURLManifestProcess -e local -i daily`
`make unit`

### DIFF
--- a/api/utils.py
+++ b/api/utils.py
@@ -41,7 +41,7 @@ class APIUtils():
         'muse': 4,
         'met': 5,
         'isac': 6,
-        'UofM': 7,
+        'UofMichigan': 7,
         'UofSC': 8,
         'hathitrust': 9,
         'oclc': 10,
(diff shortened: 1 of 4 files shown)


api/utils.py
```diff
@@ -41,6 +41,7 @@ class APIUtils():
         'muse': 4,
         'met': 5,
         'isac': 6,
+        'UofMichigan': 7,
         'UofM': 7,
```
Apophenia, line 45: Since we don't have any files ingested in production under UofM, do we need both?

kylevillegas93: I think since we have both in QA, we will need both!

mitri-slory (approved):

PR #496, opened 2025-01-03
SFR-2444: Updating analytics script to use new S3 log bucket paths

kylevillegas93:
## Description
- Changes the analytics script to pull from the limited files bucket logs rather than the ump files bucket logs


Apophenia (approved):
I think this looks good, although I can imagine another publisher requiring a different configuration -- for example, if a publisher allows full view & pdf download without login, but requires analytics, the bucket configuration would look different (at least, based on the way we're setting up the ingest).

PR #498, opened 2025-01-03
NO-REF: Remove the development config

kylevillegas93:
## Description
- The development config is no longer in use!


Apophenia (approved):
I think this is outdated and unused, cool with me to remove it.

PR #500, opened 2025-01-07
SFR-2440/search_sort_desc_date

ayan1229:
This PR does the following 
- sort search results by descending date order and validates the the language filter selection on the backend


tests/integration/api/test_search_sort_desc.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+import requests
+from .constants import API_URL
+from .utils import assert_response_status
+
+@pytest.mark.parametrize("endpoint, expected_status", [
+    ("/search?query=keyword%3ANASA&sort=date%3Adesc&filter=language%3ASerbian", 200),
```
kylevillegas93, line 7: Recommend using `urllib.parse.quote` to escape the URL keys just to make it more readable.

tests/integration/api/test_search_sort_desc.py
```diff
@@ -0,0 +1,31 @@
...
+])
+def test_search_language_and_date_sorting(endpoint, expected_status):
+    url = API_URL + endpoint
+    response = requests.get(url)
+
+    assert response.status_code is not None
+    assert_response_status(url, response, expected_status)
+
+    if expected_status == 200:
```
kylevillegas93, line 16: the assert_response_status above will cause the test to fail if the expected_status is not 200. I think we can remove this if

tests/integration/api/test_search_sort_desc.py
kylevillegas93: You can create a separate part of the tuple here for the language filter. Below you can then do 

```
assert item.get('language') == expected_language, f"Unexpected language: {item.get('language')}"
```

tests/integration/api/test_search_sort_desc.py
```diff
@@ -0,0 +1,31 @@
...
+
+        previous_date = None
+        for item in response_json.get('results', []):
+            current_date = item.get('publication_date')
+            assert current_date is not None, "Publication date is missing"
+            
+            if previous_date:
+                assert current_date >= previous_date, f"Publication dates are not sorted in descending order: {previous_date} > {current_date}"
+            
+            previous_date = current_date
```
kylevillegas93, line 31:
```
publication_dates = [item.get('publication_date') for item in response_json.get('results', [])]
assert all(publication_dates is not None for publication_date in publication_dates), f"Some publication dates are missing: {dates}"
assert publication_dates == sorted(publication_dates, reverse=True), f"Dates are not sorted in descending order: {dates}"
```

tests/integration/api/test_search_sort_desc.py
```diff
@@ -0,0 +1,31 @@
...
+        for item in response_json.get('results', []):
+            assert item.get('language') == 'Serbian', f"Unexpected language: {item.get('language')}"
+
+        previous_date = None
+        for item in response_json.get('results', []):
+            current_date = item.get('publication_date')
+            assert current_date is not None, "Publication date is missing"
+            
+            if previous_date:
+                assert current_date >= previous_date, f"Publication dates are not sorted in descending order: {previous_date} > {current_date}"
```
kylevillegas93, line 29: Is it ascending or descending order?

tests/integration/api/test_search_sort_desc.py
kylevillegas93, line 1: Ok probably makes sense to just do the sort order as part of the paramerterization of the test case instead of a separate file

kylevillegas93 (approved):

PR #501, opened 2025-01-07
SFR-2452: CLACSO Mapping

mitri-slory:
This PR focuses on creating a mapping for the new source CLACSO that will soon be added to the DRB ETL pipeline for ingestion.

### DIFF
--- a/mappings/clacso.py
+++ b/mappings/clacso.py
@@ -1,0 +1,23 @@
+from .json import JSONMapping
+
+class CLACSOMapping(JSONMapping):
+    def __init__(self, source):
+        super().__init__(source, {})
+        self.mapping = self.createMapping()
+    
+    def createMapping(self):
+        return {
+            'title': ('title', '{0}'),
+            'authors': ('authors', '{0}|||true'),
+            'abstract': ('description', '{0}'),
+            'dates': [('publication_date', '{0}|publication_date')],
+            'identifiers': 
+                [('isbn', '{0}|isbn')]
+            ,
+            'spatial': ('edition_location', '{0}'),
+            'has_part': [('url', '1|{0}|clacso|application/pdf|{{"catalog": false, "download": true, "reader": false, "embed": false}}')] 
+        }
+
+    def applyFormatting(self):
+        self.record.source = 'clacso'
+        self.record.source_id = f'{self.record.source}'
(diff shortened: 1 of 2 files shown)


mappings/clacso.py
```diff
@@ -0,0 +1,23 @@
...
+        return {
+            'title': ('title', '{0}'),
+            'authors': ('authors', '{0}|||true'),
+            'abstract': ('description', '{0}'),
+            'dates': [('publication_date', '{0}|publication_date')],
+            'identifiers': 
+                [('isbn', '{0}|isbn')]
+            ,
+            'spatial': ('edition_location', '{0}'),
+            'has_part': [('url', '1|{0}|clacso|application/pdf|{{"catalog": false, "download": true, "reader": false, "embed": false}}')]
```
jackiequach, line 18: Is there somewhere I can read about how these pipes work?

kylevillegas93: Unfortunately no but you bring up a good point. Is there a way we can better abstract this data format or make it more readable?

mitri-slory: I agree that it should be a better way to showcase this data format and for right now the FRBRization process needs the data to be formatted in this way so this can be a future investigation in changing the format of the data the FRBRization process receives.

kylevillegas93 (approved):

mappings/clacso.py
kylevillegas93: Sounds good. Would y'all mind creating a ticket to investigate/refactor the has_part mapping logic/model when you get the chance? Thanks!

kylevillegas93: @jackiequach to help cipher the has_part field, it's pipe delimited. The first part is the index of the part, the second part is the source, the third part is the media type, and the fourth part are the flags! The flags are the one that y'all see on the frontend to determine when to show the download and read online buttons or login!

jackiequach: Thanks this is helpful!

mitri-slory: > Sounds good. Would y'all mind creating a ticket to investigate/refactor the has_part mapping logic/model when you get the chance? Thanks!

I just made a ticket addressing this: [https://newyorkpubliclibrary.atlassian.net/browse/SFR-2469](url)

PR #505, opened 2025-01-09
SFR-2741: Run CI via Github Actions

kylevillegas93:
## Description
- Run CI via Github Actions
- This first pass runs the API tests that run against QA
- Making a few changes to load different env variables depending on the environment 
- There's going to be some additional work to get all of the tests running in the QA environment

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,17 +2,10 @@ name: 'Continuous Integration'
 
 on:
   workflow_call:
-    # TODO: determine what secrets we need to add to Github or get them from parameter store
     secrets:
       AWS_ACCESS_KEY_ID:
         required: true
       AWS_SECRET_ACCESS_KEY:
-        required: true
-      NYPL_API_CLIENT_ID:
-        required: true
-      NYPL_API_CLIENT_SECRET:
-        required: true
-      NYPL_API_TOKEN_URL:
         required: true
 
 permissions:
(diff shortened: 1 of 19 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    env:
+      AWS_ACCESS: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      ENVIRONMENT: qa
+    steps:
+      - uses: actions/checkout@v2
```
jackiequach, line 22: We should use checkout@v4 since v2 is running on a deprecated version of node.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+      - name: Cache dependencies
+        uses: actions/cache@v2
```
jackiequach, line 28: cache@v2 will be deprecated next month

.github/workflows/ci.yaml
kylevillegas93: good catch - upgrading!

.github/workflows/ci.yaml
kylevillegas93: ooop upgrading!

jackiequach (approved):

PR #507, opened 2025-01-09
SFR-2453: CLACSO Ingest Process

mitri-slory:
This PR focuses on the foundation of the CLACSO Ingest Process. Any code lines that involve the CLACSOService are commented out as this service will be expanded upon in future PRs that will incorporate the logic for retrieving CLACSO records.


kylevillegas93:
@jackiequach what are your thoughts on this? Overall looks good! 

I think we are getting to a point where we may just want to have one process called Ingest and we pass a -source parameter that specifies from which source to ingest!

jackiequach (approved):
lgtm

PR #509, opened 2025-01-09
SFR-2441: Search format type

ayan1229:
This test validates if format:readable in request url for search
then either reader or embed value should be True


tests/integration/api/test_search_format.py
```diff
@@ -0,0 +1,24 @@
...
+    
+    works = response_json.get('data', {}).get('works', [])
+     
+    for work in works:
+        for edition in work.get('editions', []):
+            for item in edition.get('items', []):
+                if any(link.get('mediaType') == "text/html" for link in item.get('links', [])):
+                    flags = item.get('links', [])[0].get('flags', {})
+                    assert flags.get('reader', False) or flags.get('embed', False), \
+                        f"Flags validation failed: {flags}"
```
kylevillegas93, line 24: Let's pull the html links out here and check to make sure that the reader flag is false and the embed is true for each one. 

The logic you have here checks if any link has a media type of text/html but then grabs the first link's flags. However, the link at that index may not have a mediaType of "text/html". 

You could potentially go further here and check other media type and ensure that the reader flag is True (I think).

tests/integration/api/test_search_format.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+import requests
+from urllib.parse import quote
+from .constants import API_URL
+
+@pytest.mark.parametrize("endpoint", [
+    ("/search?query=keyword:NASA&filter=format:readable")
+])
+def test_readable_format_flags(endpoint):
```
kylevillegas93, line 9: I think what we are testing here is that the search format filter is working as expected. I recommend changing the test name to reflect that.

tests/integration/api/test_search_format.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+import requests
+from urllib.parse import quote
+from .constants import API_URL
+
+@pytest.mark.parametrize("endpoint", [
+    ("/search?query=keyword:NASA&filter=format:readable")
```
kylevillegas93, line 7: You could also try to test downloadable format

kylevillegas93:
@ayan1229 Test overall looked good. I do recommend focusing on keeping things simple! No need to dive into the data too much. Just want to use the bear minimum to confirm that the format filter is working. 

I went ahead and made some changes to simplify things. I recommend checking out list comprehension! 

Also, the quote wasn't working correctly so the test wasn't actually calling the search endpoint. I've fixed that but I recommend going through the rest of the tests to confirm that they are working as expected.

kylevillegas93 (approved):

PR #510, opened 2025-01-10
NOREF: Test limited access permissions & fix error

Apophenia:
- Fixes error where required inputs were not passed
- Adds tests


tests/unit/test_publisher_backlist_service.py
kylevillegas93, line 1: Let's make sure everything is in snake case! I see a couple things like testService that need to be switched!

tests/unit/test_publisher_backlist_service.py
```diff
@@ -0,0 +1,139 @@
...
+            "Last Modified": "2024-06-20T17:15:35.000Z",
+            "DRB_Record ID": "recRvPlHdLkM0KIem",
+            "Project Name (from Project)": ["Test Backlist"],
+            "Access type in DRB (from Access types)": [
+                "Limited access/login for read & download"
+            ],
+        },
+    }]
+                
+    def test_get_records(self, test_airtable_records, testService, mocker):
```
kylevillegas93, line 112: I recommend formatting this test for readability. What I have found works well is the following format:
```
test setup, mocks, environment variables 
new line
function/method that is under test
new line
assertions
```

kylevillegas93, line 112: We are testing to get the records, but we are mostly making assertions on the mapping. Should we focus more on asserting we can get the records here vs mapping the records properly?

tests/unit/test_publisher_backlist_service.py
```diff
@@ -0,0 +1,139 @@
...
+        limited_access_manifest_link = mapped_limited_access_record.has_part[0].split('|')
+        assert limited_access_manifest_link[1] == 'https://regular-bucket.s3.amazonaws.com/manifests/publisher_backlist/Test Backlist/recRvPlHdLkM0KIem.json'
+        assert limited_access_manifest_link[2] == 'Test Backlist'
+        assert limited_access_manifest_link[3] == 'application/webpub+json'
+        assert limited_access_manifest_link[4] == '{"catalog": false, "download": false, "reader": true, "embed": false, "fulfill_limited_access": false}'
+        limited_access_title_link = mapped_limited_access_record.has_part[1].split('|')
+        assert limited_access_title_link[1] == 'https://limited-bucket.s3.amazonaws.com/titles/publisher_backlist/Test Backlist/limited_access.pdf'
+        assert limited_access_title_link[2] == 'Test Backlist'
+        assert limited_access_title_link[3] == 'application/pdf'
+        assert limited_access_title_link[4] == '{"catalog": false, "download": true, "reader": false, "embed": false, "nypl_login": true}'
```
kylevillegas93, line 139: Can we simplify the assertions here? It may be helpful to write helper functions that build the links.

tests/unit/test_publisher_backlist_service.py
Apophenia: I agree, and I think the mapping is tested elsewhere, but the part I was struggling with breaking out into a unit test was the flag block, which is not done in the mapping or covered in the mapping testing but is functionally part of the mapping.

Apophenia:
I'm now seeing a pipe error on the publisher split in the kmeans clustering, similar to what was happening with other fields earlier, so I'm going to try and fix that as well

tests/unit/test_publisher_backlist_service.py
kylevillegas93: You could potentially test parse_permissions and add_has_part_mapping and perhaps store_pdf_manifest to ensure the flags are set properly.

Apophenia:
> I'm now seeing a pipe error on the publisher split in the kmeans clustering, similar to what was happening with other fields earlier, so I'm going to try and fix that as well

I'm now confident this was an issue with my local test environment still having a leftover title with a doubly-mapped list, so just going to go ahead and refactor.

PR #514, opened 2025-01-13
SFR-2478: Setting WEBPUB_CONVERSION_URL as a constant

kylevillegas93:
## Description
- https://epub-to-webpub.vercel.app is constants across all environments
- This change sets WEBPUB_CONVERSION_URL as a constant where it's used

### DIFF
--- a/config/example.yaml
+++ b/config/example.yaml
@@ -65,9 +65,6 @@ MUSE_MARC_URL: xxx
 # DOAB OAI-PMH endpoint
 DOAB_OAI_URL: xxx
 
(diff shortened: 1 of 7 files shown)


mitri-slory (approved):
This looks good. One lingering thought I have is I'm assuming the reason for converting WEBPUB_CONVERSION_URL into a constant from an environment variable is due to the variable only being used in one file in the codebase. Is that correct?

kylevillegas93:
@mitri-slory - that's correct and it's constant across local, qa and production environments!

PR #518, opened 2025-01-13
SFR-2748: removing tugboat qa config

kylevillegas93:
## Description
- Removes tugboat qa config, which we no longer use!


Apophenia (approved):
Awesome!

PR #520, opened 2025-01-14
SFR-2478: Make Hathi urls constant

kylevillegas93:
## Description
- The Hathi URLs are constant across envs, as such let's make them constant!


jackiequach (approved):
Looks good

PR #524, opened 2025-01-15
NO-REF: Fix local elastic search index creation

kylevillegas93:
## Description
- Ok so the Index inner class name field is set before the os environment variables are set. Therefore, the index is none if we try to create it at that point. 
- This change just passes in the index from the elastic search manager when initializing the ESWork.


Apophenia (approved):
Looks good!

PR #525, opened 2025-01-15
SFR-2456: Scraping CLACSO Metadata

mitri-slory:
This PR focuses on scraping CLACSO metadata using DSpace which is also utilized in the DOAB mapping and ingest process. The CLACSO service process stops at the applyMapping() method since this PR focuses on just scraping the metadata and checking if the CLACSO record has the right metadata before it's added to the database.


mitri-slory:
With my current commit I was able to apply the CLACSO mapping successfully on a single CLACSO record and I got this as a result when printing out the metadata for the record:

```
Record metadata: Title: Las tres “d” de las vacunas del SARS-COV-2 en América Latina y el Caribe : determinación, dependencia y descoordinación, Authors: ['Basile, Gonzalo|||true', 'Feo Istúriz, Oscar|||true'], Abstract: El presente artículo busca caracterizar la arquitectura global que entro en operaciones en el campo de la investigación, desarrollo, producción y distribución de vacunas del SARS-CoV-2. Así se presenta una descripción de los engranajes entre el complejo médico industrial financiero a través de la articulación entre las Big Pharma con la Organización Mundial de la Salud (OMS), Estados del Norte global y actores de la salud global liberal en la respuesta a la emergencia de salud pública del SARS-CoV-2 a través de COVAX, CEPI y GAVI.
A su vez, describe los eslabones críticos en todo el proceso de la crisis epidemiológica internacional actual y las tensiones de soberanía sanitaria regional como fundamentos para comprender el acceso a vacunas como bienes públicos desde América Latina y el Caribe a partir de las categorías denominadas las tres “D”: determinación, dependencia y descoordinación. Sobre esta matriz de análisis de eslabones bajo las tres “D”, el presente artículo entonces desarrolla una revisión sistemática desde el pensamiento crítico latinoamericano en salud de estos flujos para caracterizar, analizar e interpretar el acceso y uso de vacunas que se reproducen en el escenario regional en respuesta al ciclo de
SARS-CoV-2., Dates: ['2022-02-22T14:26:26Z', '2022-02-22T14:26:26Z', '2022-01'], Identifiers: ['https://biblioteca-repositorio.clacso.edu.ar/handle/CLACSO/16948|clacso'], has_part: ['1|https://biblioteca-repositorio.clacso.edu.ar/handle/CLACSO/16948|clacso|text/html|{"reader": false, "download": false, "catalog": false, "embed": true}']
```

There are still questions remaining for how to scrape the CLACSO metadata including:
1. Based on the CLACSO metadata on DSpace ([https://biblioteca-repositorio.clacso.edu.ar/oai/request?verb=ListRecords&metadataPrefix=oai_dc](url)) the edition location is not present so how will we fill out the spatial attribute in the CLACSO mapping?
2. How do we make sure to only add records that are monographs? (The dc:type filter can help with this)

I plan to continue cleaning up the formatting for the metadata in my future commits as I ponder these questions.

kylevillegas93:
@mitri-slory is the spatial field required? How many records in our database don't have this field?

mitri-slory:
> @mitri-slory is the spatial field required? How many records in our database don't have this field?

The spatial field is not required so if there isn't a way to access this info through DSpace then the field can be removed from the mapping

mitri-slory:
With my recent commit I focused on formatting the dates, source_id, identifiers, and has_part fields. After adding the formating_has_part method, the ingest process is running pretty slow so any advice on how to to speed up the performance based on my code would be appreciated.

mappings/clacso.py
```diff
@@ -1,23 +1,119 @@
...
+                './dc:identifier/text()',
+                '1|{0}|clacso|text/html|{{"reader": false, "download": false, "catalog": false, "embed": true}}'
+            )]
         }
 
     def applyFormatting(self):
         self.record.source = 'clacso'
-        self.record.source_id = f'{self.record.source}'
+        handle_url = "https://biblioteca-repositorio.clacso.edu.ar/handle/CLACSO"
+        self.formatting_source_id(handle_url)
```
kylevillegas93, line 40: recommend changing these function names to `format_`. using the infinitive of the verb rather than the gerund I think is more explicit and shows action.

processes/ingest/clacso.py
```diff
@@ -11,38 +11,35 @@ class CLACSOProcess(CoreProcess):
...
+            else: 
+                logger.warning(f'Unknown CLACSO ingestion process type {self.process}')
+                return
             # for record in records:
             #     self.addDCDWToUpdateList(record)
             
-            self.saveRecords()
-            self.commitChanges()
+            # self.saveRecords()
+            # self.commitChanges()
```
kylevillegas93, line 37: oop! just need to undo the comments here

services/sources/clacso_service.py
```diff
@@ -0,0 +1,143 @@
...
+
+from constants.get_constants import get_constants
+from logger import create_log
+from mappings.clacso import CLACSOMapping
+from .source_service import SourceService
+from managers import RabbitMQManager, S3Manager, DOABLinkManager
+
+logger = create_log(__name__)
+
+class CLACSOService(SourceService):
```
kylevillegas93, line 16: Is the plan to use the DSpaceService once Jackie's PR is merged in?

kylevillegas93:
What are your thoughts on adding a functional or integration test for the ingest process? Something that we could potentially do as a separate PR/ticket.

mappings/clacso.py
```diff
@@ -1,23 +1,119 @@
...
+
+    def formatting_has_part(self, handle_url):
+        if self.record.has_part:
+            clean_has_part = [has_part for has_part in self.record.has_part if 'http' in has_part]
+            for has_part in clean_has_part:
+                has_part_url = has_part.split('|')[1]
+                if handle_url in has_part_url or ("view" in has_part_url):
+                    response = requests.get(has_part_url)
+                    
+                    soup = BeautifulSoup(response.text, 'html.parser')
```
jackiequach, line 104: Can this var be renamed to be more descriptive (i.e clasco_page)?

mappings/clacso.py
```diff
@@ -1,23 +1,119 @@
...
+    def formatting_source_id(self, handle_url):
+        for identifier in self.record.identifiers:
+            if handle_url in identifier:
+                id = identifier.split('/')[-1]
+                self.record.source_id = f'{self.record.source}|{id}'
+                break
+
+    def formatting_medium(self):
+        if self.record.medium:
+            type_list = ['book', 'bookpart', 'part', 'chapter', 'bibliography', 'appendix', 'index',
```
jackiequach, line 55: bibliography is duplicated

mappings/clacso.py
```diff
@@ -1,23 +1,119 @@
...
+            'has_part': [(
+                './dc:identifier/text()',
+                '1|{0}|clacso|text/html|{{"reader": false, "download": false, "catalog": false, "embed": true}}'
+            )]
         }
 
     def applyFormatting(self):
         self.record.source = 'clacso'
-        self.record.source_id = f'{self.record.source}'
+        handle_url = "https://biblioteca-repositorio.clacso.edu.ar/handle/CLACSO"
```
jackiequach, line 39: make this into const

mappings/clacso.py
```diff
@@ -1,23 +1,119 @@
...
+                if '/' in medium:
+                    medium_value = medium.split('/')[-1]
+                    if medium_value.lower() in type_list:
+                        self.record.medium = medium_value.lower()
+                        break
+                elif '/' not in medium and medium.lower() in type_list:
+                    medium_value = medium
+                    self.record.medium = medium_value.lower()
+                    break
+        else:
```
jackiequach, line 67: Is this else clause necessary? I don't think we should include empty strings in the mapping

mappings/clacso.py
```diff
@@ -1,23 +1,119 @@
...
+                else:
+                    year_list.append(date)
+        self.record.dates = [f'{min(year_list)}|issued']
+
+    def formatting_identifiers(self):
+        if self.record.identifiers:
+            for index, identifier in enumerate(self.record.identifiers):
+                count_digits = 0
+                if '/' not in identifier:
+                    for char in identifier:
```
jackiequach, line 85: Can we pull this logic for determining if an identifier is ISSN or ISBN into utils functions to be reused in other mappings?

services/sources/clacso_service.py
mitri-slory: That is the plan but I think including a verification method for the has_part field to check if the field has a pdf or epub link and a verification method for the medium (this is more for CLACSO since DOAB mapping doesn't include the medium). These methods should help to achieve a 100% FRBR and cluster rate for CLACSO records since non monographic material will be filtered out by the medium and only CLACSO records with an available pdf or epub should be ingested as well.

mitri-slory:
> What are your thoughts on adding a functional or integration test for the ingest process? Something that we could potentially do as a separate PR/ticket.

Integration tests I plan to add in a later ticket when I add unit tests for the CLACSO ingest and mapping process. I might try to include functional tests as well.

mappings/clacso.py
mitri-slory: Good call Jackie.

mappings/clacso.py
mitri-slory: That's a good suggestion.

mappings/clacso.py
mitri-slory: The type of the medium is unicode and in the `createmapping()` method  I initialized it as a list since there can be multiple mediums in the DSpace xml file. I want to make sure the medium when it's None doesn't have an empty list as its value in the records table. For now I'll get rid of the else statement and see how the medium field values look in the database.

mappings/clacso.py
mitri-slory: Sure. I can put this method into the `base_mapping` just like `applyFormatting` for all the mappings.

mappings/base_mapping.py
```diff
@@ -29,6 +29,21 @@ def applyMapping(self):
...
+            count_digits = 0
+            if '/' not in identifier:
+                for char in identifier:
+                    if char.isdigit():
+                        count_digits += 1
+                if count_digits == 8:
+                    identifiers[index] = f'{identifier}|issn'
+                if count_digits == 10 or count_digits == 13:
+                    identifiers[index] = f'{identifier}|isbn'
+            else:
```
jackiequach, line 44: CLACSO specific logic should not be included in this. 

What are your thoughts on the following to clean up this method a bit: 
- Using the sum function for counting the number of digits with an iterable through the identifier string
- Using a dictionary instead of multiple if statements for determining the proper identifier type (ISSN/ISBN) and passing that through the f string

mappings/clacso.py
```diff
@@ -1,23 +1,108 @@
...
+                '1|{0}|clacso|text/html|{{"reader": false, "download": false, "catalog": false, "embed": true}}'
+            )]
         }
 
     def applyFormatting(self):
         self.record.source = 'clacso'
-        self.record.source_id = f'{self.record.source}'
+        if self.record.identifiers:
+            self.format_identifiers(self.record.identifiers)
+        print(self.record.identifiers)
```
jackiequach, line 45: left a print statement here

mappings/clacso.py
```diff
@@ -1,23 +1,108 @@
...
+                        self.record.medium = medium_value.lower()
+                        break
+                elif '/' not in medium and medium.lower() in type_list:
+                    medium_value = medium
+                    self.record.medium = medium_value.lower()
+                    break
+        else:
+            self.record.medium = ''
+
+    def format_dates(self):
```
jackiequach, line 75: You can use list comprehension to make this method cleaner (as you did with clean_has_part below)

mappings/clacso.py
```diff
@@ -1,23 +1,108 @@
...
+        self.format_medium()
+        self.format_dates()
+        self.format_has_part()
+
+    def format_source_id(self):
+        for identifier in self.record.identifiers:
+            if self.HANDLE_URL in identifier:
+                id = identifier.split('/')[-1]
+                self.record.source_id = f'{self.record.source}|{id}'
+                break
```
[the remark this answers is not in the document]
mitri-slory: That's a good call to make it more clear.

mappings/clacso.py
mitri-slory: I appreciate the suggestion Jackie.

mappings/base_mapping.py
mitri-slory: I'll implement the dictionary approach you suggested. I'll add  the '|clacso' string to the CLACSO mapping instead so all identifiers have that type by default.

mappings/base_mapping.py
```diff
@@ -29,6 +29,22 @@ def applyMapping(self):
 
     def applyFormatting(self):
         return self.record
+    
+    def format_identifiers(self, identifiers, source):
+        identifier_type_dict = {'8': 'issn', '10': 'isbn', '13': 'isbn'}
```
jackiequach, line 34: I think you can just use the numbers as the keys here so you don't need to use the str() function below. Unless there is a best practice against using nums as keys that I am unaware of.

mitri-slory: For some reason I recall a time when I put an integer as a key and I got a KeyError but from looking it up and testing it with this code I see that integers can serve as keys for python dictionaries. I appreciate the observation.

mappings/base_mapping.py
```diff
@@ -29,6 +29,22 @@ def applyMapping(self):
...
         return self.record
+    
+    def format_identifiers(self, identifiers, source):
+        identifier_type_dict = {8: 'issn', 10: 'isbn', 13: 'isbn'}
+        for index, identifier in enumerate(identifiers):
+            count_digits = 0
+            if '/' not in identifier:
+                for char in identifier:
+                    if char.isdigit():
+                        count_digits += 1
```
kylevillegas93, line 40: hmm could there be cases where an identifier has 8 digits but also has other alpha non-numeric characters?

we could simplify this and use list comprehension: 

`numeric_identifier = [digit for char in identifier if char.isdigit()]`

you can then use `len` to check the number of digits. 

I also recommend using the get method with dictionaries, which can be helpful when we need to default to a specific value. We can do things like:

`identifier_type = identifier_type_dict.get(len(numeric_identifier), source)`

mappings/base_mapping.py
```diff
@@ -29,6 +29,22 @@ def applyMapping(self):
 
     def applyFormatting(self):
         return self.record
+    
+    def format_identifiers(self, identifiers, source):
```
kylevillegas93, line 33: @jackiequach @mitri-slory what are our thoughts on moving away from classes here and being purely functional? This could be a function inside a identifier utils file.

mappings/base_mapping.py
```diff
@@ -29,6 +29,22 @@ def applyMapping(self):
...
+        identifier_type_dict = {8: 'issn', 10: 'isbn', 13: 'isbn'}
+        for index, identifier in enumerate(identifiers):
```
kylevillegas93, line 35: I recommend using the `map` function for these cases where we are:
- applying a function to every item in a list
- returning the result as a list

mappings/clacso.py
```diff
@@ -1,23 +1,102 @@
...
+class CLACSOMapping(XMLMapping):
+
+    HANDLE_URL = "https://biblioteca-repositorio.clacso.edu.ar/handle/CLACSO"
+
+    
+    def __init__(self, source, namespace, constants):
+        super(CLACSOMapping, self).__init__(source, namespace, constants)
         self.mapping = self.createMapping()
     
     def createMapping(self):
```
kylevillegas93, line 16: I think the createMapping function hinders us quite a bit and is a roundabout way to map from A->B. What we are actually doing is going from A->C via the createMapping and applying the formatting in the XML mapping to get the values that we can from the XML source document. We then apply the formatting below to go from C->B.  

I think we can go directly from A->B. It will be more concise and easier to read. That being said, I'm open to pushback here. What are the benefits to keeping the createMapping function and this hierarchy of inherited mapping classes?

services/sources/clacso_service.py
jackiequach: I just merged in my PR. I am available to pair if you need help integrating the DSpaceService here.

mappings/base_mapping.py
mitri-slory: This sounds like a good idea to simplify and clean up this method.

mappings/base_mapping.py
mitri-slory: Sounds like a good idea

mappings/base_mapping.py
mitri-slory: I can agree with our talk from earlier that we can change the current mapping we have to a more functional one.

mappings/clacso.py
mitri-slory: From our conversation earlier, I agree with the shift to a more functional programming lens.

mappings/clacso.py
```diff
@@ -1,23 +1,102 @@
...
+                './dc:identifier/text()',
+                '1|{0}|clacso|text/html|{{"reader": false, "download": false, "catalog": false, "embed": true}}'
+            )]
         }
 
     def applyFormatting(self):
         self.record.source = 'clacso'
-        self.record.source_id = f'{self.record.source}'
+        if self.record.identifiers:
+            self.record.identifiers = list(map(lambda x: self.format_identifier(x, self.record.source), self.record.identifiers))
```
kylevillegas93, line 44: let's use id instead of x since it's more explicit.

kylevillegas93 (approved):
Overall, looks good! Let's move on to refactoring the mapping so that it is more functional!

PR #527, opened 2025-01-16
NO-REF: Adding ability to run scripts via main.py

kylevillegas93:
## Description
- Adds ability to run scripts via main.py and therefore on the EC2 instance, which has access to ES in order to delete the work documents

## Testing
`python main.py -sc cleanup_duplicate_works -e local`


mitri-slory (approved):
Good change! This is a cleaner way to run the scripts then having the `scriptrunner.py` file from before and matches how we run the processes as well.

PR #528, opened 2025-01-16
NO-REF: Updating cleanup duplicate works script

kylevillegas93:
## Description
- Adds a couple log statements 
- Handles not found exception in elastic search


Apophenia (approved):
Looks good!

PR #529, opened 2025-01-16
NO-REF: setting a batch size for deleting works

kylevillegas93:
## Description
- Sets a batch size of 100 when cleaning up duplicate works
- The ECS task ran out of memory!

## Testing
- ran script in QA


jackiequach (approved):
lgtm

PR #531, opened 2025-01-22
SFR-2476: Implement DSpace service and update DOAB ingest

jackiequach:
# Description
- Adds DSpace service that can be used by ingest processes
- Updates DOAB ingest process to use DSpace service

# Testing
`python main.py -p DOABProcess -e local -r "20.500.12854/62823"`


services/sources/dspace_service.py
```diff
@@ -0,0 +1,143 @@
...
+
+        self.rabbitmq_manager = RabbitMQManager()
+        self.rabbitmq_manager.createRabbitConnection()
+        self.rabbitmq_manager.createOrConnectQueue(
+            self.file_queue, self.file_route)
+
+        self.constants = get_constants()
+
+        self.base_url = base_url
+        self.SourceMapping = SourceMapping
```
kylevillegas93, line 41: I think we can do source_mapping for both the param and class variable here. 

It looks a bit funky when you instantiate an object but it's common practice to use snake case.

services/sources/dspace_service.py
```diff
@@ -0,0 +1,143 @@
...
+class DSpaceService(SourceService):
+    OAI_NAMESPACES = {
+        'oai_dc': 'http://www.openarchives.org/OAI/2.0/oai_dc/',
+        'dc': 'http://purl.org/dc/elements/1.1/',
+        'datacite': 'https://schema.datacite.org/meta/kernel-4.1/metadata.xsd',
+        'oapen': 'http://purl.org/dc/elements/1.1/',
+        'oaire': 'https://raw.githubusercontent.com/rcic/openaire4/master/schemas/4.0/oaire.xsd'
+    }
+
+    def __init__(self, base_url, SourceMapping: XMLMapping):
```
kylevillegas93, line 25: This is fun cuz I've actually never done this in python so we are learning ha!

I think when we say that the SourceMapping type is XMLMapping, we are expecting an actual object that is in instance of XMLMapping. 

Checkout typing for Python: https://docs.python.org/3/library/typing.html

I think we could do type[XMLMapping] or the like here.

services/sources/dspace_service.py
```diff
@@ -0,0 +1,143 @@
...
+        self.s3_manager.createS3Client()
+        self.s3_bucket = os.environ['FILE_BUCKET']
+
+        self.file_queue = os.environ['FILE_QUEUE']
+        self.file_route = os.environ['FILE_ROUTING_KEY']
+
+        self.rabbitmq_manager = RabbitMQManager()
+        self.rabbitmq_manager.createRabbitConnection()
+        self.rabbitmq_manager.createOrConnectQueue(
+            self.file_queue, self.file_route)
```
kylevillegas93, line 36: Hmm are these composed class instances being called? I think we can remove them if so and it makes it cleaner because this class is really only responsible for getting records from DSpace so it's following the single responsibility principle: https://stackify.com/solid-design-principles/

processes/ingest/doab.py
```diff
@@ -30,142 +27,57 @@ class DOABProcess(CoreProcess):
...
-        self.rabbitmq_manager = RabbitMQManager()
-        self.rabbitmq_manager.createRabbitConnection()
-        self.rabbitmq_manager.createOrConnectQueue(self.fileQueue, self.fileRoute)
+            records = []
 
-        self.constants = get_constants()
+            if self.process == 'daily':
+                records = self.dSpace_service.get_records(offset=self.offset, limit=self.limit)
+            elif self.process == 'complete':
+                records = self.dSpace_service.get_records(full_or_partial=True, offset=self.offset, limit=self.limit)
```
mitri-slory, line 46: I recommend changing the name `full_or_partial` to `full_import` since it's more clear what this variable does and the publisherBacklistProcess uses that variable name too.

kylevillegas93: Following up you could also just remove the full param!

services/sources/dspace_service.py
jackiequach: Moved. This was a good read, thanks!

jackiequach:
I'm working on adding integration tests for the DSpace service

processes/ingest/doab.py
```diff
@@ -30,145 +25,67 @@ class DOABProcess(CoreProcess):
...
-        except MappingError as e:
-            raise DOABError(e.message)
+            if records:
+                for record in records:
+                    self.manage_links(record)
+
+            self.saveRecords()
+            self.commitChanges()
+
+            logger.info(f'Ingested {len(self.records) if records else 1} DOAB records')
```
mitri-slory, line 69: Shouldn't the result of the else conditional be 0 if the records array is empty?

mitri-slory (commented):
These changes are looking good so far Jackie!

processes/ingest/doab.py
jackiequach: Yes, this should still return 0 since the len of the records array would be 0 if it is empty.

services/sources/dspace_service.py
```diff
@@ -0,0 +1,131 @@
...
+
+        if response.status_code == 200:
+            content = bytes()
+
+            for chunk in response.iter_content(1024 * 100):
+                content += chunk
+
+            return BytesIO(content)
+
+        raise DSpaceError(
```
kylevillegas93, line 125: I think we can just raise an Exception here unless downstream we need to handle it in a specific way.

services/sources/dspace_service.py
```diff
@@ -0,0 +1,131 @@
...
+            f'Received {response.status_code} status code from {url}')
+
+
+class DSpaceError(Exception):
```
kylevillegas93, line 129: If this is no longer needed - feel free to remove!

processes/ingest/doab.py
```diff
@@ -1,23 +1,18 @@
...
 from model import get_file_message
 
 
 logger = create_log(__name__)
 
+
 class DOABProcess(CoreProcess):
     DOAB_BASE_URL = 'https://directory.doabooks.org/oai/request?'
-    ROOT_NAMESPACE = {None: 'http://www.openarchives.org/OAI/2.0/'}
+    DOAB_IDENTIFIER = "oai:directory.doabooks.org"
```
kylevillegas93, line 15: can do single quotes here! we really need a formatter lol

processes/ingest/doab.py
```diff
@@ -30,145 +25,67 @@ class DOABProcess(CoreProcess):
...
         self.rabbitmq_manager.createRabbitConnection()
-        self.rabbitmq_manager.createOrConnectQueue(self.fileQueue, self.fileRoute)
+        self.rabbitmq_manager.createOrConnectQueue(
+            self.file_queue, self.file_route)
+
+        self.offset = int(args[5]) if args[5] else 0
+        self.limit = (int(args[4]) + self.offset) if args[4] else 10000
 
-        self.constants = get_constants()
+        self.dSpace_service = DSpaceService(base_url=self.DOAB_BASE_URL, source_mapping=DOABMapping)
```
kylevillegas93, line 43: recommend doing d_space_service or dspace_service so that it's not a mix-match of camel and snake case.

kylevillegas93 (approved):
These changes look great! Just a few minor non-blocking comments.

mitri-slory (approved):
These changes look really good Jackie!

PR #532, opened 2025-01-23
SFR-2484: Adding ability to cluster records by source

kylevillegas93:
## Description
- Adds ability to cluster records by source
- Cleanups the process logic so that we move away from process types - customer/complete.

## Testing
`make unit`
`python main.py -p ClusterProcess -e local-qa -src loc -l 1`

### DIFF
--- a/main.py
+++ b/main.py
@@ -35,6 +35,7 @@ def main(args):
     single_record = args.singleRecord
     limit = args.limit
     offset = args.offset
+    source = args.source
     options = args.options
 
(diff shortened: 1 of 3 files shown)


mitri-slory (approved):
This looks like a great QOL change. One question I have is that with this change we no longer have to input "complete" as a process in the terminal when running the ClusterProcess?

kylevillegas93:
@mitri-slory - that's correct! it's not necessary! how the cluster process will run just depends on the other params passed in!

PR #538, opened 2025-01-24
avoid duplication entry

ayan1229:


processes/local_development/seed_test_data.py
```diff
@@ -47,7 +47,15 @@ def runProcess(self):
             self.db_manager.close_connection()
 
     def save_test_record(self):
-        test_record = Record(**self.test_data)
-
-        self.db_manager.session.add(test_record)
-        self.db_manager.session.commit()
+        existing_record = self.db_manager.session.query(Record).filter_by(uuid=self.test_data['uuid']).first()
```
kylevillegas93, line 50: The test_data['uuid'] is going to change every time so existing record will always be None. 

I recommend using source_id which will remain constant

processes/local_development/seed_test_data.py
```diff
@@ -47,7 +47,15 @@ def runProcess(self):
...
+
+        if existing_record:
+            for key, value in self.test_data.items():
+                if hasattr(existing_record, key):
```
kylevillegas93, line 54: I also recommend not setting the uuid differently every time.

processes/local_development/seed_test_data.py
```diff
@@ -47,7 +47,15 @@ def runProcess(self):
...
+                    setattr(existing_record, key, value)
+            existing_record.date_modified = datetime.now(timezone.utc).replace(tzinfo=None)
+        else:
+            test_record = Record(**self.test_data)
+            self.db_manager.session.add(test_record)
+
+        self.db_manager.session.commit()
```
kylevillegas93, line 61: nit: let's add a newline here so github stops complaining.

processes/local_development/seed_test_data.py
ayan1229: Like you want us to give a value to UUID manually during creation?

kylevillegas93: Meaning we should not update the uuid here. it should remain the same!

PR #539, opened 2025-01-24
update frbr_status to complete

ayan1229:


processes/local_development/seed_test_data.py
```diff
@@ -47,15 +47,7 @@ def runProcess(self):
...
-        existing_record = self.db_manager.session.query(Record).filter_by(source_id=self.test_data['source_id']).first()
-
-        if existing_record:
-            for key, value in self.test_data.items():
-                if key != 'uuid' and hasattr(existing_record, key):
-                    setattr(existing_record, key, value)
-            existing_record.date_modified = datetime.now(timezone.utc).replace(tzinfo=None)
-        else:
-            test_record = Record(**self.test_data)
-            self.db_manager.session.add(test_record)
```
kylevillegas93, line 59: ooop - we still need this!

processes/local_development/seed_test_data.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, *args):
             'uuid' : uuid4(),
             'date_created' : datetime.now(timezone.utc).replace(tzinfo=None),
             'date_modified' :datetime.now(timezone.utc).replace(tzinfo=None),
```
kylevillegas93, line 19: You may not need to set these. I think these are set in the base postgres model class.

processes/local_development/seed_test_data.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, *args):
...
-            'frbr_status': 'to_do',
+            'frbr_status': 'complete',
             'cluster_status': False,
             "source": 'test data source',
```
kylevillegas93, line 22: Recommend pulling this up as a constant outside the class. Let's use something that we can use in other fields as well, e.g. 

TEST_SOURCE = 'test_source'

processes/local_development/seed_test_data.py
ayan1229: wait idk how it got deleted. Adding it back

processes/local_development/seed_test_data.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, *args):
...
             'authors': ['Ayan||true'],
```
kylevillegas93, line 23: Ok I think the has_part is going to cause you some trouble with the source!

Instead of met it should also be TEST_SOURCE. 

Lastly, I recommend just setting the identifiers to an empty list for now.

processes/local_development/seed_test_data.py
```diff
@@ -15,24 +16,22 @@ def __init__(self, *args):
...
+            'identifiers': [],
             'source_id': '4064148285|test',
             'contributors': ['Metropolitan Museum of Art (New York, N.Y.)|||contributor','Metropolitan Museum of Art (New York, N.Y.)|||repository','Thomas J. Watson Library|||provider'],
             'extent': ('11, 164 p. ;'),
             'is_part_of': ['Tauchnitz edition|Vol. 4560|volume'],
             'abstract': ['test abstract 1', 'test abstract 2'],
             'subjects': ['test subjects 1||'],
             'rights': ('hathitrust|public_domain|expiration of copyright term for non-US work with corporate author|Public Domain|2021-10-02 05:25:13'),
-            'has_part': ['1|example.com/1.pdf|met|text/html|{"catalog": false, "download": false, "reader": false, "embed": true}']
+            'has_part': ['1|example.com/1.pdf|{TEST_SOURCE}}|text/html|{"catalog": false, "download": false, "reader": false, "embed": true}']
```
kylevillegas93, line 34: Oof - this one is a bit tricky cuz we need to use an f string but we have some double quotes. Lemme help out here.

kylevillegas93 (approved):

PR #540, opened 2025-01-24
add clustering logic

ayan1229:


processes/local_development/seed_test_data.py
```diff
@@ -41,6 +41,11 @@ def runProcess(self):
             self.db_manager.createSession()
 
             self.save_test_record()
+
+            process_args = ['complete'] + ([None] * 4)
```
kylevillegas93, line 45: I recommend just running the cluster process against the record uuid. It's actually the 4th input. 

Recommend doing this:

```
cluster_process = ClusterProcess('complete', None, None, self.test_record['uuid'], None)

i think lol 
```

kylevillegas93 (approved):
looks good!

PR #542, opened 2025-01-27
NO-REF: Test that MET process saves manifest for each record 

ayan1229:
## Description
- Tests that the MET Process saves manifest for each record
- Adds a Part dataclass to be able to get the manifest easier

## Testing
```
pytest tests/functional/processes/ingest/test_met_process.py -s
```


kylevillegas93 (approved):
It's passing on my end. If it looks good on yours, let's ship it!

PR #543, opened 2025-01-27
SFR-2481/save_test_data

ayan1229:


processes/local_development/seed_test_data.py
```diff
@@ -68,3 +70,12 @@ def save_test_record(self):
...
         self.db_manager.session.commit()
+    
+    def save_test_data_ids_to_file(self):
+        record = self.db_manager.session.query(Record).filter_by(source_id=self.test_data['source_id']).first()
+        if record:
+            with open('test_data_ids.json', 'w') as f:
+                json.dump({
+                    'edition_id': str(record.edition_id),
+                    'uuid': str(record.uuid)
+                }, f)
```
kylevillegas93, line 81: I like this idea to share the test ids with the tests but I think we should use fixtures to seed the data instead of the process!

tests/integration/api/test_get_edition.py
```diff
@@ -1,10 +1,20 @@
 import pytest
 import requests
+import json
 from .constants import API_URL
 from .utils import assert_response_status
 
+@pytest.fixture(scope="module", autouse=True)
+def seed_test_data():
+    process = SeedTestDataProcess()
+    process.runProcess()
```
kylevillegas93, line 10: I think we can use seed the data directly from fixtures. Let's move this to conftest.py.

tests/integration/api/test_get_edition.py
```diff
@@ -1,10 +1,20 @@
...
+
+with open('test_data_ids.json', 'r') as f:
+    test_data_ids = json.load(f)
+    SEEDED_EDITION_ID = test_data_ids['edition_id']
+
 @pytest.mark.parametrize("endpoint, expected_status", [
-    ("/editions/1982731", 200),
+    ("/editions/{SEEDED_EDITION_ID}", 200),
```
kylevillegas93, line 17: Is there a way to grab the edition from a fixture after it's been seeded and use it here in the parametrize annotation?

test_data_ids.json
```diff
@@ -0,0 +1 @@
+{"edition_id": "19855", "uuid": "3264ccb1-a870-43d7-9496-2153ca99acc6"}
```
kylevillegas93, line 1: let's get rid of this file!

tests/conftest.py
```diff
@@ -16,3 +25,64 @@ def setup_env(pytestconfig, request):
 
     if not running_unit_tests and environment in ['local', 'qa']:
         load_env_file(environment, file_string=f'config/{environment}.yaml')
+
+@pytest.fixture(scope='module')
```
kylevillegas93, line 29: question on the module scope. does this run during the unit tests? 

we just want to run it during functional and integration tests

tests/conftest.py
```diff
@@ -16,3 +25,64 @@ def setup_env(pytestconfig, request):
...
+        json.dump({
+            'edition_id': edition_id,
+            'uuid': str(test_data['uuid'])
+        }, f)
+
+    return edition_id
+
+@pytest.fixture(scope='module')
+def seeded_edition_id(seed_test_data):
+    return seed_test_data
```
kylevillegas93, line 88: nit: let's add a newline

tests/integration/api/test_get_edition.py
```diff
@@ -1,17 +1,18 @@
...
+    ("/editions/{seeded_edition_id}", 200),
     ("/editions/00000000-0000-0000-0000-000000000000", 400),
     ("/editions/invalid_id_format", 400),
     ("/editions/", 404),
     ("/editions/%$@!*", 400)
 ])
-def test_get_edition(endpoint, expected_status):
-    url = API_URL + endpoint
+def test_get_edition(endpoint, expected_status, seeded_edition_id):
+    url = API_URL + endpoint.format(seeded_edition_id=seeded_edition_id)
```
kylevillegas93, line 15: Do the other tests pass even with the format? 

I think we can use lazy fixtures: https://stackoverflow.com/questions/42014484/pytest-using-fixtures-as-arguments-in-parametrize

tests/conftest.py
```diff
@@ -16,3 +25,64 @@ def setup_env(pytestconfig, request):
...
+    db_manager.session.commit()
+
+    item = db_manager.session.query(Item).filter_by(record_id=test_record.id).first()
+    edition_id = str(item.edition_id) if item else None
+
+    with open('test_data_ids.json', 'w') as f:
+        json.dump({
+            'edition_id': edition_id,
+            'uuid': str(test_data['uuid'])
+        }, f)
```
kylevillegas93, line 82: We don't have to dump this into a file since the tests should be able to use these fixtures !

tests/conftest.py
```diff
@@ -16,3 +25,64 @@ def setup_env(pytestconfig, request):
 
     if not running_unit_tests and environment in ['local', 'qa']:
         load_env_file(environment, file_string=f'config/{environment}.yaml')
+
+@pytest.fixture(scope='module')
+def db_manager():
+    db_manager = DBManager()
+    db_manager.createSession()
+    yield db_manager
+    db_manager.close_connection()
```
kylevillegas93, line 34: hell ya - looks super clean. nice way to do this to make sure we are closing the connection :)

tests/integration/api/test_get_edition.py
```diff
@@ -1,17 +1,19 @@
 import pytest
 import requests
+import json
 from .constants import API_URL
```
kylevillegas93, line 4: Ok so the API_URL will have to change depending in the env. I think we can get this from the config file - DRB_API_HOST

tests/conftest.py
```diff
@@ -16,3 +25,69 @@ def setup_env(pytestconfig, request):
...
+                setattr(existing_record, key, value)
+        existing_record.date_modified = datetime.now(timezone.utc).replace(tzinfo=None)
+        test_data['uuid'] = existing_record.uuid
+        test_record = existing_record
+    else:
+        test_record = Record(**test_data)
+        db_manager.session.add(test_record)
+    
+    db_manager.session.commit()
+
```
kylevillegas93, line 74: We need to cluster still right?

tests/conftest.py
```diff
@@ -16,3 +25,69 @@ def setup_env(pytestconfig, request):
...
+    item = db_manager.session.query(Item).filter_by(record_id=test_record.id).first()
+    edition_id = str(item.edition_id) if item else None
+
+    return {
+        'edition_id': edition_id,
+        'uuid': str(test_data['uuid'])
+    }
+
+@pytest.fixture(scope='module')
+def seeded_edition_id(request, seed_test_data):
```
kylevillegas93, line 84: You may want to double check but I think because we have seed_test_data here it'll automatically make sure seed_test_data is resolved. Thus, checking the request may not make sense.

kylevillegas93 (approved):
We may run into some issues on CI but let's see how we do!

kylevillegas93:
@ayan1229 let's get this out the door and make sure the tests are passing in CI. I'd get started on adding the other tests to CI as well.

PR #544, opened 2025-01-27
NO-REF: Asserting manifests are uploaded during ingest

kylevillegas93:
## Description
- Tests whether we have uploading manifests during each ingest process

## TODO
`make functional`


mitri-slory (approved):
This is a good change to make sure the ingest process is more successful. Once the tests are all successful I think this is good to go.

PR #550, opened 2025-02-03
SFR-2368: Refactor Chigaco ISAC Ingest

kylevillegas93:
## Description
- Refactors the Chicago ISAC ingest process to be more explicit, readable and functional
- Validates to ensure the record at least has an ISBN and a has_part URL
- Removes unit tests
- Adds functional test to validate mapping functionality
- Removes CoreProcess as an ancestor and composes the db_manager
- Adds a record buffer class to manage adding records to the database

## Testing
`make functional`


mitri-slory (approved):
These changes look good for refactoring these ISAC records. One remaining question I have besides the one I left for the ISAC mapping is why the unit tests were deleted for the process and mapping? Is it because the functional tests serve a similar purpose?

kylevillegas93:
Responded to these comments in Slack!

model/postgres/record.py
```diff
@@ -105,8 +122,11 @@ def get_parts(self) -> list[Part]:
 
         for part in self.has_part:
             index, file_url, source, file_type, flags = part.split('|')
+            
+            if not index:
+                index = 0
```
jackiequach, line 127: Will this be an issue if there are multiple parts without an index?

kylevillegas93: Possibly - yes if you use the get_parts function - lemme just make it optional

jackiequach (approved):
Looks good!

PR #551, opened 2025-02-04
NO-REF: Fix DOAB functional test error

jackiequach:
# Description
Updates while loop condition to run if offset is 0

# Testing
`make functional`

### DIFF
--- a/tests/functional/processes/ingest/test_doab_process.py
+++ b/tests/functional/processes/ingest/test_doab_process.py
@@ -5,7 +5,6 @@ from .assert_ingested_records import assert_ingested_records
 from .assert_uploaded_manifests import assert_uploaded_manifests
 
 
-@pytest.mark.skipif(os.getenv('IS_CI') == 'true', reason="Skipping in CI environment")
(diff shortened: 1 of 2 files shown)


kylevillegas93:
@jackiequach looks good - let's remove the pytest skip annotation on the functional test!

tests/functional/processes/ingest/test_doab_process.py
```diff
@@ -5,7 +5,6 @@
 from .assert_uploaded_manifests import assert_uploaded_manifests
 
 
-@pytest.mark.skipif(os.getenv('IS_CI') == 'true', reason="Skipping in CI environment")
```
kylevillegas93, line 8: oop - let's remove the unused import as well!

kylevillegas93 (approved):

tests/functional/processes/ingest/test_doab_process.py
jackiequach: oops didn't see this, I'll create a separate PR for this.

PR #555, opened 2025-02-05
NO-REF: Refactor NYPL Ingest + Mapping

kylevillegas93:


mitri-slory (commented):
I would say these changes look good. I'm curious about the the implementation of returning a Union of a RecordMapping list and a Generator as well and the thought process behind it.

PR #556, opened 2025-02-05
NO-REF: Simplify NYPL ingest

kylevillegas93:
## Description
- Removes unnecessary unit tests
- Removes CoreProcess as an ancestor of NYPLProcess
- Removes full_import usage
- Simplifies runProcess logic
- Uses yield and a generator which should hopefully be less memory intensive

## Testing
`make unit; make functional`


mitri-slory (approved):
This looks good. I appreciate the explanation here for the Generator and it makes sense to add it to help with memory.

PR #558, opened 2025-02-05
SFR-2500: Refactor DOAB mapping

jackiequach:
# Description
- Refactors and simplifies DOAB mapping to use functional programming 
- Adds data fields that were missing in the previous mapping. 
    - Unsure if datacite was completely replaced by dc in DOAB records. I looked on the [DOAB DSpace](https://directory.doabooks.org/oai/request?verb=ListRecords&metadataPrefix=oai_dc) and for the records I checked, ~10, datacite is not being used for many of the fields. As a result, I left the datacite field retrieval in case there are still records with those fields
- Updates .gitignore to allow test file for the functional test
- Removes unit tests for DOABMapping

# Testing
`python main.py -p DOABProcess -e local -r "20.500.12854/62823"`


.gitignore
```diff
@@ -3,7 +3,9 @@
 .coverage
 /.vscode/*
 *.pyc
-*.xml
+launcher manifest.xml
+nosetests.xml
+coverage.xml
```
kylevillegas93, line 8: we could do *.xml and !test-doab.xml if that works!

mappings/doab.py
```diff
@@ -2,129 +2,179 @@
...
-        }
+    def map_doab_record(self, doab_record, namespaces) -> Record:
+        identifiers, source_id = self._get_identifers(doab_record, namespaces=namespaces)
 
-    def applyFormatting(self):
-        # Set Identifiers
-        self.record.source = 'doab'
-        self.record.identifiers = self.parseIdentifiers()
+        if identifiers is None or len(identifiers) == 0:
+            return None
```
kylevillegas93, line 24: Nice - this is a good check!

mappings/doab.py
```diff
@@ -2,129 +2,179 @@
...
+            extent=[f'{extent} pages' for extent in extents],
+            abstract=doab_record.xpath('./dc:description/text()', namespaces=namespaces),
+            subjects=self._get_subjects(doab_record, namespaces=namespaces),
+            has_part=self._get_has_part(doab_record, namespaces),
+            rights=self._get_rights(doab_record, namespaces=namespaces),
+            date_created=datetime.now(timezone.utc).replace(tzinfo=None),
+            date_modified=datetime.now(timezone.utc).replace(tzinfo=None)
+        )
+
+    def _get_text_type_data(self,record, namespaces, field_xpath, format_string):
```
kylevillegas93, line 56: nit: space after self,

mappings/doab.py
```diff
@@ -2,129 +2,179 @@
...
+        datacite_authors = record.xpath('./datacite:creator/text()', namespaces=namespaces)
+        dc_authors = record.xpath('./dc:creator/text()', namespaces=namespaces)
 
-        # Clean up rights statements
-        self.record.rights = self.parseRights()
+        authors = datacite_authors or dc_authors
 
-        # Clean up links
-        self.record.has_part = self.parseLinks()
+        return [f'{author}|||true' for author in authors]
```
kylevillegas93, line 75: We may want to add `if author is not None or author == ''` to ensure we aren't adding empty strings

tests/functional/mappings/test_doab.py
```diff
@@ -0,0 +1,24 @@
...
+from model import Source
+from lxml import etree
+
+OAI_NAMESPACES = {
+    'oai_dc': 'http://www.openarchives.org/OAI/2.0/oai_dc/',
+    'dc': 'http://purl.org/dc/elements/1.1/',
+    'datacite': 'https://schema.datacite.org/meta/kernel-4.1/metadata.xsd',
+    'oapen': 'http://purl.org/dc/elements/1.1/',
+    'oaire': 'https://raw.githubusercontent.com/rcic/openaire4/master/schemas/4.0/oaire.xsd'
+}
```
kylevillegas93, line 11: These I think should be accessible form the DspaceService - so you could potentially grab them from there

tests/functional/mappings/test_doab.py
```diff
@@ -0,0 +1,24 @@
...
+        oaidc_records = etree.parse(f)
+
+    for record in oaidc_records.xpath('//oai_dc:dc', namespaces=OAI_NAMESPACES):
+        if record is None:
+            continue
+
+        parsed_record = DOABMapping(doab_record=record, namespaces=OAI_NAMESPACES).record
+        assert parsed_record.source == Source.DOAB.value
+        assert parsed_record.source_id == '20.500.12854/62823'
+        assert parsed_record.title == 'A World of Nourishment'
```
kylevillegas93, line 24: Nice looks good!

mappings/doab.py
```diff
@@ -2,129 +2,179 @@
...
+            uuid=uuid4(),
+            frbr_status=FRBRStatus.TODO.value,
+            cluster_status=False,
+            source=Source.DOAB.value,
+            source_id=source_id,
+            identifiers= identifiers,
+            authors=self._get_authors(doab_record, namespaces=namespaces),
+            contributors=self._get_contributors(doab_record, namespaces=namespaces),
+            title=title[0] if len(title) > 0 else None,
+            is_part_of=[f'{part}||series' for part in relations],
```
kylevillegas93, line 42: You may want to add if part checks in these list comprehensions to ensure we aren't adding strings like 'None||series' or '||series'

mappings/doab.py
```diff
@@ -2,129 +2,179 @@
...
+        return Record(
+            uuid=uuid4(),
+            frbr_status=FRBRStatus.TODO.value,
+            cluster_status=False,
+            source=Source.DOAB.value,
+            source_id=source_id,
+            identifiers= identifiers,
+            authors=self._get_authors(doab_record, namespaces=namespaces),
+            contributors=self._get_contributors(doab_record, namespaces=namespaces),
+            title=title[0] if len(title) > 0 else None,
```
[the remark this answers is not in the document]
jackiequach: Good idea, moved the check above with the identifiers

tests/fixtures/test-doab.xml
kylevillegas93, line 1: Nice - i like this idea of keeping the test files in a fixtures folder

services/sources/dspace_service.py
```diff
@@ -64,8 +64,7 @@ def get_records(self, full_import=False, start_timestamp=None, offset: Optional[
 
     def parse_record(self, record):
         try:
-            record = self.source_mapping(record, self.OAI_NAMESPACES, self.constants)
-            record.applyMapping()
+            record = self.source_mapping(record, self.OAI_NAMESPACES)
```
kylevillegas93, line 67: May just want to update this lhs variable name to parsed_record or the like to distinguish between record and parsed_record

kylevillegas93 (approved):
Looks good! just a couple last comments!

mappings/doab.py
```diff
@@ -1,130 +1,177 @@
...
+            authors=self._get_authors(doab_record, namespaces=namespaces),
+            contributors=self._get_contributors(doab_record, namespaces=namespaces),
+            title=title[0],
+            is_part_of=[f'{part}||series' for part in relations if part is not None or part == ''],
+            publisher=[f'{publisher}||' for publisher in publishers if publisher is not None or publisher == ''],
+            spatial=doab_record.xpath('./oapen:placepublication/text()', namespaces=namespaces),
+            dates=self._get_dates(doab_record, namespaces=namespaces),
+            languages=[f'||{language}' for language in languages if language is not None or language == ''],
+            extent=[f'{extent} pages' for extent in extents if extent is not None or extent == ''],
+            abstract=doab_record.xpath('./dc:description/text()', namespaces=namespaces),
```
kylevillegas93, line 45: Same as below - when we call xpath - i believe it tries to pull all elements that match that xpath. So we should just grab the first 1

mappings/doab.py
```diff
@@ -1,130 +1,177 @@
...
+        return new_ids, source_id
 
-            if rightsData[1] == '': continue
+    def _get_contributors(self, record, namespaces):
+        return self._get_text_type_data(record, namespaces, './datacite:contributor', '{}|||{}')
 
-            return '|'.join(rightsData)
+    def _get_dates(self, record, namespaces):
+        datacite_dates = self._get_text_type_data(record, namespaces, './datacite:date', '{}|||{}')
+        dc_dates = self._get_text_type_data(record, namespaces, './dc:date', '{}|||{}')
```
kylevillegas93, line 115: Here I recommend that if we do not know the date type, we may want to exclude it so 

```
{2021-04-20T15:29:32Z|||,2021-04-20T15:29:32Z|||,2012|||Issued}
```
becomes
```
{2012|||Issued}
```

jackiequach: Updated _get_text_type_data to do this since this also makes sense for the other fields we are calling it for (alternative identifiers and contributors)

mappings/doab.py
```diff
@@ -1,130 +1,178 @@
...
+            authors=self._get_authors(doab_record, namespaces=namespaces),
+            contributors=self._get_contributors(doab_record, namespaces=namespaces),
+            title=title[0],
+            is_part_of=[f'{part}||series' for part in relations if part],
+            publisher=[f'{publisher}||' for publisher in publishers if publisher],
+            spatial=doab_record.xpath('./oapen:placepublication/text()', namespaces=namespaces),
+            dates=self._get_dates(doab_record, namespaces=namespaces),
+            languages=[f'||{language}' for language in languages if language],
+            extent=f'{extent[0]} pages' if extent else None,
+            abstract=abstract[0] if abstract else None,
```
kylevillegas93, line 46: Nice - i like this - especially because it'll check if the array is both not none and not empty

PR #562, opened 2025-02-07
SFR-2497: Integrate CLACSO with DSpace Service

mitri-slory:
This PR alters the CLACSO ingest process to utilize the DSpaceService instead of the CLACSOService. It also adds a store_pdf_manifest method to the ingest process. One quick note is that the manifest link flags for the CLACSO records are all set to false since we want the CLACSO books to only be downloadable and not on the webreader on the frontend.


processes/ingest/doab.py
```diff
@@ -56,7 +56,7 @@ def runProcess(self):
             self.saveRecords()
             self.commitChanges()
 
-            logger.info(f'Ingested {len(self.records) if records else 1} DOAB records')
+            logger.info(f'Ingested {len(self.records) if records else 0} DOAB records')
```
jackiequach, line 59: This doesn't quite work due to the singleRecord condition above. It should probably be `len(records) if records else (1 if record else 0)`

processes/ingest/clacso.py
```diff
@@ -1,45 +1,106 @@
...
+        
+        manifest.addChapter(sourceURI, record.title)
+
+        manifest.links.append({
+            'rel': 'self',
+            'href': manifest_urI,
+            'type': 'application/webpub+json'
+        })
+
+        return manifest.toJson()
```
kylevillegas93, line 106: what are your thoughts on pulling these functions out? I think we have very similar if not the same code in several ingest processes?

processes/ingest/clacso.py
```diff
@@ -1,45 +1,106 @@
...
+            if records:
+                for record in records:
+                    self.store_pdf_manifest(record)
+                    self.addDCDWToUpdateList(record)
+
             self.saveRecords()
             self.commitChanges()
 
-            logger.info(f'Ingested {len(records)} CLACSO records')
+            logger.info(f'Ingested {len(self.records) if records else 0} CLACSO records')
```
kylevillegas93, line 58: Recommend that y'all move over to using the RecordBuffer and not inherit from the CoreProcess class.

tests/functional/processes/ingest/assert_ingested_records.py
```diff
@@ -12,7 +12,7 @@ def assert_ingested_records(source_name: str) -> list[Record]:
     records = (
         db_manager.session.query(Record)
             .filter(Record.source == source_name)
-            .filter(Record.date_modified > datetime.now(timezone.utc).replace(tzinfo=None) - timedelta(minutes=5))
+            .filter(Record.date_modified > datetime.now(timezone.utc).replace(tzinfo=None) - timedelta(minutes=10))
```
kylevillegas93, line 15: hmm - i wouldn't recommend pushing this back to 10 minutes. The records should be ingested fairly quickly. If pushing it back to 10 minutes makes the tests pass, it's likely because it's looking at records pulled in from a previous test run.

processes/ingest/doab.py
mitri-slory: Oh ok that makes sense. I appreciate the clarification.

processes/ingest/clacso.py
mitri-slory: I agree with pulling these methods out into the S3Manager like the `createManifestInS3` method is already. I'll make another PR to pull these methods into the S3Manager and change all the ingest processes to reflect the change.

processes/ingest/clacso.py
mitri-slory: Ok we'll try implementing the RecordBuffer.

tests/functional/processes/ingest/assert_ingested_records.py
mitri-slory: I'll change the timeout back to 5 minutes and look into fixing the tests failing.

jackiequach (commented):
Overall it's looking good, just a few minor comments

processes/ingest/clacso.py
```diff
@@ -1,45 +1,113 @@
...
-            else: 
-                logger.warning(f'Unknown CLACSO ingestion process type {self.process}')
-                return
-            for record in records:
-                self.addDCDWToUpdateList(record)
+
+            records = []
+
+            start_datetime = (
+                datetime.now(timezone.utc).replace(tzinfo=None) - timedelta(hours=24)
```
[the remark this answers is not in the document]
mitri-slory: You're right. I forgot that the DSpaceService already uses the datetime logic in its `download_records` method so I'l revert the logic to how it was originally.

processes/ingest/clacso.py
```diff
@@ -1,45 +1,113 @@
...
-            self.commitChanges()
+            if records:
+                for record_mapping in records:
+                    self.record_buffer.add(record_mapping.record)
+                    self.store_pdf_manifest(record_mapping.record)
 
-            logger.info(f'Ingested {len(records)} CLACSO records')
+            self.record_buffer.flush()
+
+            logger.info(f'Ingested {self.record_buffer.ingest_count if records else 0} CLACSO records')
```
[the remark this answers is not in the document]
mitri-slory: Oh that's a good catch. I'll get rid of the conditional.

processes/ingest/clacso.py
```diff
@@ -1,45 +1,113 @@
...
+                    url=manifest_uri,
+                    source=source,
+                    file_type='application/pdf',
+                    flags=json.dumps(dataclasses.asdict(FileFlags()))
+                ).to_string()
+                record.has_part.insert(0, link_string)
+                break
+
+    @staticmethod
+    def generate_manifest(record, source_uri, manifest_urI):
```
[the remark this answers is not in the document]
mitri-slory: Thanks for the catching the typo.

processes/ingest/clacso.py
```diff
@@ -1,45 +1,113 @@
...
+                manifest_json = self.generate_manifest(record, uri, manifest_uri)
+
+                self.s3_manager.createManifestInS3(manifest_path, manifest_json, self.s3_bucket)
+
+                link_string = Part(
+                    index=item_no, 
+                    url=manifest_uri,
+                    source=source,
+                    file_type='application/pdf',
+                    flags=json.dumps(dataclasses.asdict(FileFlags()))
```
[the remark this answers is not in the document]
jackiequach: Got it, thanks for the explanation!

services/sources/dspace_service.py
```diff
@@ -51,7 +51,10 @@ def get_records(self, full_import=False, start_timestamp=None, offset: Optional[
 
                 try:
                     parsed_record = self.parse_record(record)
-                    mapped_records.append(parsed_record)
+                    if parsed_record.record:
+                        mapped_records.append(parsed_record)
+                    else:
```
jackiequach, line 56: Sorry, missed this on the first go through. This else condition is not needed since it will always continue after the if statement is evaluated.

jackiequach (approved):
Looks good

services/sources/dspace_service.py
mitri-slory: I thought so as well until my CLACSO integration test kept getting an AssertionError for the len(records) due to the `parsed_record.record` being None and thus returning an mapped_record array. Adding this continue statement makes it so we only process the records that are fully mapped and check the conditional statement once the if conditional is fulfilled. The continue statement will start the next iteration of the for loop: `for record in oaidc_records.xpath('//oai_dc:dc', namespaces=self.OAI_NAMESPACES)`

PR #563, opened 2025-02-10
NO-REF: Update DOABProcess to use RecordBuffer

jackiequach:
# Description
Updates the DOABProcess to use RecordBuffer instead of inheriting from the CoreProcess class.

# Testing
Ran unit, integration, and functional tests

### DIFF
--- a/processes/ingest/doab.py
+++ b/processes/ingest/doab.py
@@ -66,6 +66,8 @@ class DOABProcess():
         except Exception as e:
             logger.exception('Failed to run DOAB process')
             raise e
+        finally:
+            self.db_manager.close_connection()
 
     def manage_links(self, record):
         linkManager = DOABLinkManager(record.record)
--- a/processes/ingest/doab.py
+++ b/processes/ingest/doab.py
@@ -1,19 +1,26 @@
 import os
 from services import DSpaceService
-from ..core import CoreProcess
 from logger import create_log
 from mappings.doab import DOABMapping
-from managers import DOABLinkManager, S3Manager, RabbitMQManager
+from managers import DBManager, DOABLinkManager, S3Manager, RabbitMQManager
 from model import get_file_message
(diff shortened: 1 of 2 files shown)


processes/ingest/doab.py
```diff
@@ -1,19 +1,26 @@
...
 logger = create_log(__name__)
 
-class DOABProcess(CoreProcess):
+class DOABProcess():
     DOAB_BASE_URL = 'https://directory.doabooks.org/oai/request?'
     DOAB_IDENTIFIER = 'oai:directory.doabooks.org'
 
     def __init__(self, *args):
-        super(DOABProcess, self).__init__(*args[:4])
+        self.process = args[0]
```
jackiequach, line 16: Not sure if this is the best way to get the arguments. Let me know if you know of a better option.

mitri-slory: For right now that is the best and easiest way to get the arguments.

processes/ingest/doab.py
```diff
@@ -34,35 +41,31 @@ def __init__(self, *args):
...
-            self.commitChanges()
-
-            logger.info(f'Ingested {len(self.records) if records else 1} DOAB records')
+            logger.info(f'Ingested {self.record_buffer.ingest_count} DOAB records')
 
         except Exception as e:
             logger.exception('Failed to run DOAB process')
             raise e
-        finally:
-            self.close_connection()
```
mitri-slory, line 65: I think we should keep this close_connection method to ensure the session is ended after the flush. What do you think?

mitri-slory (approved):
This looks great! I only have two comments but besides those this is good to go and looks similar to how I used RecordBuffer in the CLACSO process as well.

processes/ingest/doab.py
jackiequach: Good idea, re-added it

PR #565, opened 2025-02-12
UPDATE-SFR-2502/add_catalog_process

ayan1229:
this test does the following 
- adds catalog process 
- verifies if identifier is correctly updated after classifying process is run for the record

### DIFF
--- a/tests/functional/processes/frbr/test_frbr_process.py
+++ b/tests/functional/processes/frbr/test_frbr_process.py
@@ -18,6 +18,10 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
 
     db_manager.session.commit()
     db_manager.session.expire_all()
+    db_manager.session.refresh(frbrized_record)
+
+    assert any(id.endswith('|owi') for id in frbrized_record.identifiers), \
+        f"Missing OWI identifier after classification. Current IDs: {frbrized_record.identifiers}" #test is failing because indentifier field is not getting updated after classifying
 
     current_time = datetime.utcnow()
     time_window = current_time - timedelta(minutes=2)
--- a/tests/functional/processes/frbr/test_frbr_process.py
+++ b/tests/functional/processes/frbr/test_frbr_process.py
@@ -1,8 +1,9 @@
 from datetime import datetime, timedelta, timezone
-from processes import ClassifyProcess
+from processes import ClassifyProcess, CatalogProcess
+from managers import OCLCCatalogManager, RabbitMQManager
 from model import Record
 
-def test_frbr_process(db_manager, unfrbrized_record_uuid):
+def test_frbr_process(db_manager, unfrbrized_record_uuid, unfrbrized_title):
     classify_process = ClassifyProcess(
         'custom',
         None,
@@ -13,15 +14,11 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
     classify_process.runProcess()
 
     frbrized_record = db_manager.session.query(Record).filter(Record.uuid == unfrbrized_record_uuid).first()
-    
+        
     assert frbrized_record.frbr_status == 'complete'
 
     db_manager.session.commit()
     db_manager.session.expire_all()
-    db_manager.session.refresh(frbrized_record)
-
-    assert any(id.endswith('|owi') for id in frbrized_record.identifiers), \
-        f"Missing OWI identifier after classification. Current IDs: {frbrized_record.identifiers}" #test is failing because indentifier field is not getting updated after classifying
 
     current_time = datetime.utcnow()
     time_window = current_time - timedelta(minutes=2)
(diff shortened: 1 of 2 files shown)


tests/functional/processes/frbr/test_frbr_process.py
```diff
@@ -13,19 +14,65 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
...
     assert classify_count >= 1, f"Expected oclcClassify records. Found: {classify_count}"
+
+    new_record = (
+        db_manager.session.query(Record)
+        .filter(Record.title.ilike(f"%{unfrbrized_title}%"))
+        .order_by(Record.date_created.desc())
+        .first())
+    
+    if not new_record:
+        pytest.fail(f"No record found containing title: '{unfrbrized_title}'")
```
mitri-slory, line 37: Pytest needs to be imported like the other import you made at the top of the code.

tests/functional/processes/frbr/test_frbr_process.py
```diff
@@ -13,19 +14,65 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
...
+        Record.date_created >= datetime.now(timezone.utc) - timedelta(minutes=1)
+    ).all()
+
+    matching_records = [
+        record for record in new_catalog_records
+        if any(oclc_id in record.identifiers for oclc_id in oclc_identifiers)
+    ]
+    
+    assert len(matching_records) > 0, (
+        f"No catalog records matched OCLC identifiers: {oclc_identifiers}"
```
mitri-slory, line 75: oclc_identifiers is not declared as a variable before they're called here which is causing an error.

tests/functional/processes/frbr/test_frbr_process.py
```diff
@@ -25,3 +26,53 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
...
 
     assert classify_count >= 1, f"Expected oclcClassify records. Found: {classify_count}"
+
+    new_record = (
+        db_manager.session.query(Record)
+        .filter(Record.title.ilike(f"%{unfrbrized_title}%"))
+        .order_by(Record.date_created.desc())
+        .first())
+    
+    if not new_record:
```
kylevillegas93, line 36: you could assert that the record is not none here!

tests/functional/processes/frbr/test_frbr_process.py
```diff
@@ -25,3 +26,53 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
...
+        pytest.fail(f"No record found containing title: '{unfrbrized_title}'")
+
+    # Verify OWI in identifiers instead of source_id
+    assert any(id.endswith('|owi') for id in new_record.identifiers), \
```
kylevillegas93, line 40: We can check here that the source_id is an owi number.

tests/functional/processes/frbr/test_frbr_process.py
```diff
@@ -25,3 +26,53 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
...
+        f"Missing OWI identifier in: {new_record.identifiers}"
+
+    # Verify OCLC in identifiers
+    oclc_ids = [id.split('|')[0] for id in new_record.identifiers if id.endswith('|oclc')]
+    assert len(oclc_ids) > 0, f"No OCLC numbers found in: {new_record.identifiers}"
+
+    initial_catalog_count = db_manager.session.query(Record).filter(Record.source == 'oclcCatalog').count()
```
kylevillegas93, line 47: I think we can maybe streamline the check to make sure that the catalog process is working properly. Let's remove the oclcCatalog record count queries which will be quite big against QA.

tests/functional/processes/frbr/test_frbr_process.py
```diff
@@ -25,3 +26,53 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
...
+
+    catalog_process = CatalogProcess(
+        'catalog_test',
+        None,
+        datetime.now(timezone.utc) - timedelta(minutes=2),
+        None
+    )
```
kylevillegas93, line 54: let's just do catalog_process = CatalogProcess(None, None, None, None)

tests/functional/processes/frbr/test_frbr_process.py
```diff
@@ -25,3 +26,53 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
...
+
+    catalog_process.runProcess()
```
kylevillegas93, line 56: This will wait for awhile to process more messages as they come into the queue. Let's set the max attempts to 1 so it only checks once. 

`catalog_process.runProcess(max_attempts=1)`

tests/functional/processes/frbr/test_frbr_process.py
```diff
@@ -25,3 +26,53 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
...
+
+    final_catalog_count = db_manager.session.query(Record).filter(Record.source == 'oclcCatalog').count()
+
+    assert final_catalog_count > initial_catalog_count, (
+        f"Catalog records didn't increase. Before: {initial_catalog_count}, After: {final_catalog_count}"
+    )
```
kylevillegas93, line 62: We can remove this check.

tests/functional/processes/frbr/test_frbr_process.py
```diff
@@ -25,3 +26,53 @@ def test_frbr_process(db_manager, unfrbrized_record_uuid):
...
+        Record.date_created >= datetime.now(timezone.utc) - timedelta(minutes=1)
+    ).all()
+
+    matching_records = [
+        record for record in new_catalog_records
+        if any(oclc_id in record.identifiers for oclc_id in oclc_identifiers)
+    ]
+    
+    assert len(matching_records) > 0, (
+        f"No catalog records matched OCLC identifiers: {oclc_identifiers}"
```
kylevillegas93, line 75: I recommend getting oclcCatalog based on the oclc_ids that you pulled above. You may need to keep the identifier type, e.g. "1234|oclc". 

I think the final assertion is that we have added an oclcCatalog record for each oclc_id.

mappings/oclcCatalog.py
```diff
@@ -184,7 +189,7 @@ def parseLink(self, part):
 
         partDict = json.loads(partFlags)
 
-        if partDict['marcInd1'] != '4' or partLink == '':
+        if partDict.get('marcInd1') != '4' or partLink == '':
```
kylevillegas93, line 192: marcInd1 may not exist in the partDict so using get to prevent accessing null

mappings/oclcCatalog.py
```diff
@@ -167,7 +167,12 @@ def createMapping(self):
     def applyFormatting(self):
         self.record.source = 'oclcCatalog'
         self.record.identifiers[0] = self.remove_oclc_prefixes(self.record.identifiers[0])
-        self.record.source_id = self.record.identifiers[0]
+
+        oclc_number = int(self.record.identifiers[0].split('|')[0])
```
kylevillegas93, line 171: converting the oclc_number to an integer removes the leading zeroes

jackiequach (approved):
These changes look good!

kylevillegas93 (approved):

PR #566, opened 2025-02-18
SFR-2514: Add functional test for cluster process

jackiequach:
# Description
- Add functional test for the cluster process
- Add test data for the test

# Testing
`pytest tests/functional/processes/frbr/test_cluster_process.py`


ayan1229 (approved):

tests/conftest.py
```diff
@@ -125,13 +125,37 @@ def create_or_update_record(record_data):
         'date_modified': datetime.now(timezone.utc).replace(tzinfo=None)
     }
 
+    test_unclustered_record_data = {
```
kylevillegas93, line 128: we may want to consider at some point moving these test data to a separate file

PR #567, opened 2025-02-18
SFR-2515: Add Publisher to CLACSO Mapping

mitri-slory:
This PR adds the publisher metadata from DSpace to the CLACSO mapping to allow the ClusterProcess to create editions from the CLACSO records and works.

### DIFF
--- a/mappings/clacso.py
+++ b/mappings/clacso.py
@@ -5,7 +5,6 @@ from bs4 import BeautifulSoup
 from uuid import uuid4
 import json
 from typing import Optional
-
 from model import Record, FRBRStatus, FileFlags, Part, Source
 from .base_mapping import BaseMapping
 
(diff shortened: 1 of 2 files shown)


mappings/clacso.py
```diff
@@ -67,7 +69,7 @@ def _get_identifiers(self, record, namespaces):
     def _get_dates(self, record, namespaces):
         years_issued = [date.split('-')[0] if '-' in date else date for date in record.xpath('./dc:date/text()', namespaces=namespaces)]
         
-        return [f'{min(years_issued)}|issued']
+        return [f'{min(years_issued)}|publisher_date']
```
kylevillegas93, line 72: should this be publication_date or publisher_date?

mitri-slory: It should be publication_date.

kylevillegas93 (approved):

PR #570, opened 2025-02-19
Add concurrency in GHA

ayan1229:
This PR does the following

-     adds concurrency in ci/cd on workflow level
-     ensures one instance of the pipeline can run at a time for the main branch
-     new pushes will cancel any in-progress pipeline runs
-     production deployments will complete before new ones start
-     avoids race conditions in deployment processes

### DIFF
--- a/.github/workflows/full-ci-cd.yaml
+++ b/.github/workflows/full-ci-cd.yaml
@@ -4,6 +4,10 @@ on:
   push:
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   deploy_to_qa:
     uses: NYPL/drb-etl-pipeline/.github/workflows/build-qa.yaml@main
--- a/.github/workflows/full-ci-cd.yaml
+++ b/.github/workflows/full-ci-cd.yaml
@@ -5,8 +5,8 @@ on:
     branches: [main]
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/full-ci-cd.yaml
```diff
@@ -4,6 +4,10 @@ on:
   push:
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
kylevillegas93, line 8: I think `github.ref` references the commit SHA, right? So the group will change on every merge. 

From the docs:

> The concurrency key is used to group workflows or jobs together into a concurrency group. When you define a concurrency key, GitHub Actions ensures that only one workflow or job with that key runs at any given time. If a new workflow run or job starts with the same concurrency key, GitHub Actions will cancel any workflow or job already running with that key. The concurrency key can be a hard-coded string, or it can be a dynamic expression that includes context variables.

I think for our case the group should be hardcoded, e.g. continuous-deployment.

.github/workflows/full-ci-cd.yaml
```diff
@@ -4,6 +4,10 @@ on:
...
+  cancel-in-progress: true
```
kylevillegas93, line 9: Also for our case, we don't want to cancel in progress ci/cd workflows. We want to let any previous currently running action to finish before the next merge kicks off a ci/cd. 

Does that make sense?

kylevillegas93 (approved):

PR #571, opened 2025-02-20
SFR-2518: Map languages and subjects in CLACSO mapping

kylevillegas93:
# Description
- Fixes limit/offset bug where we skipped the first 100 records
- Adds languages and subjects in CLACSO mapping

## Testing
`make functional`


mitri-slory (approved):
This looks great!

PR #574, opened 2025-02-20
NO-REF: Add book medium in Spanish and Portuguese

mitri-slory:
This hotfix adds the Spanish and Portuguese word for 'book' as one of the possible medium types that a CLACSO record mapping should have.


kylevillegas93 (approved):
Looks good!

PR #586, opened 2025-02-26
SFR-2537 Pointing the QA stack to the new QA database

kylevillegas93:
## Description
- https://newyorkpubliclibrary.atlassian.net/browse/DOPS-1112 Devops needs to upgrade the postgres version to 16 since 12 is no longer supported. 
- We decided it was time to create a separate database cluster altogether for QA 
- This change points the QA stack to the new database cluster via modifying the config. 
- Once we have tested that everything looks good in QA we can upgrade the production database cluster version. 

## Local Dev Changes
- The postgres volume's files will no longer be compatible with the postgres docker container
- Each developer must delete their volume and rerun the local developer setup process to reinitialize the database. 

## Testing 
`python main.py -p CLACSOProcess -e local-qa -i complete -l 1`

### DIFF
--- a/managers/oclc_catalog.py
+++ b/managers/oclc_catalog.py
@@ -17,7 +17,6 @@ class OCLCCatalogManager:
     LIMIT = 50
     MAX_NUMBER_OF_RECORDS = 100
     BEST_MATCH = 'bestMatch'
-    OK_STATUS_CODES = [200, 206]
 
     def query_catalog(self, oclc_no):
         catalog_query = self.METADATA_BIB_URL.format(oclc_no)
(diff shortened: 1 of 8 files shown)


mitri-slory (approved):

kylevillegas93:
@mitri-slory @ayan1229 @jackiequach - heads up that tests seemed to be failing sometimes so fixed them in this PR. 

This change will also require y'all to reinitialize your local database after the upgrade to postgres 16.

managers/oclc_catalog.py
```diff
@@ -102,7 +103,7 @@ def _get_other_editions(self, oclc_number: int, offset: int=0):
 
             status_code = other_editions_response.status_code
 
-            if other_editions_response.status_code != 200:
+            if other_editions_response.status_code not in self.OK_STATUS_CODES:
```
jackiequach, line 106: is there a reason why we shouldn't just check `other_editions_response.ok`?

kylevillegas93: you make a good point!

mitri-slory (approved):

mitri-slory:
> @mitri-slory @ayan1229 @jackiequach - heads up that tests seemed to be failing sometimes so fixed them in this PR.
> 
> This change will also require y'all to reinitialize your local database after the upgrade to postgres 16.

Thanks for the heads up.

jackiequach (approved):
Looks good!

PR #588, opened 2025-02-26
SFR-2533: Update oclc cat api param to narrow results

jackiequach:
# Description

Updates the OCLC catalog API params to narrow down results only to include digital books.

Reasoning for replacing `itemType` with `itemSubType`:

When both `itemType` and `itemSubType` parameters are provided in the query, the `itemType` will take precedence and limit the results based on the top-level facet type (book). As a result, the `itemSubType` will not affect the results, as it is overridden by the `itemType`. The `itemSubType` of 'book-digital' refines the results within the 'book' category.

This reduced the results of the API call in the `test_frbr_process.py` catalog process test from 8 records to 3.

# Testing 
make functional


kylevillegas93 (approved):
Looks good!

PR #590, opened 2025-02-26
SFR-2504: Refactor manifest methods in ingest processes

mitri-slory:
This PR refactors the generateManifest and createManifestInS3 methods in many of the ingest processes by making the method names camel case and moving the generateManifest method to the S3Manager class.


managers/muse.py
```diff
@@ -151,7 +151,7 @@ def constructS3Link(self, bucketLocation):
             self.s3Bucket, bucketLocation
         )
 
-    def createManifestInS3(self, bucketLocation):
+    def create_manifest_in_s3(self, bucketLocation):
         s3URL = self.constructS3Link(bucketLocation)
```
kylevillegas93, line 155: I think we may want to start using `get_stored_file_url` in the digital_assets/utils folder. 

This'll make it work for your local env, too.

managers/muse.py
```diff
@@ -93,7 +93,7 @@ def addReadableLinks(self):
         if self.pdfWebpubManifest:
             self.s3PDFReadPath = 'manifests/muse/{}.json'.format(self.museID)
             self.record.add_has_part_link(
-                self.createManifestInS3(self.s3PDFReadPath),
+                self.create_manifest_in_s3(self.s3PDFReadPath),
```
kylevillegas93, line 96: Are there other ingest processes that have the exact same code to create the manifest?

mitri-slory: Yeah CLACSO, MET, DOAB, etc. have the same code for creating the manifest which is why I updated the name of the method in all of those processes in my commits.

managers/muse.py
mitri-slory: Ok I'll add that method to other ingest processes.

processes/ingest/loc.py
```diff
@@ -215,28 +216,26 @@ def addHasPartMapping(self, resultsRecord, record):
 
     def storePDFManifest(self, record):
```
kylevillegas93, line 217: What are your thoughts on pulling this function out? I think this is duplicate code in many of the ingest processes.

mitri-slory: I can pull this function out and I'll refactor it in my next commit.

kylevillegas93: ok sounds good!

managers/s3.py
```diff
@@ -2,10 +2,15 @@
...
 import hashlib
+import json
+import dataclasses
 from io import BytesIO
 import mimetypes
 import os
 from zipfile import ZipFile
+from managers import WebpubManifest
+from digital_assets import get_stored_file_url
+from model import Part, FileFlags\
```
kylevillegas93, line 13: typo here i think with the ending backslash

managers/s3.py
```diff
@@ -37,11 +42,79 @@ def createS3Bucket(self, bucketName, bucketPermissions):
         raise S3Error('Unable to create bucket in s3')
 
 
-    def createManifestInS3(self, manifestPath, manifestJSON, s3_bucket: str):
+    def store_pdf_manifest(self, record, bucket_name, requires_login: bool):
```
kylevillegas93, line 45: I think we can make this function body definitely work with the other sources. Is the plan to make those changes in another commit?

managers/s3.py
mitri-slory: Nice catch

managers/s3.py
mitri-slory: This method works with the other sources already. I only added some conditionals for the clacso and pub backlist works but the else statement at line 86 applies to the other sources that use this method.

kylevillegas93: Ok - let's try to simplify! I recommend using the met.py ingest process store_pdf_manifest as the foundation: 

```
record_id = record.source_id.split('|')[0]
        pdf_part = next(filter(lambda part: part.file_type == 'application/pdf', record.get_parts()), None)

        if pdf_part is not None:
            manifest_path = f'manifests/{record.source}/{record_id}'
            manifest_url = get_stored_file_url(storage_name=self.s3_bucket, file_path=manifest_path)
            manifest_json = self.generate_manifest(record=record, source_uri=pdf_part.url, manifest_uri=manifest_url)

            self.s3_manager.createManifestInS3(manifestPath=manifest_path, manifestJSON=manifest_json, s3_bucket=self.s3_bucket)

            record.has_part.insert(0, Part(
                index=pdf_part.index,
                url=manifest_url,
                source=record.source,
                file_type='application/webpub+json',
                flags=pdf_part.flags
            ).to_string())
```

managers/s3.py
```diff
@@ -37,11 +42,79 @@ def createS3Bucket(self, bucketName, bucketPermissions):
...
+        for link in record.has_part:
+            item_no, uri, source, media_type, _ = link.split('|')
+
+            if media_type == 'application/pdf':
+                record_id = record.identifiers[0].split('|')[0]
+
+                if requires_login:
+                    manifest_path = f'manifests/publisher_backlist/{source}/{record_id}.json'
```
kylevillegas93, line 53: We should generalize this logic for other sources. We could consider passing in an optional prefix.

managers/s3.py
```diff
@@ -37,11 +42,79 @@ def createS3Bucket(self, bucketName, bucketPermissions):
...
+                else:
+                    manifest_path = f'manifests/{source}/{record_id}.json''manifests/{}/{}.json'
```
kylevillegas93, line 55: There might be a typo here.

managers/s3.py
```diff
@@ -37,11 +42,79 @@ def createS3Bucket(self, bucketName, bucketPermissions):
...
+                self.create_manifest_in_s3(manifest_path, manifest_json, bucket_name)
+
+                if source == 'clacso':
+                    link_string = Part(
+                        index=item_no, 
+                        url=manifest_uri,
+                        source=source,
+                        file_type='application/webpub+json',
+                        flags=json.dumps(dataclasses.asdict(FileFlags()))
+                    ).to_string()
```
kylevillegas93, line 70: I think we should try to not have any source specific logic in this function.

managers/s3.py
```diff
@@ -37,11 +42,79 @@ def createS3Bucket(self, bucketName, bucketPermissions):
...
+                elif requires_login:
+                    manifest_flags = {
+                    'catalog': False,
+                    'download': False,
+                    'reader': True,
+                    'embed': False,
+                    **({'fulfill_limited_access': False} if requires_login else {})
+                    }
```
kylevillegas93, line 78: I think we could just add a fullfill_limited_access field to the FileFlags data class and set it in the FileFlags.

managers/s3.py
kylevillegas93, line 45: We can default to `requires_login: bool=False` so the other ingest processes don't have to set this parameter.

managers/s3.py
mitri-slory: What is your suggestion to keep CLACSO books from being available on the web reader if we don't want source specific logic with the flags all being false in this function?

managers/s3.py
mitri-slory: That makes sense

managers/s3.py
kylevillegas93: Also for CLACSO, if it's not readable, do we need the manifest?

mitri-slory: I'll pass the flags as a parameter in my next commit.  Also, we don't need the manifest for non-readble books from talking to Jackie since they should only be used for the web reader and the download comes from item links found by the item download flag. We will need these manifests when we decide to add the read online button back for the CLACSO books so I suggest we keep them and then update the item download flags for them. What are your thoughts?

managers/s3.py
```diff
@@ -37,11 +42,49 @@ def createS3Bucket(self, bucketName, bucketPermissions):
         raise S3Error('Unable to create bucket in s3')
 
 
-    def createManifestInS3(self, manifestPath, manifestJSON, s3_bucket: str):
+    def store_pdf_manifest(self, record, bucket_name, flags=None, requires_login: bool=False):
```
kylevillegas93, line 45: let's add some types to the params to make sure they are clear. 

Instead of the flags defaulting to None, we can default to FileFlags(reader=True)

managers/s3.py
```diff
@@ -37,11 +42,49 @@ def createS3Bucket(self, bucketName, bucketPermissions):
...
+            self.create_manifest_in_s3(manifest_path=manifest_path, manifest_json=manifest_json, s3_bucket=bucket_name)
+
+            if flags:
+                record.has_part.insert(0, Part(
+                    index=pdf_part.index,
+                    url=manifest_url,
+                    source=record.source,
+                    file_type='application/webpub+json',
+                    flags=flags
+                ).to_string())
```
kylevillegas93, line 66: I think my question for CLACSO is if the PDF is not readable, do we need to store the manifest until it is readable? In other words, do we need to call this function altogether? 

This would mean that flags should not be None and we can remove this if.

processes/ingest/met.py
```diff
@@ -99,38 +101,3 @@ def get_cover_path(self, file_type: Optional[str], record_id: str) -> Optional[s
...
-            manifest_json = self.generate_manifest(record=record, source_uri=pdf_part.url, manifest_uri=manifest_url)
-
-            self.s3_manager.createManifestInS3(manifestPath=manifest_path, manifestJSON=manifest_json, s3_bucket=self.s3_bucket)
-
-            record.has_part.insert(0, Part(
-                index=pdf_part.index,
-                url=manifest_url,
-                source=record.source,
-                file_type='application/webpub+json',
-                flags=pdf_part.flags
```
kylevillegas93, line 119: Interestingly, the manifest flags for MET were that of the PDF flags which were download only as True. 

Should the PDF manifest only have the reader flag as True in all cases?

managers/s3.py
kylevillegas93: See my comment below as well - we could perhaps just do FileFlags(reader=True) when adding the manifest has_part.

managers/s3.py
```diff
@@ -37,11 +42,49 @@ def createS3Bucket(self, bucketName, bucketPermissions):
...
-    def createManifestInS3(self, manifestPath, manifestJSON, s3_bucket: str):
+    def store_pdf_manifest(self, record, bucket_name, flags=None, requires_login: bool=False):
+        record_id = record.source_id.split('|')[0]
+        pdf_part = next(filter(lambda part: part.file_type == 'application/pdf', record.get_parts()), None)
+
+        if pdf_part is not None:
+            if requires_login:
+                manifest_path = f'manifests/publisher_backlist/{record.source}/{record_id}.json'
+            else:
+                manifest_path = f'manifests/{record.source}/{record_id}'
```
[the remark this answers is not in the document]
mitri-slory: That makes sense and I'll do that instead of using requires_login

managers/s3.py
mitri-slory: We don't need to call the store_pdf_manifest for CLACSO in its current state since we don't want these books to be readable currently. We already have CLACSO manifests in the production bucket currently but I can remove the store_pdf_manifest function from CLACSO for now and I'll remove the flag conditional. What are your thoughts?

processes/ingest/met.py
mitri-slory: The PDF manifest reader flag must always be true in all cases to allow the manifest to be accessed by the web reader.

managers/s3.py
kylevillegas93: That makes sense - we can always reingest these records, too.

managers/s3.py
```diff
@@ -37,11 +42,49 @@ def createS3Bucket(self, bucketName, bucketPermissions):
...
+            self.create_manifest_in_s3(manifest_path=manifest_path, manifest_json=manifest_json, s3_bucket=bucket_name)
+
+            print(json.dumps(dataclasses.asdict(flags)))
+            record.has_part.insert(0, Part(
+                index=pdf_part.index,
+                url=manifest_url,
+                source=record.source,
+                file_type='application/webpub+json',
+                flags=json.dumps(dataclasses.asdict(flags))
+            ).to_string())
```
kylevillegas93, line 66: Nice - it's looking great!

Let's just remove the print statements. 

You could simplify the path prefix param and just pass in if there's an additional path prefix after the manifests folder. This would ensure all manifests are always stored in the manifests folder in S3.

kylevillegas93 (commented):
Looking good - thanks for making this change!

managers/s3.py
mitri-slory: Ok that makes sense with the path prefix

kylevillegas93 (approved):
Great work!

PR #591, opened 2025-02-27
SFR-2538/parallelization_tests)

ayan1229:
- runs tests in parallel using matrix strategy

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        groups: [1, 2, 3]
+        groups: [1, 2]
     env:
       IS_CI: "true"
       AWS_ACCESS: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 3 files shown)


kylevillegas93:
The matrix strategy could get tricky but also doesn't speed up the tests when we run them locally on our machines. You may want to see how pytest-parallel https://pypi.org/project/pytest-parallel/ works or the like. Try it locally and see what happens!

.github/workflows/tests.yaml
```diff
@@ -4,69 +4,130 @@ on:
...
-      - name: Wait for docker containers to be ready
-        run: |
-          echo "Waiting for docker container..."
-          sleep 90
-      - name: Run functional tests
+          pip install pytest-xdist
+      - name: Run Unit Tests
+        run: pytest tests/unit -n 4 -v
+
+  functional-tests:
```
kylevillegas93, line 62: I recommend not having separate jobs for these. While it's nice to be able to run them at the same time, it mean we have to checkout, cache, and install the dependencies all over again.

PR #597, opened 2025-02-28
SFR-2524: Update DOAB mapping to return single right

jackiequach:
- Return single right instead of list of rights


kylevillegas93 (approved):
looks good !

PR #598, opened 2025-02-28
SFR-2526: Refactor MET ingest and mapping

kylevillegas93:
## Description
- Refactors MET Process
- Uses service architecture
- Refactors MET mapping to be more functional 

## Testing
- Ingested all MET records


services/sources/met_service.py
```diff
@@ -0,0 +1,78 @@
...
+                    yield mapped_met_record
+
+            if limit and current_position >= limit:
+                break
+
+            current_position += page_size
+
+    def _get_met_records(self, page_size: int=50, current_position: int=0) -> list:
+        try:
+            records_response = METService.query_met_api(query=self.LIST_QUERY.format(page_size, current_position))
```
jackiequach, line 52: this can be self.query_met_api

services/sources/met_service.py
```diff
@@ -0,0 +1,78 @@
...
+
+            return records_response.get('records', [])     
+        except Exception:
+            logger.exception('Faile to get met records')   
+            return []
+
+    def _map_met_record(self, record: dict) -> Optional[METMapping]:
+        try:            
+            met_record = METService.query_met_api(query=self.ITEM_INFO_QUERY.format(record.get('pointer')))
```
jackiequach, line 61: same as above

services/sources/met_service.py
```diff
@@ -0,0 +1,78 @@
...
+
+            return METMapping(met_record)
+        except Exception:
+            logger.exception('Failed to process MET record')
+            return None
+
+    def query_met_api(query: str, method: str='GET') -> Union[str, dict]:
```
jackiequach, line 68: add _ at the beginning to denote internal use

tests/integration/services/sources/test_met_service.py
```diff
@@ -0,0 +1,11 @@
+
+from services import METService
+
+
+def test_get_records():
+    gutenberg_service = METService()
```
jackiequach, line 6: typo: met_service

jackiequach (commented):
Looking good, just a few minor comments

services/sources/met_service.py
kylevillegas93: This function is also called by the met ingest but I've moved it up so the public/internal functions are grouped together.

services/sources/met_service.py
```diff
@@ -26,7 +26,7 @@ def get_records(
         offset: Optional[int]=None,
         limit: Optional[int]=None
     ) -> Generator[METMapping, None, None]:
-        current_position = offset
+        current_position = offset if offset is not None else 0
```
jackiequach, line 29: Can the parameter definition be updated to 0 instead of None to eliminate the need for this logic?

kylevillegas93: yes - good callout!

PR #600, opened 2025-03-03
SFR-2537: Migrating to new QA cluster

kylevillegas93:
## Description
- We are going to try to migrate to the new QA cluster today
- Adding some logging to see if we are closing db sessions properly when querying for collections

## Testing
- Started the API against the QA cluster locally - looks good!


jackiequach (approved):
Looks good!

PR #604, opened 2025-03-06
NO-REF: Skeleton Record Pipeline Process

kylevillegas93:
## Description
- This is just a skeleton for the record pipeline process
- I think I want to flush out the details next sprint and complete the implementation 
- Putting this PR up tho for feedback!

### DIFF
--- a/processes/record_pipeline.py
+++ b/processes/record_pipeline.py
@@ -1,0 +1,34 @@
+
+
+'''
+High-Level Flow:
+
+1. Ingestors get and map records and send message to record pipeline queue
+2. Record pipeline process processes messages from the queue
+3. For each message, the record pipeline stores the files, frbrizes and clusters the record
+4. Optionally, the process fulfills the links
+
+Pros: 
+- Scalability: We can spin up n tasks to process messages from the queue. 
+- Simplicity: Removes potentially 4 processes and ECS tasks
+- Testability: Possible to test 
+- Observability/Debuggability: Easier to see where a record fails to be processed in the pipeline
+- State Management: We can ensure the database state fields truthfully match the state of a record. 
+'''
+
+class RecordPipelineProcess:
+
+    def __init__(self):
+        pass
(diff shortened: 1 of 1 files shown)


processes/record_pipeline.py
```diff
@@ -0,0 +1,34 @@
...
+3. For each message, the record pipeline stores the files, frbrizes and clusters the record
+4. Optionally, the process fulfills the links
+
+Pros: 
+- Scalability: We can spin up n tasks to process messages from the queue. 
+- Simplicity: Removes potentially 4 processes and ECS tasks
+- Testability: Possible to test 
+- Observability/Debuggability: Easier to see where a record fails to be processed in the pipeline
+- State Management: We can ensure the database state fields truthfully match the state of a record. 
+'''
```
kylevillegas93, line 17: @jackiequach @mitri-slory 

Do y'all see any other pros or cons?

mitri-slory: This is an interesting pipeline to create. So with this pipeline the ingest processes, classify, catalog, and cluster process will be allocated all in this one pipeline? One con I could potentially see is the clarity of the frbr and cluster processes since it might be confusing to see how they all work being in just one process.

mitri-slory (approved):
This skeleton looks good so far and I support the idea of creating a pipeline for the record process.

processes/record_pipeline.py
kylevillegas93: Good callout - we'd probably move those FRBR and cluster functions as a service instead of a process. The code may look like:

```
...
self.frbr_service.frbrize_record(record)
self.cluster_service.cluster_record(record)
...
```

jackiequach: A concern I have is the rate limiting by OCLC that may cause a bottleneck in this.

jackiequach (approved):
I think the overview for this looks good!

PR #606, opened 2025-03-06
SFR-2524: Add remove functionality to RecordBuffer and DOAB process

jackiequach:
# Description
- Adds delete functionality to RecordBuffer (this can be improved to also delete the OCLC records if the frbr process was run)
- Adds check for DOAB record to remove deleted records 

# Testing
`make integration` and `make functional`


kylevillegas93:
Overall - it's looking great! I will take a bigger pass tomorrow. Feel free to cruise on to MUSE!

mitri-slory (approved):

mappings/doab.py
```diff
@@ -14,7 +15,27 @@ class DOABMapping(BaseMapping):
...
     
-    def map_doab_record(self, doab_record, namespaces) -> Record:
+    def map_doab_record(self, record, namespaces) -> Record:
+        header = record.find('.//header', namespaces=DSpaceService.ROOT_NAMESPACE)
+        deletion_flag = header is not None and header.get('status') == 'deleted'
+
+        if header is not None:
+            identifier = header.find('.//identifier', namespaces=DSpaceService.ROOT_NAMESPACE)
+            if identifier is not None:
+                source_id = identifier.text.split(':')[-1]
```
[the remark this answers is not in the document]
kylevillegas93: ah I see - you are using the header to get the source_id. I recommend pulling this out into a separate function. Will the header exist for non-deleted records?

mappings/doab.py
```diff
@@ -14,7 +15,27 @@ class DOABMapping(BaseMapping):
...
+
+        if deletion_flag and source_id:
+            delete_record=Record(
```
kylevillegas93, line 28: Let's add a space around the equals sign. 

I think we can maybe just do `Record(source_id=source_id)` since the other values don't necessarily matter!

processes/ingest/doab.py
```diff
@@ -82,3 +83,15 @@ def manage_links(self, record):
         for epubLink in linkManager.ePubLinks:
             ePubPath, ePubURI = epubLink
             self.rabbitmq_manager.sendMessageToQueue(self.file_queue, self.file_route, get_file_message(ePubURI, ePubPath))
+
+    def _process_record(self, record):
```
kylevillegas93, line 87: recommend adding a type for the param here and perhaps call it record_mapping so it's not record.record

processes/record_buffer.py
```diff
@@ -24,6 +25,16 @@ def add(self, record: Record):
         if len(self.records) > self.batch_size:
             self.flush()
 
+    def delete(self, record: Record):
```
kylevillegas93, line 28: We may want to add a TODO to delete the rest of the FRBR model.

mappings/doab.py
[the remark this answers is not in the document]
jackiequach: Yes the header exists for non-deleted records. Updated so the source_id is only retrieved from the header.

kylevillegas93 (approved):
Looks good!

PR #608, opened 2025-03-06
NO-REF: Removing CoreProcess from PublisherBacklistProcess

kylevillegas93:
## Description
- Removes the CoreProcess
- Uses record buffer
- Removes full import field and instead relies on start_datetime!

## Testing 
`make unit`

### DIFF
--- a/processes/ingest/publisher_backlist.py
+++ b/processes/ingest/publisher_backlist.py
@@ -42,7 +42,7 @@ class PublisherBacklistProcess():
             
             self.record_buffer.flush()
 
-            logger.info(f'Ingested {len(self.record_buffer.ingest_count)} Publisher Backlist records')
+            logger.info(f'Ingested {self.record_buffer.ingest_count} Publisher Backlist records')
         except Exception as e:
             logger.exception('Failed to run Publisher Backlist process')
             raise e   
--- a/processes/ingest/publisher_backlist.py
+++ b/processes/ingest/publisher_backlist.py
@@ -1,18 +1,25 @@
 import os
 from services import PublisherBacklistService
 
-from ..core import CoreProcess
 from logger import create_log
-from managers import S3Manager
+from managers import DBManager, S3Manager
+from ..record_buffer import RecordBuffer
+from .. import utils
 
 logger = create_log(__name__)
 
-class PublisherBacklistProcess(CoreProcess):
+class PublisherBacklistProcess():
     def __init__(self, *args):
-        super(PublisherBacklistProcess, self).__init__(*args[:4])
+        self.process = args[1]
+        self.ingest_period = args[2]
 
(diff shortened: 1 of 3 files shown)


processes/ingest/publisher_backlist.py
```diff
@@ -1,50 +1,50 @@
...
+from ..record_buffer import RecordBuffer
+from .. import utils
 
 logger = create_log(__name__)
 
-class PublisherBacklistProcess(CoreProcess):
+class PublisherBacklistProcess():
     def __init__(self, *args):
-        super(PublisherBacklistProcess, self).__init__(*args[:4])
+        self.process = args[1]
```
jackiequach, line 13: this should be args[0]

processes/ingest/publisher_backlist.py
```diff
@@ -1,50 +1,50 @@
...
+            )
             
-            self.saveRecords()
-            self.commitChanges()
+            for record_mapping in records:
+                self.record_buffer.add(record_mapping.record)
+            
+            self.record_buffer.flush()
 
             logger.info(f'Ingested {len(self.records)} Publisher Backlist records')
```
jackiequach, line 45: This can be updated to use record_buffer.ingest_count

jackiequach (commented):
some minor comments

processes/ingest/publisher_backlist.py
kylevillegas93: oops lol

processes/ingest/publisher_backlist.py
kylevillegas93: oops !

processes/ingest/publisher_backlist.py
kylevillegas93: Good catch - I've been thinking about creating a utils function to parse the args

jackiequach: I think that's a good idea!

processes/ingest/publisher_backlist.py
```diff
@@ -1,50 +1,50 @@
...
-            self.saveRecords()
-            self.commitChanges()
-
-            logger.info(f'Ingested {len(self.records)} Publisher Backlist records')
+            for record_mapping in records:
+                self.record_buffer.add(record_mapping.record)
+            
+            self.record_buffer.flush()
 
+            logger.info(f'Ingested {len(self.record_buffer.ingest_count)} Publisher Backlist records')
```
jackiequach, line 45: don't need len() here

kylevillegas93: Soon my hope is that there's just 1 ingest process and we can remove all of this duplicate code :)

kylevillegas93:
gutenberg is probably failing because of the request limit - ill rerun the tests later

jackiequach (approved):

PR #611, opened 2025-03-06
NO-REF: Use start datetime logic in Clacso ingest

kylevillegas93:
## Description
- This change uses the start datetime getter util function to simplify the CLACSO ingest process

## Testing
`make unit`


jackiequach (approved):
Looks good

PR #612, opened 2025-03-07
NO-REF: Adding parse process args utils function

kylevillegas93:
## Description
- Adding parse process args utils function
- This should remove redundant code to parse args in the processes

## Testing
`make unit`

### DIFF
--- a/processes/utils.py
+++ b/processes/utils.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


mitri-slory (approved):
Really good change that helps make it more clear what each index of the parse process args represent.

processes/utils.py
```diff
@@ -1,6 +1,8 @@
+from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from typing import Optional
 
+
```
jackiequach, line 5: Just curious, is it the standard to have 2 line breaks for spacing?

jackiequach (approved):
Looks good!

processes/utils.py
kylevillegas93: Yes! 

https://peps.python.org/pep-0008/#:~:text=Surround%20top%2Dlevel%20function%20and,separate%20groups%20of%20related%20functions.

PR #613, opened 2025-03-07
SFR-2541: python dependencies and docker caching for CI time improvements

ayan1229:
this or does the following 

1. updates tests.yaml workflow file to cache python dependencies and docker


.github/workflows/tests.yaml
```diff
@@ -4,18 +4,18 @@ on:
...
+      - "mappings/**"
+      - "model/**"
+      - "processes/**"
+      - "services/**"
+      - "tests/**"
+      - "utils/**"
+      - "dev-requirements.txt"
+      - "requirements.txt"
+      - "main.py"
+      - "Makefile"
```
kylevillegas93, line 18: let's keep single quotes!

.github/workflows/tests.yaml
```diff
@@ -25,48 +25,51 @@ jobs:
...
-      - name: Install dependencies
+
+      - name: Install Python dependencies
         run: |
-          python -m pip install --upgrade pip wheel
-          pip install -r dev-requirements.txt
-          pip install -r requirements.txt
-      - name: Run unit tests
+          python -m pip install --upgrade pip
+          pip install --user -r requirements.txt -r dev-requirements.txt
```
kylevillegas93, line 53: do we need to install use --user so that the cache picks up the deps in the home directory?

.github/workflows/tests.yaml
```diff
@@ -25,48 +25,51 @@ jobs:
...
-        run: |
-          echo "Waiting for docker container..."
-          sleep 90
-      - name: Run functional tests
+
+      - name: Wait for containers
         run: |
-          make functional
-      - name: Run integration tests
+          echo "Waiting for services..." && sleep 90
```
kylevillegas93, line 66: im curious if we can wait for less time here but no need to check for this PR!

kylevillegas93 (changes requested):
Looks great! Nice work - just a couple minor comments and questions!

.github/workflows/tests.yaml
ayan1229: we don't need to. 
Without --user: Packages go to venv path. Cache should work fine

.github/workflows/tests.yaml
```diff
@@ -4,18 +4,18 @@ on:
...
-        - 'main.py'
-        - 'Makefile' 
+      - 'api/**'
+      - 'managers/**'
+      - 'mappings/**'
+      - 'model/**'
+      - 'processes/**'
+      - 'services/**'
+      - 'tests/**'
+      - "utils/**"
```
kylevillegas93, line 14: oops - missed one here!

kylevillegas93 (approved):
Looks good!

PR #615, opened 2025-03-11
NO-REF: Make new additions more pythonic

kylevillegas93:
## Description
- Coming from a predominantly java background, I'm still thinking like a java developer but I need to think like a python developer
- Uses the __str__ method to convert Part and FileFlags to a string - while this is a bit out of step with the purpose of this special method, I think it should suffice. 
- Uses @property to get the parts from a Record

## Testing
`make functional`


jackiequach (approved):
Good improvements!

ayan1229 (approved):

PR #616, opened 2025-03-11
NO-REF: Building alternative DRB ETL pipeline

kylevillegas93:
## Description
- Builds the alternative DRB ETL pipeline
- Creates 1 ingest process which will ingest from a single source

## Testing
`make functional`

## Performance Improvements
- Reduces the amount of time to ingest Chicago ISAC records and files from 15 seconds to 8 seconds


processes/record_ingestor.py
```diff
@@ -0,0 +1,113 @@
...
+            rabbitmq_manager.createRabbitConnection()
+            rabbitmq_manager.createOrConnectQueue(record_pipeline_queue, record_pipeline_route)
+
+            rabbitmq_manager.sendMessageToQueue(
+                queueName=record_pipeline_queue,
+                routingKey=record_pipeline_route,
+                message=json.dumps({ 'recordId': record_id })
+            )
+        finally:
+            rabbitmq_manager.closeRabbitConnection()
```
mitri-slory, line 90: This is interesting to utilize RabbitMQ to send a record pipeline message instead of mainly using the service for the FRBRization processes. What's the purpose of having this pipeline message in this alternative DRB ETL pipeline and what will it improve in comparison to our current ETL pipeline?

processes/ingest/chicago_isac.py
```diff
@@ -1,50 +1,14 @@
...
 
-from mappings.chicago_isac import map_chicago_isac_record
-from managers import DBManager, S3Manager, WebpubManifest
-from model import Record, Part, FileFlags
-from logger import create_log
-from ..record_buffer import RecordBuffer
-from digital_assets import get_stored_file_url
 
-logger = create_log(__name__)
+class ChicagoISACProcess(RecordIngestor):
```
sarangj, line 6: Curious if you'd considered composition instead to reduce coupling? Like, the `ChicagoISACProcess` could *have* a record ingestor to do the ingest, that way if an ingest needs to do something weird with params or something, could be simpler. That said, the inheritance looks like it works well too!

processes/record_ingestor.py
```diff
@@ -0,0 +1,113 @@
...
+            logger.exception(f'Failed to ingest record: {record}')
+            return False
+        
+    @staticmethod
+    def _save_record(record: Record):
+        try:
+            db_manager = DBManager()
+
+            with Session(db_manager.generateEngine()) as session, session.begin():
+                existing_record = session.query(Record).filter(Record.source_id == record.source_id).first()
```
sarangj, line 65: Do you need to filter on `source` here as well, say if two records from different sources have the same source id?

processes/record_ingestor.py
```diff
@@ -0,0 +1,113 @@
...
+
+            return True
+        except Exception:
+            logger.exception(f'Failed to ingest record: {record}')
+            return False
+        
+    @staticmethod
+    def _save_record(record: Record):
+        try:
+            db_manager = DBManager()
```
sarangj, line 62: would it make sense to reuse DBManager instances across calls to this method, either by passing it in, or by making this method not static and attaching the manager to `self`?

processes/record_ingestor.py
```diff
@@ -0,0 +1,113 @@
...
+        
+    @staticmethod
+    def _send_record_pipeline_message(record_id: int):
+        try:
+            record_pipeline_queue = os.environ.get('RECORD_PIPELINE_QUEUE')
+            record_pipeline_route = os.environ.get('RECORD_PIPELINE_ROUTING_KEY')
+
+            rabbitmq_manager = RabbitMQManager()
+            rabbitmq_manager.createRabbitConnection()
+            rabbitmq_manager.createOrConnectQueue(record_pipeline_queue, record_pipeline_route)
```
sarangj, line 82: similar with this, I wonder if it would make sense to just create the connection once? Unless there's already some kinda singleton logic in the manager?

sarangj (commented):
Just a few questions as an outside observer!

processes/record_ingestor.py
```diff
@@ -0,0 +1,113 @@
...
+
+        try:
+            records = self.source_service.get_records(
+                start_timestamp=utils.get_start_datetime(process_type=self.params.process_type, ingest_period=self.params.ingest_period),
+                offset=self.params.offset,
+                limit=self.params.limit
+            )
+
+            ingest_pipeline_pool = multiprocessing.Pool(processes=4)
+            record_ingest_processes = [ingest_pipeline_pool.apply_async(self._ingest_record, args=(record,)) for record in records]
```
mitri-slory, line 33: What are the 4 processes that will be in the multiprocessing pool?

processes/ingest/chicago_isac.py
kylevillegas93: good callout - i suppose each source process are in fact record ingestors but composition i think works well because it's just a bit more explicit and readable.

processes/record_ingestor.py
```diff
@@ -0,0 +1,113 @@
...
+from logger import create_log
+from managers import DBManager, RabbitMQManager, S3Manager
+from model import Record, FRBRStatus
+from services.sources.source_service import SourceService
+from . import utils
+
+logger = create_log(__name__)
+
+
+class RecordIngestor:
```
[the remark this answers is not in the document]
kylevillegas93: I think the other classes have a new line before the constructor so ill leave it for now!

processes/record_ingestor.py
kylevillegas93: Basically there are 4 subprocesses that will ingest records at a given time. 

This matches what the S3 process does - just spins up 4 processes.

processes/record_ingestor.py
kylevillegas93: it basically gives the the opportunity to scale better. once we've ingested a record and that's the only responsibility of each source ingestion process, we add it to a queue so any number of ECS tasks can process messages from that queue. That's where the record pipeline process comes into play - it FRBRizes and clusters and optionally fulfills the each record on the queue.

processes/record_ingestor.py
kylevillegas93: oof that's a good question - source_ids are likely going to be unique but we aren't enforcing that. 

I'll add a TODO !

processes/record_ingestor.py
kylevillegas93: unfortunately we can't pickle it amongst the subprocesses

processes/record_ingestor.py
kylevillegas93: lemme see if we can pickle it between the threads but i think it gave me some errors

processes/record_ingestor.py
kylevillegas93: nevermind - i like that - let's filter by source too

processes/record_ingestor.py
```diff
@@ -0,0 +1,97 @@
+import json
+from multiprocessing import Pool
+import os
+
+from logger import create_log
+from managers import DBManager, RabbitMQManager, S3Manager
+from model import Record
+from services.sources.source_service import SourceService
+from . import utils
+import tracemalloc
```
jackiequach, line 10: debugging import left here

jackiequach (approved):
This looks great, good work!

processes/record_ingestor.py
mitri-slory: Oh that makes a lot of sense and it definitely smooths the pipeline process out.

processes/record_ingestor.py
kylevillegas93: oops ! good catch!

mitri-slory (approved):
This looks really good!

PR #617, opened 2025-03-12
SFR-2527: Refactor MUSE mapping

jackiequach:
# Description
- Refactors the MUSE mapping 
- Updates ingest process to use RecordBuffer
- Fixes error with process params util function
    - `int() argument must be a string, a bytes-like object or a real number, not 'NoneType'`
- Removes MUSE unit tests

Work to be done in separate PRs to chunk the work: refactor the ingest process to service, add functional tests, move pdf manifest/has_part generation to mapping 

# Testing
`make integration; make functional`


processes/utils.py
```diff
@@ -33,7 +33,7 @@ def parse_process_args(*args) -> ProcessParams:
         custom_file=args[1] if len(args) > 1 else None,
         ingest_period=args[2] if len(args) > 2 else None,
         record_id=args[3] if len(args) > 3 else None,
-        limit=int(args[4]) if len(args) > 4 else None,
-        offset=int(args[5]) if len(args) > 5 else 0,
+        limit=int(args[4]) if len(args) > 4 and args[4] else None,
+        offset=int(args[5]) if len(args) > 5  and args[5] else 0,
```
kylevillegas93, line 37: Thank you :)

mappings/muse.py
```diff
@@ -1,100 +1,170 @@
...
-                ('700', '{a} {b} {c} {d}|||{e}'),
-                ('710', '{a} {b} {c} {d}|||{e}'),
-                ('711', '{a} {e}|||{j}'),
-            ],
-            'is_part_of': ('490', '{a}|{v}|volume'),
-            'has_part': [
-                ('856', '1|{u}|muse|text/html|{{"reader": false, "download": false, "catalog": false, "embed": true}}')
-            ]
-        }
+class MUSEMapping(BaseMapping):
```
kylevillegas93, line 10: Feel free to remove this class altogether and just have the function map_muse_record!

mappings/muse.py
```diff
@@ -1,100 +1,170 @@
...
+            date_created=datetime.now(timezone.utc).replace(tzinfo=None),
+            date_modified=datetime.now(timezone.utc).replace(tzinfo=None)
+        )
 
-    def applyFormatting(self):
-        self.record.source = 'muse'
-        self.record.source_id = list(self.record.identifiers[0].split('|'))[0]
+        return new_record
+
+    def _get_formatted_field(self, record, field, subfields, string_format):
```
kylevillegas93, line 53: may make sense to add types here so that we are explicit

mappings/muse.py
```diff
@@ -1,100 +1,170 @@
...
+
+        return [self._cleanup_identifier(id) for id in all_identifiers]
+    
+    def _get_authors(self, record):
+        all_authors = []
+
+        all_authors.extend(self._get_formatted_field(record, '100', ['a', 'b', 'c', 'd'], '{} {} {} {}|||true'))
+        all_authors.extend(self._get_formatted_field(record, '110', ['a', 'b', 'c', 'n', 'd'], '{} {} {} {} {}|||true'))
+
+        return all_authors
```
kylevillegas93, line 91: up to you but you could also do 

```
return [
  self._get_formatted_field(record, '100', ['a', 'b', 'c', 'd'], '{} {} {} {}|||true'),
  self._get_formatted_field(record, '110', ['a', 'b', 'c', 'n', 'd'], '{} {} {} {} {}|||true')
]
```

more of a style choice but it removes the all_authors array variable and shortens up the line length.

mappings/muse.py
kylevillegas93, line 53: What are your thoughts on adding the subfields in the string_format? 

That could simplify lines, e.g. 

```
self._get_formatted_field(record, '100', ['a', 'b', 'c', 'd'], '{} {} {} {}|||true'),
```
to
```
self._get_formatted_field(record, '100', '{a} {b} {c} {d}|||true'),
```

When we get into these more formatted strings, python's way of formatting a dict comes in handy.

mappings/muse.py
```diff
@@ -1,100 +1,170 @@
...
+    
+    def _get_title(self, record):
+        title_245 = self._get_formatted_field(record, '245', ['a', 'b'], '{} {}')
+        title_130 = self._get_formatted_field(record, '130', ['a'], '{}')
+
+        all_titles = list(set(title_245 + title_130))
```
kylevillegas93, line 97: hmm - does this return a set of the titles or a set of the two titles concatenated together?

mappings/muse.py
```diff
@@ -1,100 +1,170 @@
...
+
+        return all_titles[0]
+    
+    def _get_publishers(self, record):
+        publishers = self._get_formatted_field(record, '264', ['b'], '{}||')
+
+        if len(publishers) < 1:
+            publishers.append(DEFAULT_PUBLISHER)
+
+        return publishers
```
kylevillegas93, line 107: Could simplify here and do

```
return publishers or [DEFAULT_PUBLISHER]
```

mappings/muse.py
```diff
@@ -1,100 +1,170 @@
...
+    
+    def _get_dates(self, record):
+        dates = self._get_formatted_field(record, '264', ['c'], '{}|publication_date')
+
+        if len(dates) < 1:
+            publication_date = self.source['008'].data[11:15]
+            dates.append('{}|publication_date'.format(publication_date))
```
kylevillegas93, line 114: We could use an f string here

mappings/muse.py
```diff
@@ -1,100 +1,170 @@
...
+        if len(publishers) < 1:
+            publishers.append(DEFAULT_PUBLISHER)
+
+        return publishers
+    
+    def _get_dates(self, record):
+        dates = self._get_formatted_field(record, '264', ['c'], '{}|publication_date')
+
+        if len(dates) < 1:
+            publication_date = self.source['008'].data[11:15]
```
kylevillegas93, line 113: We might want to use the record here instead of the source. But could this come back None?

mappings/muse.py
```diff
@@ -1,100 +1,170 @@
...
+        all_contributors.extend(self._get_formatted_field(record, '710', ['a', 'b', 'c', 'd', 'e'], '{} {} {} {}|||{}'))
+        all_contributors.extend(self._get_formatted_field(record, '711', ['a', 'e', 'j'], '{} {}|||{}'))
+
+        return all_contributors
+
+    def add_has_part_link(self, url, media_type, flags):
+        last_item_no = int(self.record.has_part[-1][0])
+
+        self.record.has_part.append(
+            '{}|{}|muse|{}|{}'.format(last_item_no, url, media_type, flags)
```
kylevillegas93, line 164: let's use a Part dataclass here for consistency!

kylevillegas93 (changes requested):
Looks good! The mapping was definitely a tough one.

mappings/muse.py
jackiequach: This returns a list of titles, title_245 and title_130 are lists of titles. set() is used here to remove any dupes.

mappings/muse.py
jackiequach: Updated this to have a separate helper method _get_formatted_fields that calls _get_formatted_field for a list of passed in field and string formats with the subfields.

mappings/muse.py
jackiequach: Yes, it can be none, updated to handle this.

managers/muse.py
```diff
@@ -68,22 +69,25 @@ def identifyReadableVersions(self):
 
     def addReadableLinks(self):
         if self.pdfDownloadURL:
-            self.record.add_has_part_link(
+            add_has_part_link(
```
kylevillegas93, line 72: hmm - this is interesting - almost seems like a function that should be a part of the record class but we can pick that up later!

mappings/muse.py
```diff
@@ -1,139 +1,222 @@
...
+
+def map_muse_record(muse_record) -> Record:
+    identifiers = _get_identifiers(muse_record)
+    alternative = _get_formatted_field(muse_record, '240', '{a} {k}')
+    has_version = _get_formatted_field(muse_record, '250', '{a} {b}|')
+    spatial = _get_formatted_field(muse_record, '264', '{a}')
+    extent = _get_formatted_field(muse_record, '300', '{a}{b}{c}')
+    toc = _get_formatted_field(muse_record, '505', '{a}')
+
+    rights = 'muse|https://creativecommons.org/licenses/by-nc/4.0/||Creative Commons Attribution-NonCommercial 4.0 International|'
```
kylevillegas93, line 21: Recommend using the get_rights_string function here for consistency !

kylevillegas93 (approved):
Looks good!

I think we can ship this but we should circle back to add a functional test and create the service.

PR #620, opened 2025-03-13
SFR-2550: Implement FRBRize record step

kylevillegas93:
## Description
- Implements the FRBRize record step in the record pipeline process

##
`make functional`

### DIFF
--- a/processes/frbr/classify.py
+++ b/processes/frbr/classify.py
@@ -83,7 +83,7 @@ class ClassifyProcess():
             self.db_manager.session.add(unfrbrized_record)
             self.db_manager.session.commit()
 
-            if self.limit and self.classified_count >= self.limit:
+            if self.limit and self.record_buffer.ingest_count >= self.limit:
                 break
 
             if self.redis_manager.checkIncrementerRedis('oclcCatalog', 'API'):
@@ -138,7 +138,7 @@ class ClassifyProcess():
                 related_oclc_numbers=list(set(related_oclc_numbers))
             )
 
-            self.record_buffer.add(oclc_record)
+            self.record_buffer.add(oclc_record.record)
             self.get_oclc_catalog_records(oclc_record.record.identifiers)
 
     def get_oclc_catalog_records(self, identifiers):
--- a/processes/record_frbrizer.py
+++ b/processes/record_frbrizer.py
@@ -1,0 +1,120 @@
+from collections import defaultdict
+from lxml import etree
+import re
+
+from logger import create_log
+from mappings.oclc_bib import OCLCBibMapping
+from mappings.oclcCatalog import CatalogMapping
+from managers import DBManager, OCLCCatalogManager, RedisManager
+from model import Record
+from .record_buffer import RecordBuffer
+
+
+logger = create_log(__name__)
+
+
+class RecordFRBRizer:
+
+    def __init__(self, db_manager: DBManager):
+        self.db_manager = db_manager
+
+        self.oclc_catalog_manager = OCLCCatalogManager()
+        
+        self.redis_manager = RedisManager()
+        self.redis_manager.createRedisClient()
+
+        self.record_buffer = RecordBuffer(db_manager=self.db_manager)
+
+    def frbrize_record(self, record: Record) -> Record:
+        owi_to_oclc_numbers = self._add_works_for_record(record=record)
+        self._add_edition_records(owi_to_oclc_numbers=owi_to_oclc_numbers)
+
+        self.record_buffer.flush()
+
+        record.frbr_status = 'complete'
+
+        self.db_manager.session.add(record)
+        self.db_manager.session.commit()
+
+        logger.info(f'FRBRized record: {record}')
+
+        return record
+
+    def _add_works_for_record(self, record: Record) -> defaultdict:
+        queryable_ids = self._get_queryable_identifiers(record.identifiers)
+        author = record.authors[0].split('|')[0] if record.authors else None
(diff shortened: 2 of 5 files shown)


processes/frbr/classify.py
```diff
@@ -83,7 +83,7 @@ def classify_records(self, start_datetime: Optional[datetime]=None, record_uuid:
             self.db_manager.session.add(unfrbrized_record)
             self.db_manager.session.commit()
 
-            if self.limit and self.classified_count >= self.limit:
+            if self.limit and self.record_buffer.ingest_count >= self.limit:
```
kylevillegas93, line 86: This is my bad

processes/record_frbrizer.py
```diff
@@ -0,0 +1,120 @@
...
+            catalog_record_mapping.applyMapping()
+            catalog_record_mapping.record.identifiers.append('{}|owi'.format(owi_number))
+            
+            self.record_buffer.add(catalog_record_mapping.record)
+        except Exception:
+            logger.exception(f'Unable to map OCLC catalog record for OCLC number: {oclc_number}')
+            return
+
+    def _get_queryable_identifiers(self, identifiers) -> set:
+        return set([(id.split('|')[0], id.split('|')[1]) for id in identifiers if re.search(r'\|(?:isbn|issn|oclc)$', id) != None])
```
sarangj, line 120: small thing but you could just use a set comprehension:
```
return {(id.split('|')[0], id.split('|')[1]) for id in identifiers if...}
```

jackiequach (approved):
Looks good! This is a lot more readable and understandable than our current setup.

processes/record_frbrizer.py
```diff
@@ -0,0 +1,120 @@
...
+        record.frbr_status = 'complete'
+
+        self.db_manager.session.add(record)
+        self.db_manager.session.commit()
+
+        logger.info(f'FRBRized record: {record}')
+
+        return record
+
+    def _add_works_for_record(self, record: Record) -> defaultdict:
```
sarangj, line 43: just a note - returning a defaultdict *can* result in surprising behavior downstream (eg, a key appearing to be in there when it's not actually). Doesn't seem like it'll be a problem with how it's currently used, but might be worth future-proofing by returning a plain dict

processes/record_frbrizer.py
```diff
@@ -0,0 +1,120 @@
...
+            search_query = self.oclc_catalog_manager.generate_search_query(identifier=id, identifier_type=id_type)
+            owi_to_oclc_numbers.update(self._add_works(self.oclc_catalog_manager.query_bibs(query=search_query)))
+
+        if len(owi_to_oclc_numbers) == 0 and author and title:
+            search_query = self.oclc_catalog_manager.generate_search_query(author=author, title=title)
+            owi_to_oclc_numbers.update(self._add_works(self.oclc_catalog_manager.query_bibs(query=search_query)))
+
+        return owi_to_oclc_numbers
+
+    def _add_works(self, oclc_bibs: list) -> set:
```
sarangj, line 63: Seems like this typehint might be off? I think we're returning a dict here (or really, looks like `dict[set[str]]`?)

processes/record_frbrizer.py
```diff
@@ -0,0 +1,120 @@
...
+            )
+
+            self.record_buffer.add(record=oclc_bib_mapping.record)
+            owi_to_oclc_numbers[owi_number] = set(related_oclc_numbers)
+
+        return owi_to_oclc_numbers
+
+    def _add_edition_records(self, owi_to_oclc_numbers: defaultdict[set]):
+        for owi_number, oclc_numbers in owi_to_oclc_numbers.items():
+            cached_oclc_numbers = self.redis_manager.multiCheckSetRedis('catalog', list(oclc_numbers), 'oclc')
```
sarangj, line 91: Would it be worth doing this cache lookup once with all the oclc numbers in the dict?

sarangj (commented):
Since i don't really know what I'm doing here, mostly left style / readability comments, but looks good!

processes/record_frbrizer.py
kylevillegas93: good call!

processes/record_frbrizer.py
kylevillegas93: oops!

processes/record_frbrizer.py
kylevillegas93: ah ya good call

processes/record_frbrizer.py
kylevillegas93: interesting - good callout. I'll explore using a set of tuples instead. I think that it may be easier on memory and quicker

processes/record_frbrizer.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class RecordFRBRizer:
+
+    def __init__(self, db_manager: DBManager):
+        self.db_manager = db_manager
+
+        self.oclc_catalog_manager = OCLCCatalogManager()
+        
+        self.redis_manager = RedisManager()
```
sarangj, line 22: last question - is there a reason these are constructed here, while the db manager is passed in? My gut is to just do DI and pass everything in, but I also don't have a great handle on this codebase 😆

sarangj (approved):
seems reasonable to me!

processes/record_frbrizer.py
kylevillegas93: good question! we can but I think this might be the only step in the process that requires the redis_manager!

PR #621, opened 2025-03-13
NO-REF: Migrating Hathi to use source service

kylevillegas93:
## Description
- Moves Hathi ingest process to use source service
- Improves performance via ingesting from multiple files at once

## Testing
`make functional`

### DIFF
--- a/services/sources/hathi_trust_service.py
+++ b/services/sources/hathi_trust_service.py
@@ -1,0 +1,127 @@
+import csv
+from datetime import datetime
+import gzip
+from io import BytesIO
+import requests
+from typing import Optional, Generator
+
+from constants.get_constants import get_constants
+from mappings.hathitrust import HathiMapping
+from model import Record
+from .source_service import SourceService
+from logger import create_log
+
+
+logger = create_log(__name__)
+
+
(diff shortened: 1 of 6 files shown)


services/sources/hathi_trust_service.py
```diff
@@ -0,0 +1,127 @@
...
+        self.constants = get_constants()
+        self.record_count = 0
+
+    def get_records(
+        self, 
+        full_import: bool = False, 
+        start_timestamp: Optional[datetime] = None, 
+        offset: int = 0, 
+        limit: Optional[int] = None
+    ) -> Generator[Record, None, None]:
```
sarangj, line 33: It *might* be worth rethinking this method with a while loop or something instead of recursion - I found tracing how the record count var is used and set a little difficult to trace, especially, since it's an instance variable rather than just a simple param.  But that's more of a subjective thing, and I know unwinding this would be a bit of elbow grease!

sarangj (commented):
Some of my comments I know are just on code that was moved, so feel free to ignore!  The pattern seems really nice, just one thought here that may or may not make sense

PR #624, opened 2025-03-14
SFR-2550: Implement record cluster step

jackiequach:
# Description
- Implements record clustering step for the record pipeline

# Testing
`make functional`


processes/record_clusterer.py
```diff
@@ -0,0 +1,202 @@
...
+            Record.title.isnot(None),
+            Record.uuid == record_uuid
+        ]
+
+        get_unclustered_record_query = self.db_manager.session.query(Record).filter(*query_filters)
+
+        unclustered_record = get_unclustered_record_query.first()
+        if not unclustered_record:
+            logger.info(f'No unclustered record found with UUID {record_uuid}')
+            return
```
kylevillegas93, line 43: We should have the record from the record pipeline process - so method signature can be cluster_record(self, record: Record)

Then we don't need this logic!

processes/record_clusterer.py
```diff
@@ -0,0 +1,202 @@
...
+    def _cluster_matched_records(self, record_ids: list[str]):
+        records = self.db_manager.session.query(Record).filter(Record.id.in_(record_ids)).all()
+
+        kmean_manager = KMeansManager(records)
+
+        kmean_manager.createDF()
+        kmean_manager.generateClusters()
+        editions = kmean_manager.parseEditions()
+
+        return editions, records
```
kylevillegas93, line 86: I think the fulfill process will want the records back that have been clustered!

tests/functional/processes/frbr/test_cluster_process.py
```diff
@@ -106,3 +101,22 @@ def test_cluster_multi_item(db_manager, unclustered_multi_item_uuid):
...
+            .join(Work, Work.id == Edition.work_id)
+            .filter(Item.record_id == clustered_record.id)
+            .first()
+    )
+
+    item, edition, work = frbrized_model
+
+    assert item is not None
+    assert edition is not None
+    assert work is not None
```
kylevillegas93, line 122: let's add a newline here

tests/functional/processes/frbr/test_cluster_process.py
```diff
@@ -12,23 +13,17 @@ def test_cluster_process(db_manager, unclustered_record_uuid):
...
-    
-    frbrized_model = (
-        db_manager.session.query(Item, Edition, Work)
-            .join(Edition, Edition.id == Item.edition_id)
-            .join(Work, Work.id == Edition.work_id)
-            .filter(Item.record_id == clustered_record.id)
-            .first()
-    )
+@pytest.mark.skip
+def test_cluster_record(db_manager, unclustered_record_uuid):
```
kylevillegas93, line 19: we may want to add a separate record because I think it could get clustered by the other tests right?

processes/record_clusterer.py
```diff
@@ -0,0 +1,188 @@
...
+
+        work_data = record_manager.buildWork(records, editions)
+
+        record_manager.saveWork(work_data)
+
+        stale_work_ids = record_manager.mergeRecords()
+
+        return record_manager.work, stale_work_ids
+
+    def _index_works_in_elastic_search(self, works: Work):
```
kylevillegas93, line 156: oops - mind making the type hint change in the cluster process - i think this should be list[Work] there. 

But let's simplify here - I think we can just pass in 1 work

processes/record_clusterer.py
```diff
@@ -0,0 +1,188 @@
...
+        work_documents = []
+
+        for work in works:
+            elastic_manager = SFRElasticRecordManager(work)
+            elastic_manager.getCreateWork()
+            work_documents.append(elastic_manager.work)
+
+        self.elastic_search_manager.saveWorkRecords(work_documents)
```
kylevillegas93, line 164: We can get rid of the for loop!

We may want to add a TODO here to save a single work!

kylevillegas93 (approved):
looks good! just a couple minor comments

processes/cluster.py
```diff
@@ -237,7 +237,7 @@ def create_work_from_editions(self, editions, records):
 
         return record_manager.work, stale_work_ids
 
-    def index_works_in_elastic_search(self, works: Work):
```
kylevillegas93, line 240: THANK YOU

PR #625, opened 2025-03-17
NO-REF: Moving off of core process/refactoring to use process utils

kylevillegas93:
## Description
- Moves off the core process inheritance besides MUSE, which is in a separate PR
- Refactors the processes to use the utils functions 

## Testing
`make unit; make functional; make integration`

### DIFF
--- a/services/sources/source_service.py
+++ b/services/sources/source_service.py
@@ -10,9 +10,8 @@ class SourceService(ABC):
     @abstractmethod
     def get_records(
         self,
-        full_import: bool=False,
         start_timestamp: Optional[datetime]=None,
-        offset: Optional[int]=None,
+        offset: int=0,
         limit: Optional[int]=None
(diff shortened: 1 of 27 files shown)


services/sources/source_service.py
```diff
@@ -10,9 +10,8 @@ class SourceService(ABC):
     @abstractmethod
     def get_records(
         self,
-        full_import: bool=False,
```
mitri-slory, line 13: Since we don't have this full_import parameter anymore what is the refactor to replace the use of it?

mitri-slory (approved):
This looks really good. I only have one question about a change made but otherwise it's good to go.

services/sources/source_service.py
kylevillegas93: basically if the start_timestamp is not specified, the ingest processes will assume that it's a complete process!

PR #626, opened 2025-03-18
SFR-2552: Wire the record pipeline process

jackiequach:
# Description
- Wires the record pipeline process
- Fixes cluster bugs
- Adds better work/record logging

# Testing
`make functional`


model/postgres/record.py
```diff
@@ -120,9 +121,7 @@ def __init__(self, *args, **kwargs):
         self._deletion_flag = False
 
     def __repr__(self):
-        return '<Record(title={}, uuid={}, authors={})>'.format(
-            self.title, self.uuid, self.authors
-        )
+        return f'<Record(title={shorten(self.title, width=50)}, uuid={self.uuid})>'
```
jackiequach, line 124: How is width determined here?

kylevillegas93: width is the character limit but I think it truncates the entire word out if it doesn't fit within the limit.

kylevillegas93: 50 seems decent enough imo from testing but we can play around with it - just sometimes the titles are really long and it takes up a lot of space in the log statement.

kylevillegas93 (approved):
I think it's ready :)

PR #627, opened 2025-03-19
Add quickstart development setup guide to README.md

samanthaandrews:
- Create a new docker compose file to seed the DB
- Update the README.md with the latest install instructions for local development since new team members are onboarding this sprint
- Refactor the README.md to simplify the information architecture and remove outdated information


.python-version
samanthaandrews, line 1: @kylevillegas93 i committed this bc i think it's helpful to know which version of Python to install without having to ask someone on the team. lmk what you think.

kylevillegas93: sounds good :)

README.md
```diff
@@ -2,208 +2,269 @@
...
+   Verify the API is running by visiting: http://127.0.0.1:5050/apidocs/
 
-To run the processes individually in a local environment, the command should be in this format and should use the `local.yaml` file: 
-`python main.py -p APIProcess -e local` or `python main.py -p APIProcess -e local`.
+6. Initialize the local development environment:
 
-An example of running one of these processes is: `python main.py -p LOCProcess -e local -i complete`
+   ```bash
+   python main.py -p LocalDevelopmentSetupProcess -e local
+   ```
```
kylevillegas93, line 92: The local dev setup process should run as part of `docker compose up`

README.md
```diff
@@ -2,208 +2,269 @@
...
-- Update dependencies
+1. Create new version branch (e.g. `v0.12.0`)
+2. Create PR titled "Release v0.12.0"
+3. After approval, squash and merge to main
+4. Create Github release:
+   - Tag: v0.12.0
+   - Target: main
+   - Generate notes: `cat <(git log <first-commit>..HEAD --pretty=format:"%h - %s (%an, %ad)" --date=short)`
+5. Notify team in #researchnow_aka_sfr Slack channel
+6. Monitor deployment in Github Actions and ECS console
```
kylevillegas93, line 248: We can just say we are full CD - so maybe some details on what that process looks like.

kylevillegas93:
Overall, it's look great! Just a few minor comments and then let's ship it!

README.md
priya-pai: I had to run this again like Sam has here...but the command looks right in the dockerfile 🤔

README.md
```diff
@@ -2,208 +2,269 @@
...
+  - `daily`: Regular incremental import
+  - `complete`: Full import
+  - `custom`: Import from custom source
+- `--inputFile`/`-f`: Path to input file when using `custom` ingest type
+- `--startDate`/`-s`: Start date for custom period imports
+- `--limit`/`-l`: Limit number of records processed
+- `--offset`/`-o`: Skip first N records
+- `--singleRecord`/`-r`: Process single record by ID (ignores `ingestType`, `limit`, and `offset`)
+- `--source`/`-src`: Run against records from a specific source
+- `options`: Additional arguments passed as a list
```
sarangj, line 168: It's worth noting that because `main` uses the stdlib `argparse` for CLI args, it comes with a help (--h or -h) flag:
```
$ python -m main -h
usage: main.py [-h] [-p PROCESS] [-sc SCRIPT] -e ENVIRONMENT [-i INGESTTYPE] [-f INPUTFILE] [-s STARTDATE] [-l LIMIT]
               [-o OFFSET] [-r SINGLERECORD] [-src SOURCE]
               [options ...]

Run DCDW Data Ingest Jobs

positional arguments:
  options               Additional arguments

options:
  -h, --help            show this help message and exit
  -p PROCESS, --process PROCESS
                        The name of the process job to be run
  -sc SCRIPT, --script SCRIPT
                        The name of the script to run
  -e ENVIRONMENT, --environment ENVIRONMENT
                        Environment for deployment, sets env file to load
  -i INGESTTYPE, --ingestType INGESTTYPE
                        The interval to run the ingest over. Generally daily/complete/custom
  -f INPUTFILE, --inputFile INPUTFILE
                        Name of file to ingest. Ignored if -i custom is not set
  -s STARTDATE, --startDate STARTDATE
                        Start point for coverage period to query/process
  -l LIMIT, --limit LIMIT
                        Set overall limit for number of records imported in this process
  -o OFFSET, --offset OFFSET
                        Set start offset for current processed (for batched import process)
  -r SINGLERECORD, --singleRecord SINGLERECORD
                        Single record ID for ingesting an individual record
  -src SOURCE, --source SOURCE
                        Run a process against records from a specified source
```

README.md
```diff
@@ -2,208 +2,269 @@
...
+
+Processing:
+
+- `CatalogProcess`: Retrieve OCLC Catalog records
+- `ClassifyProcess`: FRBRize records
+- `ClusterProcess`: Group records into editions
+- `S3Process`: Fetch associated files (ePubs, covers)
+- `CoverProcess`: Fetch edition covers
+
+## Testing
```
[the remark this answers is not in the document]
sarangj: Maybe link to pytest docs since there would be lots of different ways to run tests with it?  Maybe a note that you can just run `python -m pytest` and use whatever flags are shows here: https://docs.pytest.org/en/stable/how-to/usage.html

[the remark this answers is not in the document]
sarangj: Not to bikeshed this already very useful PR further haha

README.md
```diff
@@ -2,208 +2,269 @@
...
+Data Ingestion:
+
+- `HathiTrustProcess`: Import from HathiTrust
+- `NYPLProcess`: Import from NYPL catalog
+- `GutenbergProcess`: Import from Project Gutenberg
+- `MUSEProcess`: Import from Project MUSE
+- `METProcess`: Import from MET Watson Digital Collections
+- `DOABProcess`: Import from Directory of Open Access Books
+- `LOCProcess`: Import from Library of Congress
+- `PublisherBacklistProcess`: Import from Publisher Backlist Project
```
sarangj, line 208: I do wonder if we want to maintain a list of these specifically since presumably, this list will change fairly often (?) and this feels likely to go stale. What do ya think of linking to https://github.com/NYPL/drb-etl-pipeline/tree/main/processes/ingest where these all live?

README.md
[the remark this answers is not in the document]
samanthaandrews: updated

README.md
samanthaandrews: i updated the docker-compose file and this seems to work now, so the only thing you need to run is `docker compose up` and everything works!

docker-compose.yml
```diff
@@ -89,6 +78,50 @@ services:
         target: /src
         read_only: true
 
+  seeddb:
```
samanthaandrews, line 81: @priya-pai @kylevillegas93 updated this file to include a seeddb step, which is nice bc now you don't have to run it separately. but it is a little annoying that it now runs every time you run `docker compose up`. lmk what you think.

docker-compose.yml
```diff
@@ -89,6 +78,50 @@ services:
...
+      devsetup:
+        condition: service_completed_successfully
+      database:
+        condition: service_started
+      elasticsearch:
+        condition: service_started
+      redis:
+        condition: service_started
+      s3:
+        condition: service_started
```
samanthaandrews, line 95: added all of these conditions bc otherwise it would sometimes fail when rabbitmq wasn't yet ready to accept connections or the devsetup wasn't completely finished

README.md
```diff
@@ -2,208 +2,206 @@
...
-- Add commenting/documentation strings
-- ~~Generate C4 diagrams for application~~
-- ~~Integrate ePub processor into standard processing flow~~
-- ~~Add cover fetching process~~
-- Create test suite, including:
-  - ~~Unit tests for all components~~
-  - Functional tests for each process
-  - Integration tests for the full cluster
-- Update dependencies
+This application uses continuous deployment (CD) via Github Actions to AWS ECS. The full CI/CD pipeline runs automatically when code is merged to `main`.
```
samanthaandrews, line 163: @kylevillegas93 full CD described here

README.md
```diff
@@ -2,208 +2,206 @@
...
+```bash
+# Run all tests using make commands
+make unit           # Run unit tests
+make integration    # Run integration tests
+make functional     # Run functional tests (requires running Docker environment)
 
-To keep sensitive settings out of git, some secrets configuration must be done to run the cluster. To set up for running on your local machine, copy the `config/example.yaml` file and provide the necessary configuration (ask a colleague if you need some of the keys required there). Then provide the name of this file as the `--environment` argument when you run scripts.
+# Run tests directly with pytest
+python -m pytest                     # Run all tests
+python -m pytest path/to/test.py     # Run a specific test file
```
samanthaandrews, line 154: @kylevillegas93 info on how to run one test

docker-compose.yml
priya-pai: Hmm personally, I think seeding the database upon each `docker compose up` is unnecessary since ideally, you would only do it once when setting up the DB. I like keeping it as a process only ran on demand. Just my two cents!

kylevillegas93: Could go either way too! But keeping it separate removes this container and depends on logic which keeps the compose file lean.

kylevillegas93 (approved):
Looks good to me - let's ship it! Thanks for leaving it better!

docker-compose.yml
sarangj: A compromise might be to just add a `Make` command that does it in a container (probably could just run the process in the running container for the API process)?  Basically just `docker exec -it drb_local_api python -m ...`

but maybe for another PR!

docker-compose.setup.yml
samanthaandrews, line 1: @sarangj @priya-pai @kylevillegas93 ok one more change: i separated out the devsetup and seeddb steps into a separate docker compose file, and in the readme you are instructed to run this as a one-time setup thing. then you can just run `docker compose up` after that. lmk what you think.

samanthaandrews: oop looks like this made tests fail

PR #628, opened 2025-03-20
SFR-2562: Creating separate session for each record pipeline message

kylevillegas93:
## Description
- Previous sessions were causing issues with subsequent db transactions. 
- Ensures that we rollback the changes if we fail to commit 
- Creates a new session per message and closes that session after
- Disposes the db engine after processing all messages 

## Testing
`make functional`

### DIFF
--- a/processes/record_pipeline.py
+++ b/processes/record_pipeline.py
@@ -62,7 +62,7 @@ class RecordPipelineProcess:
             self.link_fulfiller.fulfill_records_links(clustered_records)
                 
             self.rabbitmq_manager.acknowledgeMessageProcessed(message_props.delivery_tag)
-        except Exception as e:
+        except Exception:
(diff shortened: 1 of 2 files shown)


processes/record_pipeline.py
```diff
@@ -61,9 +62,12 @@ def _process_message(self, message):
             self.link_fulfiller.fulfill_records_links(clustered_records)
                 
             self.rabbitmq_manager.acknowledgeMessageProcessed(message_props.delivery_tag)
-        except Exception:
+        except Exception as e:
```
jackiequach, line 65: e is not being used

jackiequach (approved):
looks good

processes/record_pipeline.py
kylevillegas93: oop!

PR #629, opened 2025-03-20
SFR-2560: Add timeout for cluster process

jackiequach:
# Description
- Use [concurrent.futures](https://docs.python.org/3/library/concurrent.futures.html) to add timeout to the clustering process

# Testing
`make functional`


kylevillegas93:
I think we can work without this one for now - good work though! Feel free to close the request

PR #631, opened 2025-03-20
SFR-2565: Cleaning up S3 manager

kylevillegas93:
## Description
- Cleans up the S3 manager
- Removes unused functions
- Moves to snake case 

## Testing
`make unit`


managers/s3.py
```diff
@@ -83,81 +65,64 @@ def generate_manifest(self, record, source_url, manifest_url):
...
-
-        '''# Loading batches of data using a paginator until there are no more batches'''
-
-        paginator = self.s3Client.get_paginator('list_objects_v2')
-        page_iterator = paginator.paginate(Bucket=bucket, Prefix=objKey)
-        return page_iterator
 
     @staticmethod
-    def getmd5HashOfObject(obj):
+    def get_md5_hash(object):
```
sarangj, line 122: just a note: `object` is a reserved work in python (everything inherits from it) - *probably* won't cause any issues, but might trigger some linters / get highlighted weirdly

sarangj (approved):
seems like a good cleanup, especially like the removal of the `create` method!

PR #632, opened 2025-03-20
SFR-2563: Fix db mismatch issue

jackiequach:
# Description

- Fixes issue where the record fails to cluster due to stale data.
- Both `_create_work_from_editions` and `_delete_stale_works` may modify the same rows in the `Editons` or `Works` tables, so we need to commit after each transaction

### DIFF
--- a/processes/record_pipeline.py
+++ b/processes/record_pipeline.py
@@ -63,9 +63,9 @@ class RecordPipelineProcess:
                 
             self.rabbitmq_manager.acknowledgeMessageProcessed(message_props.delivery_tag)
         except Exception:
-            logger.exception(f'Failed to process record with source_id: {source_id} and source: {source}')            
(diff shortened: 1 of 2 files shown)


processes/record_pipeline.py
```diff
@@ -63,9 +63,9 @@ def _process_message(self, message):
                 
             self.rabbitmq_manager.acknowledgeMessageProcessed(message_props.delivery_tag)
         except Exception:
-            logger.exception(f'Failed to process record with source_id: {source_id} and source: {source}')            
+            logger.exception(f'Failed to process record with source_id: {source_id} and source: {source}')
+            self.rabbitmq_manager.reject_message(delivery_tag=message_props.delivery_tag)
```
kylevillegas93, line 67: oops!

kylevillegas93 (approved):

kylevillegas93 (approved):

PR #633, opened 2025-03-21
SFR-2550 Final record pipeline tweaks

kylevillegas93:
## Description
- Adds option to do complete redrive or partial redrive based on cluster status
- Adds option to view different queue message counts 

## Testing
`make functional`

### DIFF
--- a/processes/record_pipeline.py
+++ b/processes/record_pipeline.py
@@ -1,5 +1,6 @@
 import json
 import os
+from time import sleep
 
 from .record_frbrizer import RecordFRBRizer
 from .record_clusterer import RecordClusterer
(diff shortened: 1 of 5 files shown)


jackiequach (approved):
looks good!

processes/record_pipeline.py
```diff
@@ -28,15 +29,22 @@ def __init__(self, *args):
...
+                    logger.info(f'Waiting {wait_time}s for record messages')
+                    sleep(wait_time)
+
+                while message := self.rabbitmq_manager.getMessageFromQueue(self.record_queue):
+                    message_props, _, message_body = message
+
+                    if not message_props or not message_body:
+                        break
+                    
+                    self._process_message(message)
```
sarangj, line 47: Is this loop for retries? Do we need to add a try / except inside the loop body or am I misunderstanding?

sarangj: oh I see, it's for if it doesn't receive a message, carry on!

kylevillegas93: it is yes! we should migrate away from message pollers like this and instead use batch consumption. i was reading this morning that rabbitmq definitely tries to sway people away from 1x1 message pollers lol.

PR #635, opened 2025-03-24
SFR-2572: Implementing script to report missing files

kylevillegas93:
## Description
- Implements script ot report missing files

## Testing
`python main.py --script ReportMissingFiles -e qa options source=muse`

### DIFF
--- a/scripts/report_missing_files.py
+++ b/scripts/report_missing_files.py
@@ -16,7 +16,7 @@ def main(*args):
     s3_manager = S3Manager()
     s3_manager.createS3Client()
 
-    source_arg = next(filter(lambda option: option.startswith('source'), args), None)
+    source_arg = next((option for option in args if option.startswith('source=')), None)
     source = source_arg.split('=')[1] if source_arg else 'gutenberg'
 
     record_parts = db_manager.session.query(Record.has_part, Record.source_id).filter(Record.source == source).yield_per(1000)
@@ -36,8 +36,6 @@ def main(*args):
                     print(f'{source_id} is missing a file: {file_part}')
                     records_with_missing_files += 1
                     break
-            else:
-                requests.head(url)
                 
     print(f'Percentage of {source} records with missing files: {(records_with_missing_files / record_count) * 100}%')
 
--- a/scripts/report_missing_files.py
+++ b/scripts/report_missing_files.py
@@ -1,0 +1,46 @@
+import botocore
+import requests
+import sys
+
+from managers import DBManager, S3Manager
+from model import Part, Record
+
+
+'''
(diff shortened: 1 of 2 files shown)


scripts/report_missing_files.py
```diff
@@ -0,0 +1,46 @@
...
+
+            if file_part.file_bucket:
+                try:
+                    s3_manager.s3Client.head_object(Bucket=file_part.file_bucket, Key = file_part.file_key)
+                except botocore.exceptions.ClientError:
+                    print(f'{source_id} is missing a file: {file_part}')
+                    records_with_missing_files += 1
+                    break
+            else:
+                requests.head(url)
```
sarangj, line 40: should this be checking for a 404 or something and incrementing the missing files counter?

scripts/report_missing_files.py
```diff
@@ -0,0 +1,46 @@
...
+Usage:  python main.py --script ReportFileAccessibility -e <env> options source=<source>
+'''
+def main(*args):
+    db_manager = DBManager()
+    db_manager.createSession()
+
+    s3_manager = S3Manager()
+    s3_manager.createS3Client()
+
+    source_arg = next(filter(lambda option: option.startswith('source'), args), None)
```
sarangj, line 19: kinda subjective, but comprehensions are *generally* considered more pythonic:

```python
source_args = next((option for option in args if option.startswith("source"), None)
```

scripts/report_missing_files.py
kylevillegas93: oop - was messing around with but will get to it - want to focus on S3

scripts/report_missing_files.py
kylevillegas93: just learned actually you can do generator comps too!

scripts/report_missing_files.py
```diff
@@ -0,0 +1,44 @@
...
+            index, url, source, file_type, flags = part.split('|')
+            file_part = Part(index, url, source, file_type, flags)
+
+            if file_part.file_bucket:
+                try:
+                    s3_manager.s3Client.head_object(Bucket=file_part.file_bucket, Key = file_part.file_key)
+                except botocore.exceptions.ClientError:
+                    print(f'{source_id} is missing a file: {file_part}')
+                    records_with_missing_files += 1
+                    break
```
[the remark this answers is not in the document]
kylevillegas93: yes - just so we don't overcount - we just want to know the number of records that have at least 1 missing file.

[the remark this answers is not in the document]
sarangj: oh whoops, i missed that there were nested loops!

sarangj (approved):

PR #636, opened 2025-03-24
SFR-2573: Fix MET rights mapping

kylevillegas93:
## Description
- Fixes a rights mapping error in MET that prevented clustering
- Fixes bug where we were ingesting in copyright records
- Removes cover logic since no record has a cover through the implemented method. 

## Testing
`python main.py -p METProcess -e local`


jackiequach (approved):
Looks good!

PR #637, opened 2025-03-26
NO-REF: Improve report missing files script

kylevillegas93:
## Description
- Checks the head of files at a URI outside of S3
- Adds parallel processing
- Pulls record parts into memory - the session is not pickleable 

## Testing
```
python main.py --script ReportMissingFiles -e local-qa options source=nypl
nypl records with missing files: 1/20489 - total: 1031136
```

### DIFF
--- a/scripts/report_missing_files.py
+++ b/scripts/report_missing_files.py
@@ -1,5 +1,7 @@
 import botocore
+import json
 import requests
+from multiprocessing import Pool, Value, Lock
 import sys
 
 from managers import DBManager, S3Manager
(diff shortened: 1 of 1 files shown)


scripts/report_missing_files.py
```diff
@@ -1,43 +1,75 @@
...
+
+
+def is_record_missing_files(record_parts: tuple) -> bool:
     s3_manager = S3Manager()
     s3_manager.createS3Client()
 
+    parts, *_ = record_parts
+
+    for part in parts:
+        index, url, source, file_type, flags = part.split('|')
```
kristojorg, line 27: Could be added as a constructor on `Part` if we are constructing from string in multiple places

kristojorg (approved):
Code makes sense to me! Still getting used to Python again and setting up the local instance so didn't run it locally yet. So that the approval with a grain of salt ;)

scripts/report_missing_files.py
kylevillegas93: Good callout - we definitely want to move away from this pipe delimiter in the data model. We will get there lol.

PR #638, opened 2025-03-27
Monorepo v0

samanthaandrews:
This PR adds web code to the drb-etl-pipeline repo and separates it into the following structure:
- `.github/workflows/`
  - `<all workflow files here>`
- `etl-pipeline/`
- `web/`
- `.gitignore`
- `README.md` (currently empty but i'll be adding content in a future commit)

There are many ways that we could combine the code into a monorepo, and I'd like to get your thoughts on this structure before moving forward. As it stands, this PR mostly just puts all of the existing etl-pipeline code into a folder and adds the web code in a folder alongside it. Eventually, I think we should pull out the `api` code into its own top-level folder.

The structure for the `ereading-clients` monorepo was much different, but that repo included multiple different apps and all of the packages that were shared across those apps. In this case, we only have one app, so I didn't think it made sense to split it in the same way.

TO DO in this PR before merging:
- [ ] Rename the repo to `drb` or `digital-research-books`
- [ ] Add top-level README.md content
- [ ] Add instructions for how to run the backend & frontend at the same time with the PDF proxy
- [ ] Test CI files to make sure they all run with this new structure


kylevillegas93:
Looking good!

my recommendation is to split this work up into: 
1. Move the webapp in with it's GH workflows - merge and make sure deployment looks good
2. Move the etl-pipeline to a new folder - merge and make sure the ETL tasks work with the current task commands pointing to the main.py file - it may need to live outside both folders.

PR #639, opened 2025-03-27
NO-REF: Backfill Gutenberg files

kylevillegas93:
## Description
- Creating a separate ingest process to backfill the Gutenberg files
- Separates the S3 file storage logic into a record file saver

## Testing
`python main.py -p GutenbergV2Process -e local-qa`


processes/record_file_saver.py
```diff
@@ -0,0 +1,70 @@
...
+            if '.epub' in file_path:
+                file_root = '.'.join(file_path.split('.')[:-1])
+
+                web_pub_manifest = self.generate_webpub(file_root)
+                self.storage_manager.putObjectInBucket(web_pub_manifest, f'{file_root}/manifest.json', self.file_bucket)
+
+            logger.info(f'Stored file {file_path} from {file_url}')
+        except Exception as e:
+            logger.exception(f'Failed to store file {file_path} from {file_url}')
+            raise
```
kylevillegas93, line 35: oops - i will raise the error here

services/sources/gutenberg_service.py
```diff
@@ -40,7 +40,8 @@ def get_records(
         self,
         start_timestamp: Optional[datetime]=None,
         offset: int=0,
-        limit: Optional[int]=None
+        limit: Optional[int]=None,
+        record_only: bool=False,
```
kylevillegas93, line 44: this is kind of hacky but the gutenbergmapping class is not pickleable with multiprocessing.

processes/record_file_saver.py
```diff
@@ -0,0 +1,70 @@
...
+
+from digital_assets import get_stored_file_url
+from logger import create_log
+from managers import S3Manager
+
+
+logger = create_log(__name__)
+
+
+class RecordFileSaver:
```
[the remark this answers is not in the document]
kylevillegas93: ya im thinking it may go there - we need to decide where these steps in the pipeline live.

processes/record_file_saver.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __init__(self, storage_manager: S3Manager):
+        self.storage_manager = storage_manager
+        self.file_bucket = os.environ.get('FILE_BUCKET')
+
+    def store_file(self, file_url: str, file_path: str):
+        try:
+            file_contents = self.get_file_contents(file_url)
+            self.storage_manager.putObjectInBucket(file_contents, file_path, self.file_bucket)
+            del file_contents
```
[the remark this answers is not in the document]
kylevillegas93: dope - good callout.

services/sources/gutenberg_service.py
kylevillegas93: I like that. I think the goal with the eventual ETL pipeline is that we don't return mapping objects in the future!

PR #642, opened 2025-03-27
NO-REF: Adding user agent headers to missing files script

kylevillegas93:
## Description
- Was seeings lots of missing files - adding the user agent definitely ensures we are not getting any false positives.

### DIFF
--- a/scripts/report_missing_files.py
+++ b/scripts/report_missing_files.py
@@ -34,10 +34,11 @@ def is_record_missing_files(record_parts: tuple) -> bool:
             try:
                 s3_manager.s3Client.head_object(Bucket=file_part.file_bucket, Key = file_part.file_key)
             except botocore.exceptions.ClientError:
+                print(f'Missing S3 file: {url}')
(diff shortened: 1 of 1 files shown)


scripts/report_missing_files.py
```diff
@@ -34,10 +34,11 @@ def is_record_missing_files(record_parts: tuple) -> bool:
             try:
                 s3_manager.s3Client.head_object(Bucket=file_part.file_bucket, Key = file_part.file_key)
             except botocore.exceptions.ClientError:
+                print(f'Missing S3 file: {url}')
                 return True
         else:
             try:
-                url_head_response = requests.head(url)
+                url_head_response = requests.head(url, headers={ 'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5)' })
```
samanthaandrews, line 41: funny that you can set this programmatically!

samanthaandrews (approved):
🚀

PR #644, opened 2025-03-27
NO-REF: Implement Schomburg ingest

kylevillegas93:
## Description
- Implements Schomburg ingest 
- Adds pdf url from pdf pipeline bucket if there is a Hathi ID
- TODO - make sure files are moved over to the drb-files buckets and make sure they are publicly accessible 

## Testing
- Marked a record as ready to ingest in Airtable: https://airtable.com/appBoLf4lMofecGPU/tblXzsYyX9ZNrmP0H/viwifoeLgCJgZWzUu?blocks=hide
`python main.py -p PublisherBacklistProcess -e local`
`python main.py -p RedriveRecordsProcess -e local`
`python main.py -p RecordPipelineProcess -e local`

- Ensured that the record was properly clustered and available on frontend:
Screenshot 2025-03-27 at 17 42 37

### DIFF
--- a/services/sources/publisher_backlist_service.py
+++ b/services/sources/publisher_backlist_service.py
@@ -199,6 +199,7 @@ class PublisherBacklistService(SourceService):
                 try:
                     self.s3_manager.s3Client.head_object(Bucket=aws_file_bucket, Key=f'titles/publisher_backlist/Schomburg/{hathi_id}/{hathi_id}.pdf')
                 except Exception:
+                    logger.exception(f'PDF already exists at key: titles/publisher_backlist/Schomburg/{hathi_id}/{hathi_id}.pdf')
                     return None
                 
                 try:
--- a/services/sources/publisher_backlist_service.py
+++ b/services/sources/publisher_backlist_service.py
@@ -186,6 +186,7 @@ class PublisherBacklistService(SourceService):
     
     def download_file_from_location(self, record_metadata: dict, record_permissions: dict, record: Record) -> str:
         file_location = record_metadata.get('DRB_File Location')
+        aws_file_bucket = os.environ.get("FILE_BUCKET", "drb-files-local")
 
         if not file_location:
             hath_identifier = next((identifier for identifier in record.identifiers if identifier.endswith('hathi')), None)
@@ -205,15 +206,13 @@ class PublisherBacklistService(SourceService):
                 try:
                     s3.copy(
                         source_bucket_key,
-                        Bucket='drb-files-qa',
+                        Bucket=aws_file_bucket,
                         Key=f'titles/publisher_backlist/Schomburg/{hathi_id}/{hathi_id}.pdf'
                     )
                 except Exception:
                     logger.exception("Error during copy response")
 
-                # TODO - move converted PDFs over to drb-files S3 buckets
-                # TODO - ensure these files are publicly accessible
-                return f'https://pdf-pipeline-store-qa.s3.amazonaws.com/tagged-pdfs/{hathi_id}.pdf'
+                return f'https://{aws_file_bucket}.s3.us-east-1.amazonaws.com/titles/publisher_backlist/Schomburg/{hathi_id}/{hathi_id}.pdf'
                 
             raise Exception(f'Unable to get file for {record}')
 
--- a/services/sources/publisher_backlist_service.py
+++ b/services/sources/publisher_backlist_service.py
@@ -219,7 +219,8 @@ class PublisherBacklistService(SourceService):
                 except Exception:
                     logger.exception("Error during copy response")
 
-                return f'https://{aws_file_bucket}.s3.us-east-1.amazonaws.com/titles/publisher_backlist/Schomburg/{hathi_id}/{hathi_id}.pdf'
+                manifest_url = get_stored_file_url(storage_name=aws_file_bucket, file_path=f'titles/publisher_backlist/Schomburg/{hathi_id}/{hathi_id}.pdf')
+                return manifest_url
                 
             raise Exception(f'Unable to get file for {record}')
 
--- a/services/sources/publisher_backlist_service.py
+++ b/services/sources/publisher_backlist_service.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 import os
+import boto3
 import requests
 import urllib.parse
 from enum import Enum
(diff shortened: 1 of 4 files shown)


mitri-slory (commented):
This is looking good and I have one question about a section of code. I'll give further feedback once the TODOs are committed to this PR.

mitri-slory:
Additional comment: When I tried to ingest, frbrize, and cluster the SCH books in QA I couldn't see them on the QA frontend. I was using the old design of how we ran the FRBRization and cluster process though which might have contributed to it.

services/sources/publisher_backlist_service.py
```diff
@@ -158,41 +159,77 @@ def get_records(
...
+                        source_bucket_key,
+                        Bucket='drb-files-qa',
+                        Key=f'titles/publisher_backlist/Schomburg/{hathi_id}/{hathi_id}.pdf'
+                    )
+                except Exception:
+                    logger.exception("Error during copy response")
+
+                # TODO - move converted PDFs over to drb-files S3 buckets
+                # TODO - ensure these files are publicly accessible
+                return f'https://pdf-pipeline-store-qa.s3.amazonaws.com/tagged-pdfs/{hathi_id}.pdf'
```
sarangj, line 216: Now that we're copying over, I think the link here should be the DRB S3 key, right?

services/sources/publisher_backlist_service.py
```diff
@@ -158,41 +159,77 @@ def get_records(
...
+                    s3.head_object(Bucket=os.environ.get('PDF_BUCKET', None), Key=f'tagged_pdfs/{hathi_id}.pdf')
+                except Exception:
+                    logger.exception('PDF object does not exist')
+                    raise Exception
+
+                source_bucket_key = {'Bucket': os.environ.get('PDF_BUCKET', None), 'Key': f'tagged_pdfs/{hathi_id}.pdf'}
+                try:
+                    s3.copy(
+                        source_bucket_key,
+                        Bucket='drb-files-qa',
```
sarangj, line 208: should this be an env var so we can use this in prod? Can also be done in a separate PR!

services/sources/publisher_backlist_service.py
mitri-slory: Not the DRB S3 Key but the S3 object url to the pdf since this method is returned to a variable named 'file_url.'  I'll make this change in my next commit.

services/sources/publisher_backlist_service.py
mitri-slory: We actually have a env var in our config files for the DRB S3 Bucket called "FILE_BUCKET." I'll get the env var and replace the specific bucket name here with that instead.

services/sources/publisher_backlist_service.py
```diff
@@ -158,41 +159,76 @@ def get_records(
...
+        file_location = record_metadata.get('DRB_File Location')
+        aws_file_bucket = os.environ.get("FILE_BUCKET", "drb-files-local")
+
+        if not file_location:
+            hath_identifier = next((identifier for identifier in record.identifiers if identifier.endswith('hathi')), None)
+
+            if hath_identifier is not None:
+                hathi_id = hath_identifier.split('|')[0]
+
+                s3 = boto3.client("s3")
```
sarangj, line 197: small thing I missed earlier here, but I think we should just reuse the `self.s3_manager.s3Client` instead of instantiating a fresh client on every call here

services/sources/publisher_backlist_service.py
```diff
@@ -158,41 +159,76 @@ def get_records(
...
+
+                try:
+                    s3.head_object(Bucket=os.environ.get('PDF_BUCKET', None), Key=f'tagged_pdfs/{hathi_id}.pdf')
```
sarangj, line 200: it seems like we get the `PDF_BUCKET` key here multiple times with a default, even though all these operations will fail if there is no bucket var.  To be more explicit, I'd suggest putting a block like:

```
pdf_bucket = os.environ["PDF_BUCKET"]
```
This will raise early if the var is missing, which I think is what we want (since the s3 calls will fail anyway in that case).

sarangj (approved):
Looks good!  Just a couple pre-merge pieces of feedback

services/sources/publisher_backlist_service.py
```diff
@@ -158,41 +159,76 @@ def get_records(
...
+                source_bucket_key = {'Bucket': os.environ.get('PDF_BUCKET', None), 'Key': f'tagged_pdfs/{hathi_id}.pdf'}
+                try:
+                    s3.copy(
+                        source_bucket_key,
+                        Bucket=aws_file_bucket,
+                        Key=f'titles/publisher_backlist/Schomburg/{hathi_id}/{hathi_id}.pdf'
+                    )
+                except Exception:
+                    logger.exception("Error during copy response")
+
```
sarangj, line 214: Hmmm actually, couple more things!

Maybe we should avoid copying the object if it already is in the target bucket - eg, do a separate head request to see if the object is already in `FILE_BUCKET` before doing the copy.

Also, I wonder if the files in the DRB bucket are also public qfter running this. If not, we might need to use `copy_object` and set the `ACL` as "public-read": https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/copy_object.html#

But if it already makes them public, we're good on that I think!

services/sources/publisher_backlist_service.py
mitri-slory: That makes sense. i'll implement that.

services/sources/publisher_backlist_service.py
mitri-slory: I get your idea and this would help deduplicate the code here too.

services/sources/publisher_backlist_service.py
```diff
@@ -158,41 +159,82 @@ def get_records(
...
+        if not file_location:
+            hath_identifier = next((identifier for identifier in record.identifiers if identifier.endswith('hathi')), None)
+
+            if hath_identifier is not None:
+                hathi_id = hath_identifier.split('|')[0]
+
+                try:
+                    self.s3_manager.s3Client.head_object(Bucket=aws_file_bucket, Key=f'titles/publisher_backlist/Schomburg/{hathi_id}/{hathi_id}.pdf')
+                except Exception:
+                    return None
```
sarangj: Might make sense to add a small log here, like f"PDF already exists at {key}" or something!

sarangj (approved):
looks good to me!

PR #646, opened 2025-04-01
NO-REF: Adding retries to Gutenberg graphql queries

kylevillegas93:
## Description
- Adds retries to graphql queries
- Defaults to daily ingest if the ingest type is not provided


jackiequach (approved):
Looks good!

PR #648, opened 2025-04-02
NO-REF: Fix work/record string repr

kylevillegas93:
## Description
- Sometimes title may be empty which causes shorten to fail


jackiequach (commented):
If a record/work does not have a title, should we be ingesting it?

kylevillegas93:
@jackiequach - we should not - but we could have a title placeholder for records that don't have a title.

jackiequach (approved):
got it, that makes sense! this looks good!

PR #650, opened 2025-04-02
Monorepo v1

samanthaandrews:
- Moves all backend code into `etl-pipeline` folder
- Renames and updates GH Actions workflow files


etl-pipeline/Dockerfile
kylevillegas93, line 1: I think you might have to change the entrypoint to src/etl-pipeline/main.py

kylevillegas93: The ECS task definitions have the command and use the entry point in the Dockerfile: `ENTRYPOINT [ "python", "/src/main.py" ]`

```
"command": [
        "--process",
        "APIProcess",
        "--environment",
        "${environment}"
      ],
```

kylevillegas93: But maybe not since the working directory is etl-pipeline!

samanthaandrews: got it ok i see what you're saying. where are these ECS task definitions defined? in terraform? or directly in the ECS task?

kylevillegas93: They are defined in Terraform: https://stash.nypl.org/projects/SE/repos/nypl-terraform/browse/nypl-digital-dev/infrastructure/sfr-pipeline/scheduled_task_template.json

kylevillegas93: Looking at what you have for the web Dockerfile, I think just setting the working directory in the GH action should work

```
ENTRYPOINT ["node", "server.js"]

CMD ["r", "@newrelic/next"]
```

mitri-slory (approved):
This looks good to me!

PR #652, opened 2025-04-02
Implement a publisher backlist cover fetcher 

sarangj:
Add a new cover fetcher that looks for PDFs in DRBs publisher backlist directory in s3 and tries to "infer" a cover from a PDF, by looking for the first page in the PDF with a non-empty appearing image.

For determining "non-empty" images, we use the pillow package's entropy value, since these images are scans, and the blank images that sometimes are included in the PDF before the actual cover are not true blanks (ie, there are minor variations in pixel color)

Also note, instead of using the S3 API to access the files, I opted to use directly fetch the object urls using http requests, since the objects are public, and so that this could potentially be adapted to non-DRB hosted PDFs that are missing cover images.

A few open questions:

  1. I hardcoded 'Schomburg' into the s3 path, since those are the only backlist titles for now. However, maybe it makes sense to either move the PDFs into a folder without the backlist source in the name, so that this could work in the future for non-Schomburg titles?
  2. Should I set an explicit fetcher priority here? Might make sense to prioritize this last, since this fetcher is kind of guessing 🤷 
  3. Should I make this generic to any publicly downloadable PDF? The one thing I think preventing that is that the abstract fetcher class only has access to identifiers, so figuring out the url would require some changes to how the fetchers are constructed / used.  Happy to implement that though!

### DIFF
--- a/managers/coverFetchers/publisher_backlist_cover_fetcher.py
+++ b/managers/coverFetchers/publisher_backlist_cover_fetcher.py
@@ -1,0 +1,80 @@
+import hashlib
+import io
+import os
+
+import PIL
+import pypdf
+import requests
+import requests.exceptions
+
+from logger import create_log
+from managers.coverFetchers.abstractFetcher import AbstractFetcher
+
+logger = create_log(__name__)
+
+
+class PublisherBacklistCoverFetcher(AbstractFetcher):
+
+    def __init__(self, *args):
+        super().__init__(*args)
+
+        self.session = requests.Session()
+        self.bucket = os.environ["FILE_BUCKET"]
+        self.pdf_url: str | None = None
+
+
+    def hasCover(self):
+        for value, source in self.identifiers:
+            if source != "hathi":
+                continue
+
+            url = (
+                f"https://{self.bucket}.s3.us-east-1.amazonaws.com"
+                f"/titles/publisher_backlist/Schomburg/{value}/{value}.pdf"
+            )
+
+            response = self.session.head(url)
+
+            try:
+                response.raise_for_status()
+            except requests.exceptions.HTTPError:
+                continue
+            except Exception as e:
+                logger.exception("Unexpected error when fetching PDF")
+                continue
+            else:
+                self.pdf_url = url
+                return True
+
+        return False
+
+    def downloadCoverFile(self) -> bytes:
+        with io.BytesIO() as stream:
+            response = self.session.get(self.pdf_url, stream=True)
+            response.raise_for_status()
+            for chunk in response.iter_content(chunk_size=8192):
+                stream.write(chunk)
+
+            with pypdf.PdfReader(stream) as pdf:
+                image = PublisherBacklistCoverFetcher.find_cover_in_pdf(pdf)
+
+        with io.BytesIO() as outstream:
+            # image.tobytes() returns Pillow's internal image representation, which appears
+            # to not be a file type that is generally readable by a browser or Preview.
+            # Instead, write to a stream using the`save` method, and read the bytes off the
+            # stream
+            image.save(outstream, format="PNG")
+            return outstream.getvalue()
+
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


sarangj:
Worth noting, I haven't *fully* tested this e2e yet, but I've tested the class in isolation.  Happy to write an integration test of some kind!

priya-pai (approved):
This is looking great to me! All non-blocking comments. With regards to your questions:
# 1 - I like the idea to move PDFs into a more generic folder and allowing capability for non-Schomburg titles.

I'm going to leave # 2 and # 3 for other DRB devs because I'm not as familiar with how the fetchers are currently working. My two cents is prioritizing this fetcher last due to "inferring" going on makes sense to me. I'm also curious how common it is that any publicly downloadable PDFs do not provide covers.

managers/coverFetchers/publisher_backlist_cover_fetcher.py
```diff
@@ -0,0 +1,80 @@
...
+            else:
+                self.pdf_url = url
+                return True
+
+        return False
+
+    def downloadCoverFile(self) -> bytes:
+        with io.BytesIO() as stream:
+            response = self.session.get(self.pdf_url, stream=True)
+            response.raise_for_status()
```
Toxiapo, line 54: Do you think we should include a try/catch here?

Toxiapo (approved):
I noticed that the `__init__.py` file imports all the scripts in this folder. Do we need to add this new file there as well?

managers/coverFetchers/publisher_backlist_cover_fetcher.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, *args):
+        super().__init__(*args)
+
+        self.session = requests.Session()
+        self.bucket = os.environ["FILE_BUCKET"]
+        self.pdf_url: str | None = None
+
+
+    def hasCover(self):
```
kylevillegas93, line 26: has_cover!

kylevillegas93: oh wait ! is this part of the abstract fetcher? lol

managers/coverFetchers/publisher_backlist_cover_fetcher.py
```diff
@@ -0,0 +1,80 @@
...
+        for value, source in self.identifiers:
+            if source != "hathi":
+                continue
+
+            url = (
+                f"https://{self.bucket}.s3.us-east-1.amazonaws.com"
```
kylevillegas93, line 32: We are moving towards using the get_file_location method in the digital_assets module. This way it'll work locally as well.

managers/coverFetchers/publisher_backlist_cover_fetcher.py
```diff
@@ -0,0 +1,80 @@
...
+import requests
+import requests.exceptions
+
+from logger import create_log
+from managers.coverFetchers.abstractFetcher import AbstractFetcher
+
+logger = create_log(__name__)
+
+
+class PublisherBacklistCoverFetcher(AbstractFetcher):
```
kylevillegas93, line 16: Could this be just the PDFCoverFetcher? Understand we do have some hardcoding here for the Schomburg files.

kylevillegas93 (approved):

managers/coverFetchers/publisher_backlist_cover_fetcher.py
thinkwebengineer: I have a PR up for this to be a correct suggestion.

managers/coverFetchers/publisher_backlist_cover_fetcher.py
thinkwebengineer: download_cover_file will also change with the PR I have up

sarangj:
closing in favor of https://github.com/NYPL/digital-research-books/pull/670

PR #668, opened 2025-04-07
SFR-2630: Fix - do not proxy S3 urls

kylevillegas93:
## Describe your changes
- We shouldn't have to proxy S3 urls for files that we store in S3. 
- The issue with doing that sometimes is we get a 403 response back when trying to proxy the request. 
- In this case, we error out and never load the PDF in the web reader. 
- Removes changelog and adds description about CI/CD to README.

## How to test
```
npm run dev;
python main.py -p APIProcess -e local-production;
navigate to http://localhost:3000/read/18203225 and http://localhost:3000/read/8596055
```


kylevillegas93:
@jackiequach good catch - back to the drawing board :(

jackiequach (approved):
Looks good! Just test failures :/

PR #677, opened 2025-04-09
SFR-2630: Set proxy resources state to proxy urls

kylevillegas93:

### DIFF
--- a/web/src/components/ReaderLayout/ReaderLayout.tsx
+++ b/web/src/components/ReaderLayout/ReaderLayout.tsx
@@ -68,6 +68,7 @@ const ReaderLayout: React.FC<{
   const edition = link.work.editions[0];
   const [manifestUrl, setManifestUrl] = useState(url);
   const [isLoading, setIsLoading] = useState(true);
+  const [proxyResource, setProxyResource] = useState(false);
 
   const isEmbed = MediaTypes.embed.includes(link.media_type);
   const isRead = MediaTypes.read.includes(link.media_type);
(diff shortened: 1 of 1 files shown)


web/src/components/ReaderLayout/ReaderLayout.tsx
```diff
@@ -85,7 +86,8 @@ const ReaderLayout: React.FC<{
    */
   const getProxiedResource = (proxyUrl?: string) => async (href: string) => {
     // Generate the resource URL using the proxy if the URI is not stored in S3
-    const url: string = proxyUrl && !href.includes('drb-files')
+    setProxyResource(proxyUrl && !href.includes('drb-files'));
```
Toxiapo, line 89: Should we move this to the point where we first receive the manifest data so that it can handle multi-PDF resources as well?

Toxiapo: e.g.

```ts
const containsSelfHosted = manifest?.readingOrder?.some((link) =>
  link.href.includes("drb-files")
);
setProxyResource(proxyUrl && !isLimitedAccess && !containsSelfHosted);
```

kylevillegas93: lemme see! good callout

kylevillegas93: I think we are gonna keep the changes up in the getProxiedResource function - mainly because there could be resources that are self hosted but those that are not that are part of the same record - unlikely but something to consider

jackiequach (approved):
This seems like a good workaround for now. Tested with a few different sources and it looks good

PR #678, opened 2025-04-10
Resize PDF cover images

sarangj:
## Describe your changes
Noticed the PDF generated covers had much larger file sizes on average
than externally sourced covers. Dug a bit and saw that we're resizing
external covers in the cover manager, so extract that logic into a
separate module and use it in the new PDF cover generator.

Also adds a script to quickly do this to all the existing covers in the drb bucket, and to do a couple things
I missed while I'm here, namely, setting the covers to be public, and going ahead and copying the covers to prod.

## How to test
`python -m scripts.resize_publisher_backlist_images`

### DIFF
--- a/etl-pipeline/digital_assets/cover_images.py
+++ b/etl-pipeline/digital_assets/cover_images.py
@@ -1,0 +1,20 @@
+import PIL
+
+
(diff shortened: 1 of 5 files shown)


kylevillegas93 (approved):
looks good to me!

PR #680, opened 2025-04-10
fix: Paginate collection OPDS pubications

sarangj:
## Describe your changes
We've been fetching the collection and then later filtering with pagination params. My previous PR to eager load works and identifiers kinda made things worse, in that it led to us eager loading data for every edition, not just the filtered ones.

So, pass the pagination params into the DB method, first fetch the page of editions and then attach them to the collection object.

Id tried to do this using the orm loading strategies but couldn't really get it working, so I figured, just doing two pretty small quorm loading strategies but couldn't really get it working, so I figured, just doing two pretty small queries is fine and probably easier to understand.

## How to test
Hit the collections/{uuid} endpoint locally

### DIFF
--- a/etl-pipeline/api/blueprints/drbCollection.py
+++ b/etl-pipeline/api/blueprints/drbCollection.py
@@ -409,7 +409,7 @@ def get_collections():
             uuid = collection.uuid
             path = '/collection/{}'.format(uuid)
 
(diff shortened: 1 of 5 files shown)


mitri-slory (approved):
This looks good to me. I would suggest fixing the unit test failing before merging this into main.

PR #682, opened 2025-04-11
SFR-2614: Add file save and record save steps to record pipeline

kylevillegas93:
## Describe your changes
- Sends record messages to the queue
- Adds save file step in the record pipeline process
- Adds save record step in the record pipeline process

## How to test

### DIFF
--- a/etl-pipeline/processes/record_pipeline.py
+++ b/etl-pipeline/processes/record_pipeline.py
@@ -4,10 +4,11 @@ from time import sleep
 
 from .record_frbrizer import RecordFRBRizer
 from .record_clusterer import RecordClusterer
+from .record_file_saver import RecordFileSaver
 from .link_fulfiller import LinkFulfiller
 
 from logger import create_log
-from managers import DBManager, RabbitMQManager
+from managers import DBManager, RabbitMQManager, S3Manager
 from model import Record
 
 logger = create_log(__name__)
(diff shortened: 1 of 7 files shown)


etl-pipeline/processes/record_pipeline.py
```diff
@@ -18,13 +19,16 @@ class RecordPipelineProcess:
     def __init__(self, *args):
         self.db_manager = DBManager()
 
-        self.record_queue = os.environ['RECORD_PIPELINE_QUEUE']
-        self.record_route = os.environ['RECORD_PIPELINE_ROUTING_KEY']
+        self.record_queue = os.environ.get('RECORD_PIPELINE_QUEUE')
+        self.record_route = os.environ.get('RECORD_PIPELINE_ROUTING_KEY')
```
sarangj, line 23: curious why the move to `get`: I assume `create_or_connect_queue` will still fail if the env var is missing?

sarangj (approved):
🥳

etl-pipeline/processes/record_pipeline.py
kylevillegas93: ya it's a good question - the queue methods will fail. so i guess should we err out in the constructor or later on?

sarangj: I think just getting the key error here would be the most clear, instead of having to decipher why connecting to the queue failed.

kylevillegas93: copy!

PR #684, opened 2025-04-11
NO-REF: Fix muse chapter table card scrape

kylevillegas93:
## Describe your changes
- Muse looks to switch from span to h3 and from li elements to divs
- Adds a fallback for these elements just in case

## How to test
`make up; make functional`


sarangj (approved):
🤷 scraping lol

PR #686, opened 2025-04-14
Sfr 2597/Refactor `FRBRize` -> `Embellish`

kristojorg:
## Describe your changes

This is a simple naming refactor to rename the `RecordFRBRizer` to `RecordEmbellisher` and change related code. The motivation behind this is that the `RecordFRBRizer` was really responsible for bringing in new records for OCLC to give the clustering process more context. This is not actually the FRBRization process, which I would argue refers to the whole process of turning a record into a work/edition/item hierarchy.

## How to test

No testing should be needed so long as automated tests and typechecks pass


etl-pipeline/processes/record_embellisher.py
```diff
@@ -24,36 +32,41 @@ def __init__(self, db_manager: DBManager):
 
         self.record_buffer = RecordBuffer(db_manager=self.db_manager)
 
-    def frbrize_record(self, record: Record) -> Record:
+    def embellish_record(self, record: Record) -> Record:
         self._add_works_for_record(record=record)
 
         self.record_buffer.flush()
 
+        # TODO: change this to embellish_status
```
kristojorg, line 40: @kylevillegas93 how do you recommend making a change requiring a DB migration like this one?

kylevillegas93: let's leave it as is for now, especially since the classify process is still up and running while we implement the new redesign. feel free to create a ticket for it and we will get to it later

kylevillegas93 (approved):
These changes look good!

PR #687, opened 2025-04-14
Directly add covers to Schomburg editions

sarangj:
## Describe your changes
There's some order of operation issues wrt the Schomburg ingest and the covers - since the ingest was done before the cover stuff was added to it, the editions didn't get cover images and becaue of the non-determinism of the clustering, I don't want to re-run the ingest, as it will affect the number of editions / clusters in the collection.

So instead, just add the covers directly onto the images.
https://newyorkpubliclibrary.atlassian.net/browse/SFR-2559

## How to test
Ran this against QA and loaded a Schomburg edition: https://drb-qa.nypl.org/edition/7076343
Screen Shot 2025-04-14 at 10 09 15 AM

### DIFF
--- a/etl-pipeline/scripts/add_covers_to_schomburg_editions.py
+++ b/etl-pipeline/scripts/add_covers_to_schomburg_editions.py
@@ -1,0 +1,65 @@
+import argparse
+import os
+
+import boto3
+
+from digital_assets.utils.get_stored_file_url import get_stored_file_url
+from load_env import load_env_file
+from model import Collection, Edition, Link
+from managers import DBManager
+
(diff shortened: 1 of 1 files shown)


kylevillegas93 (approved):
This looks good ! the cover dimensions look a little long on the y-axis - is that to be expected?

sarangj:
Probably a little - I used the same resizing logic we do for other covers, but because these are pulled off of PDF pages and we try to more / less maintain aspect ratio, they're all gonna kind of end up looking 'long' (the height should be the same as the default, just less wide)

PR #688, opened 2025-04-14
fix: Properly load covers on collection page

sarangj:
## Describe your changes

Was testing some of the server side cover work I did and noticed I could get covers to load for an individual edition but not on the collection page - inspected the components and looks like we were trying to pass an OpdsPublication to a cover method that expects a full feed. Easy enough to just write a new cover finding method for a single edition.

https://newyorkpubliclibrary.atlassian.net/browse/SFR-2559

## How to test

Ran locally against the QA API and took a look at the Schomburg collection, saw that the covers are now loaded:
Screen Shot 2025-04-14 at 10 10 28 AM


kylevillegas93 (approved):
Looks good to me! :)

PR #690, opened 2025-04-14
Finish QA deploy before integration tests run

sarangj:
Something I noticed the other day - we tell ECS to deploy the new image and then immediately run integration tests. However, with how ECS deployments work, it's not clear whether the tests are hitting the old or new tasks. Waiting for services to be stable will ensure we're testing against the new image and will also avoid running the tests if the new instance never becomes healthy


sarangj:
Might be worth doing a similar thing for web

kylevillegas93 (approved):

kylevillegas93:
nice catch!

PR #694, opened 2025-04-14
Decouple RecordIngestor from SourceService

sarangj:
## Describe your changes

Some ingests have custom manifest logic, and future ingest processes might have more custom processing. So decouple the source service from the ingestor, instead simply passing a (usually lazy) collection of records to the ingestor, allowing the ingest process to do any special processing of records before handing them off to the ingestor.

For now, the changes in this PR are boilerplate, since we haven't transitioned any of the special ingests to the record ingestor

### DIFF
--- a/etl-pipeline/processes/ingest/nypl.py
+++ b/etl-pipeline/processes/ingest/nypl.py
@@ -15,7 +15,8 @@ class NYPLProcess():
 
     def runProcess(self):
         start_timestamp = utils.get_start_datetime(
-            process_type=self.params.process_type, ingest_period=self.params.ingest_period,
+            process_type=self.params.process_type,
+            ingest_period=self.params.ingest_period,
         ),
         return self.record_ingestor.ingest(
             self.nypl_bib_service.get_records(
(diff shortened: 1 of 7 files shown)


etl-pipeline/processes/ingest/nypl.py
```diff
@@ -11,7 +11,16 @@ class NYPLProcess():
...
         self.params = utils.parse_process_args(*args)
-        self.record_ingestor = RecordIngestor(source_service=self.nypl_bib_service, source=Source.NYPL.value)
+        self.record_ingestor = RecordIngestor(source=Source.NYPL.value)
 
-    def runProcess(self):    
-        return self.record_ingestor.ingest(params=self.params)
+    def runProcess(self):
+        start_timestamp = utils.get_start_datetime(
+            process_type=self.params.process_type, ingest_period=self.params.ingest_period,
+        ),
```
kylevillegas93, line 19: More of a style thing but could plop each param on a new line, tab indented - same with the above examples!

kylevillegas93 (approved):

etl-pipeline/processes/ingest/nypl.py
sarangj: done!

PR #695, opened 2025-04-14
SFR-2619: Extract deletion logic from publisher backlist ingest

kylevillegas93:
## Describe your changes
- Moves the deletion logic from publisher backlist ingest to a record deleter
- The record deleter delete_record function is called when the deletion flag on a record is true
- Still need to add a test
- Will add the deleted record retrieval back in after migrating out the rest of the logic 

## How to test

### DIFF
--- a/etl-pipeline/processes/record_deleter.py
+++ b/etl-pipeline/processes/record_deleter.py
@@ -1,0 +1,102 @@
+import os
+from sqlalchemy import joinedload
+
+from logger import create_log
+from model import Edition, Item, Record, Work
+from managers import DBManager, ElasticsearchManager, S3Manager
+
+logger = create_log(__name__)
+
+
+class RecordDeleter:
+
+    def __init__(self, db_manager: DBManager, store_manager: S3Manager, es_manager: ElasticsearchManager):
+        self.db_manager = db_manager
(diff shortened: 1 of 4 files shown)


etl-pipeline/processes/record_deleter.py
```diff
@@ -0,0 +1,102 @@
...
+        self.store_manager = store_manager
+        self.es_manager = es_manager
+
+    def delete_record(self, record: Record):
+        self._delete_record_digital_assets(record)
+        self._delete_record(record)
+
+        logger.info(f'Deleted {record}')
+
+    def _delete_record(self, record: Record):
```
mitri-slory, line 24: The names of both methods are nearly identical here. Is that intentional?

mitri-slory: I think a different name for the second delete record could be delete_whole_record

mitri-slory (approved):
This looks good to go besides one name suggestion I have for one of the methods.

etl-pipeline/processes/record_deleter.py
kylevillegas93: will separate it out!

PR #696, opened 2025-04-15
SFR-2599/extract candidate finding

kristojorg:
## Describe your changes

In this PR I extract the logic to find candidate records for clustering (a preprocessing step) into a separate class called the `RecordCandidateFinder`. This logic can be easier monitored and improved in future changes. 

## How to test

No logical changes were made - just semantic. Should be fine if tests pass and types pass.


etl-pipeline/processes/record_candidate_finder.py
```diff
@@ -0,0 +1,194 @@
...
+from model import Record
+from logger import create_log
+
+logger = create_log(__name__)
+
+
+class CandidateRecordFinder:
+    
+    # Maximum number of "hops" to follow when matching records through identifiers
+    MAX_MATCH_DISTANCE = 4
```
kylevillegas93, line 17: Is there a better variable name that explains what we are doing here with the hops?

etl-pipeline/processes/record_candidate_finder.py
```diff
@@ -0,0 +1,194 @@
...
+    # Maximum number of records that can be in a candidate pool
+    CANDIDATE_POOL_LIMIT = 10000
```
kylevillegas93, line 19: `MAX_NUMBER_OF_CANDIDATE_RECORDS` may be more explicit?

etl-pipeline/processes/record_candidate_finder.py
kylevillegas93, line 1: I like matching the file name with the class name - we can do candidate_record_finder.py - doesn't necessarily need to match up with the other step names. Or we can do RecordCandidateFinder for the class name!

etl-pipeline/processes/record_candidate_finder.py
```diff
@@ -0,0 +1,194 @@
...
+    # Regular expression to identify identifiers that should be matched
+    IDENTIFIERS_TO_MATCH = r"\|(?:isbn|issn|oclc|lccn|owi)$"
+    
+    def __init__(self, db_manager: DBManager):
+        self.db_manager = db_manager
+    
+    def find_candidate_records(self, record: Record) -> List[Record]:
+        """Find records that might be related to the input record.
+        Raises:
+            Exception: If the number of candidate records exceeds CANDIDATE_POOL_LIMIT
```
kylevillegas93, line 29: Hmm could we not throw an exception here and just cap the number of records returned?

etl-pipeline/processes/record_candidate_finder.py
```diff
@@ -0,0 +1,194 @@
...
+        Returns:
+            Formatted string for Postgres array overlap query
+        """
+        formatted_ids = []
+
+        for id in identifiers:
+            formatted_id = f'"{id}"' if re.search(r"[{},]{1}", id) else id
+            formatted_ids.append(formatted_id)
+
+        return "{{{}}}".format(",".join(formatted_ids))
```
kylevillegas93, line 194: These methods to me may be long in a separate file! They are also static and don't necessarily need to be part of a class.

etl-pipeline/processes/record_candidate_finder.py
```diff
@@ -0,0 +1,194 @@
...
+                    self.db_manager.session.query(
+                        Record.title, Record.id, Record.identifiers, Record.has_part
+                    )
+                    .filter(~Record.id.in_(already_matched_record_ids))
+                    .filter(Record.identifiers.overlap(id_batch))
+                    .filter(Record.title.isnot(None))
+                    .all()
+                )
+
+                matched_records.extend(records)
```
kylevillegas93, line 128: We may want to consider understanding the performance of this query at some point!

etl-pipeline/processes/record_candidate_finder.py
kristojorg: I had that thought too but didn't think I knew enough to change it yet : ). Happy to simply cap it though, especially if I can send a warning to NewRelic when that happens

etl-pipeline/processes/record_candidate_finder.py
```diff
@@ -0,0 +1,194 @@
...
+    def find_candidate_records(self, record: Record) -> List[Record]:
+        """Find records that might be related to the input record.
+        Raises:
+            Exception: If the number of candidate records exceeds CANDIDATE_POOL_LIMIT
+        """
+        candidate_record_ids = self._find_candidate_record_ids(record)
+        
+        # Include the original record's ID if it exists in the database
+        if record.id:
+            candidate_record_ids.append(record.id)
```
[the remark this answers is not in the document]
kristojorg: Good question - I just extracted this plainly from the `record_clusterer.py`. I can get rid of this and just allow an exception to occur if there is no id.

etl-pipeline/processes/record_candidate_finder.py
```diff
@@ -0,0 +1,194 @@
...
+        
+        # Query the actual Record objects
+        return self._get_records_by_ids(candidate_record_ids)
+    
+    def _find_candidate_record_ids(self, record: Record) -> List[str]:
```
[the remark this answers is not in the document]
kristojorg: Yea I hear ya. I sometimes feel it is easier, so long as the method does one thing, to keep it all in one place rather than making the reader click back and forth between multiple smaller methods that are only used in one place and essentially just become ways to "name" a certain block of code. Happy to split it up if you prefer tho!

etl-pipeline/processes/record_candidate_finder.py
```diff
@@ -0,0 +1,194 @@
...
+                        if re.search(self.IDENTIFIERS_TO_MATCH, id)
+                        and id not in checked_ids
+                    }
+                )
+                candidate_record_ids.add(matched_record_id)
+
+        if len(candidate_record_ids) > self.CANDIDATE_POOL_LIMIT:
+            raise Exception(
+                f"Candidate pool size {len(candidate_record_ids)} exceeds limit of {self.CANDIDATE_POOL_LIMIT}"
+            )
```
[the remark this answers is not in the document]
kristojorg: I think probably best answered with a bit of metrics. I am curious if we ever are even getting close to this limit. If not, then just record a warning when we hit it. If we are hitting it, then we can look at how to select a tighter candidate pool (or whether the max size actually should be larger)

etl-pipeline/processes/record_candidate_finder.py
kristojorg: I figured they should be colocated because I only seem them being used here (and the file size didn't seem too long yet), and then I debated making them free functions but ultimately made them static methods just because they are only really used as part of the class anyway. But a free function in the same file would be fine too. Curious what you prefer?

etl-pipeline/processes/record_candidate_finder.py
kylevillegas93: Ya we can make the change after!

etl-pipeline/processes/record_candidate_finder.py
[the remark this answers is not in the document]
kylevillegas93: Ya let's keep it and fine tune later!

etl-pipeline/processes/record_candidate_finder.py
kylevillegas93: let's iterate on it!

kylevillegas93 (approved):

PR #697, opened 2025-04-15
SFR-2612: DOAB Record Ingestor

mitri-slory:
## Describe your changes
This PR refactors the DOAB ingest process with the record ingestor process. It also makes changes to the `store_pdf_manifest` method in the `S3Manager` to compensate for the DOAB process having a custom webpub manifest process due to there being multiple manifest and epub links. Some questions of the implementation still remain such as:

1.  Do we add each manifest the `LinkManager` catches as a value of the has_part attribute or does one manifest link suffice? 
2. Do we save text/html files or only the epub file links?

## How to test
`make up`
`make functional`


etl-pipeline/managers/s3.py
```diff
@@ -27,27 +29,59 @@ def __init__(self):
...
+
+            link_manager.parse_links()
+
+            print(f'Length of manifests: {len(link_manager.manifests)}')
+            
+            for manifest in link_manager.manifests:
+                manifest_path, manifest_json = manifest
+                self.create_manifest_in_s3(manifest_path=manifest_path, manifest_json=manifest_json, bucket=bucket_name)
+
+                print(f'manifest_path={manifest_path}')
```
mitri-slory, line 44: I added these print statements to find that when running the functional test I get three of the same manifests from the link manager class which shouldn't be the case.

etl-pipeline/processes/record_ingestor.py
```diff
@@ -29,12 +29,12 @@ def ingest(self, records: typing.Iterator[Record]) -> int:
...
-                    queue_name=self.records_queue,
-                    routing_key=self.records_route,
-                    message=json.dumps(record.to_dict(), default=str)
+                        queue_name=self.records_queue,
+                        routing_key=self.records_route,
+                        message=json.dumps(record.to_dict(), default=str)
                 )
-
                 ingest_count += 1
+
```
kylevillegas93, line 37: hmm - i think we don't necessarily need the extra tabs and new line!

etl-pipeline/managers/s3.py
kylevillegas93, line 44: I recommend pulling this up into the doab mapper to get the right links/manifest into the has_part.

etl-pipeline/processes/ingest/doab.py
```diff
@@ -3,86 +3,46 @@
...
+        self.params = utils.parse_process_args(*args)
+        self.record_ingestor = RecordIngestor(source=Source.DOAB.value)
+
+    def runProcess(self) -> int:
+        start_timestamp = utils.get_start_datetime(
+            process_type=self.params.process_type,
+            ingest_period=self.params.ingest_period,
+        )
+        
+        if self.params.record_id is not None:
```
kylevillegas93, line 34: I think similar to muse we can have a get_record function in the source_service which is implemented in the child classes

kylevillegas93 (changes requested):
Looking good - just a few minor changes!

etl-pipeline/processes/ingest/doab.py
mitri-slory: I originally did put a get_record function in the source_service and then I realized that we would have to implement the function in every child classes despite if they implemented it or not. I could have some of the child classes just have the method and only have pass in the code block for now until we decide at a later time to implement each child source service with logic to retrieve a single record. What are your thoughts?

etl-pipeline/managers/s3.py
```diff
@@ -28,39 +27,9 @@ def __init__(self):
...
-                record.has_part.append(str(Part(
-                    index=1,
-                    url=epub_path,
-                    source=Source.LOC.value,
-                    file_type='application/epub+zip',
-                    flags=str(FileFlags(download=True)),
-                    source_url=epub_uri
-                )))
-        else:
+        if record.source != Source.DOAB.value:
```
kylevillegas93, line 32: I think we should steer away from having source-specific logic in any of our managers and the record pipeline.

etl-pipeline/mappings/doab.py
```diff
@@ -144,24 +147,54 @@ def _get_subjects(self, record, namespaces):
...
 
     def _get_has_part(self, record, namespaces):
+
+        has_part_array = []
+
+        link_manager = DOABLinkManager(record)
+
+        link_manager.parse_links()
+
+        print(f'Length of manifests: {len(link_manager.manifests)}')
```
kylevillegas93, line 157: let's remove this print statement

etl-pipeline/mappings/doab.py
```diff
@@ -144,24 +147,54 @@ def _get_subjects(self, record, namespaces):
...
+        
+        for manifest in link_manager.manifests:
+            manifest_path, manifest_json = manifest
+            S3Manager.create_manifest_in_s3(manifest_path=manifest_path, manifest_json=manifest_json, bucket=self.file_bucket)
+
+            print(f'manifest_path={manifest_path}')
```
kylevillegas93, line 163: same as above

etl-pipeline/managers/s3.py
mitri-slory: Ok that's a good conditional to make sure the manifest isn't being created again. That was my original intent for adding the `if record.source != Source.DOAB.value` conditional.

etl-pipeline/services/sources/dspace_service.py
```diff
@@ -76,6 +81,39 @@ def parse_record(self, record):
             return parsed_record
         except MappingError as e:
             raise Exception(e.message)
+        
+    def manage_links(self, record):
+
+        if record.source == 'doab':
+            #Not sure how to make it to where this method only runs for the DOAB ingest process besides specifying the source
```
kylevillegas93, line 88: Gothu - i'll make a couple changes!

sarangj (approved):

kylevillegas93 (approved):

PR #699, opened 2025-04-16
SFR-2619: Migrate publisher backlist project ingest process

kylevillegas93:
## Describe your changes
- Moves the file saving part from the publisher backlist project to the record file saver step
- Moves some of the mapping logic from the source service to the mapper

## How to test


etl-pipeline/managers/rabbitmq.py
```diff
@@ -88,13 +88,11 @@ def send_message_to_queue(self, queue_name: str, routing_key: str, message: Unio
                 routing_key=routing_key,
                 body=message
             )
-        except (ConnectionWrongStateError, StreamLostError):
-            logger.debug('Stale RabbitMQ connection - reconnecting.')
+        except (ConnectionWrongStateError, StreamLostError, ChannelWrongStateError):
```
sarangj, line 91: Gonna be so nice when sqs *just works* lol

etl-pipeline/mappings/publisher_backlist.py
```diff
@@ -44,6 +60,90 @@ def applyFormatting(self):
...
+                source=self.record.source,
+                file_type='application/pdf',
+                flags=str(
+                    FileFlags(
+                        download=record_permissions['is_downloadable'], 
+                        nypl_login=record_permissions['requires_login'], 
+                        fulfill_limited_access=record_permissions['requires_login']
+                    )
+                ),
+                source_url=f'https://{pdf_bucket}.s3.amazonaws.com/tagged_pdfs/{hathi_id}.pdf'
```
sarangj, line 116: `get_stored_file_url`?

etl-pipeline/managers/s3.py
```diff
@@ -26,15 +26,19 @@ def __init__(self):
...
+            manifest_json = self.generate_manifest(record=record, source_url=pdf_part.url, manifest_url=manifest_part.url)
+            self.create_manifest_in_s3(manifest_path=manifest_part.file_key, manifest_json=manifest_json, bucket=bucket_name)
+
+            return
 
         if pdf_part is not None:
-            if path:
-                manifest_path = f'manifests/{path}/{record.source}/{record_id}.json'
-            else:
-                manifest_path = f'manifests/{record.source}/{record_id}.json'
```
sarangj, line 37: Just to check - the existing thing puts these under the 'publisher_backlist' folder - are there any implications to no longer doing that?

etl-pipeline/processes/record_file_saver.py
```diff
@@ -26,20 +29,46 @@ def save_record_files(self, record: Record):
...
+        
+    def _store_file_from_drive(self, part: Part):
+        file_id = f'{self.drive_service.id_from_url(part.source_url)}'
+        file = self.drive_service.get_drive_file(file_id)
+
+        if not file:
+            raise Exception(f'Unable to get file for: {part}')
+
+        file_permissions = None if part.file_bucket == self.limited_file_bucket else 'public-read'
+        self.storage_manager.put_object(file.getvalue(), part.file_key, part.file_bucket, bucket_permissions=file_permissions)
```
sarangj, line 60: just fyi, you can use `s3.upload_fileobj` to directly send the bytesio stream, instead of loading all the bytes and passing them to s3.  I *think* that also results in a multipart upload, but not sure tbh lol

etl-pipeline/tests/functional/mappings/test_publisher_backlist.py
```diff
@@ -0,0 +1,50 @@
+import json
+import os
+
+from digital_assets import get_stored_file_url
+from mappings.publisher_backlist import PublisherBacklistMapping
+from model import Part, FileFlags
+
+
+def test_map_pubclisher_backlist_record():
```
sarangj, line 9: nit: s/pubclisher/publisher

Also, I feel like this could be a unit test?

sarangj (approved):
Generally looks good, feel free to consider the comments optional

etl-pipeline/mappings/publisher_backlist.py
kylevillegas93: urgh I think it's always going to be in the pdf_bucket !

etl-pipeline/tests/functional/mappings/test_publisher_backlist.py
kylevillegas93: ya this is one is tough since nothing is being mocked but we are functionally testing that the mapper works properly

etl-pipeline/managers/s3.py
kylevillegas93: I think the logic is the same - i just add the manifest part in the mapper with the proper path and flags!

etl-pipeline/processes/record_file_saver.py
kylevillegas93: Good call - changed!

etl-pipeline/mappings/publisher_backlist.py
sarangj: I see, because when running locally, we'll still want to grab from the remote bucket?

etl-pipeline/managers/s3.py
sarangj: Hmmm, I guess I'm just confused because of the removal of the `path` parameter from this method and from the manifest path in the line below.

etl-pipeline/mappings/publisher_backlist.py
kylevillegas93: I think so! It won't really work tho :(

PR #700, opened 2025-04-16
SFR-2617: Migrate MUSE Ingest Process to use Record Ingestor

mitri-slory:
## Describe your changes
This PR creates a MUSE Service process that currently uses the record buffer and in a future PR will use the record ingestor instead. The store pdf manifest and epub logic is laid out in the MUSE ingest process within the runProcess method instead of separate methods since the logic will be refactored into the store pdf manifest and epub methods of the S3Manager in a future PR. I will add snake casing to the rest of the variables and methods in a future commit.

## How to test
`make up` 
`make functional`


etl-pipeline/services/sources/source_service.py
```diff
@@ -12,6 +12,7 @@ def get_records(
         self,
         start_timestamp: Optional[datetime]=None,
         offset: int=0,
-        limit: Optional[int]=None
+        limit: Optional[int]=None,
+        record_id: Optional[str]=None
```
kylevillegas93, line 16: Hmm what are your thoughts on creating a separate method get_record for cases where we just want to get a single record?

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+MARC_URL = 'https://about.muse.jhu.edu/lib/metadata?format=marc&content=book&include=oa&filename=open_access_books&no_auth=1'
+MARC_CSV_URL = 'https://about.muse.jhu.edu/static/org/local/holdings/muse_book_metadata.csv'
+MUSE_ROOT_URL = 'https://muse.jhu.edu'
+
+class MUSEService(SourceService):
+
+    def __init__(self):
+        pass
+
+    def get_records(
```
kylevillegas93, line 29: See below -

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+        self,
+        start_timestamp: Optional[datetime]=None,
+        offset: int=0,
+        limit: Optional[int]=None,
+        record_id: Optional[str]=None
+    ) -> Generator[Record, None, None]:
+        yield from self.importMARCRecords(start_timestamp=start_timestamp, limit=limit, record_id=record_id)
+        
+    def importMARCRecords(self, start_timestamp, limit, record_id):
```
kylevillegas93, line 38: import_marc_records

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+        self.updateDates = {}
+        for row in csvReader:
+            try:
+                self.updateDates[row[7]] = \
+                    datetime.strptime(row[11], '%Y-%m-%d')
+            except (IndexError, ValueError):
+                logger.warning('Unable to parse MUSE')
+                logger.debug(row)
+
+    def recordToBeUpdated(self, record, startDate, record_id):
```
kylevillegas93, line 106: hmm - i know this was brought over from the ingest process, but would it be better to have skip_record_update and return true  if we should skip updating the record?

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+        recordURL = record.get_fields('856')[0]['u']
+
+        updateDate = self.updateDates.get(recordURL[:-1], datetime(1970, 1, 1))
+
+        updateURL = 'https://muse.jhu.edu/book/{}'.format(record_id)
+
+        return (updateDate >= startDate) or updateURL == recordURL[:-1]
```
kylevillegas93, line 113: Let's convert these to snake case

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+            next(csvReader, None)  # Skip 4 header rows
+
+        self.updateDates = {}
+        for row in csvReader:
+            try:
+                self.updateDates[row[7]] = \
+                    datetime.strptime(row[11], '%Y-%m-%d')
+            except (IndexError, ValueError):
+                logger.warning('Unable to parse MUSE')
+                logger.debug(row)
```
kylevillegas93, line 104: Let's make these variables snake case!

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+            content += chunk
+
+        return BytesIO(content)
+
+    def download_record_updates(self):
+        try:
+            csvResponse = requests.get(MARC_CSV_URL, stream=True, timeout=30)
+            csvResponse.raise_for_status()
+        except(ReadTimeout, HTTPError) as e:
+            logger.error('Unable to load MUSE CSV records')
```
kylevillegas93, line 85: We could use exception instead of error here to gather more error details

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+            logger.debug(e)
```
kylevillegas93, line 86: These debug statements are really only helpful if the log level is debug, which is only ever locally if you change it. We could remove these.

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+            try:
+                mapped_muse_record = map_muse_record(marc_record)
+                if mapped_muse_record is not None:
+                    yield mapped_muse_record
+                    processed_record_count += 1
+            except MUSEError as e:
+                logger.warning('Unable to parse MUSE record')
+                logger.debug(e)
+
+    def download_marc_records(self):
```
kylevillegas93, line 65: Let's move these method variables to snake case!

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+            except (IndexError, ValueError):
+                logger.warning('Unable to parse MUSE')
+                logger.debug(row)
+
+    def recordToBeUpdated(self, record, startDate, record_id):
+        recordURL = record.get_fields('856')[0]['u']
+
+        updateDate = self.updateDates.get(recordURL[:-1], datetime(1970, 1, 1))
+
+        updateURL = 'https://muse.jhu.edu/book/{}'.format(record_id)
```
kylevillegas93, line 111: We could use an f string here as well as use the base muse url !

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+                    and self.recordToBeUpdated(marc_record, start_timestamp, record_id)\
+                    is False:
+                continue
+
+            try:
+                mapped_muse_record = map_muse_record(marc_record)
+                if mapped_muse_record is not None:
+                    yield mapped_muse_record
+                    processed_record_count += 1
+            except MUSEError as e:
```
kylevillegas93, line 61: Is the MUSEError thrown from the mapper? we may want to generally catch exceptions here

etl-pipeline/services/sources/muse_service.py
```diff
@@ -0,0 +1,113 @@
...
+        record_id: Optional[str]=None
+    ) -> Generator[Record, None, None]:
+        yield from self.importMARCRecords(start_timestamp=start_timestamp, limit=limit, record_id=record_id)
+        
+    def importMARCRecords(self, start_timestamp, limit, record_id):
+        self.download_record_updates()
+
+        muse_file = self.download_marc_records()
+
+        marcReader = MARCReader(muse_file)
```
kylevillegas93, line 43: marc_reader

kylevillegas93 (changes requested):
Looking good! Few minor comments!

etl-pipeline/services/sources/muse_service.py
mitri-slory: The MUSEError comes from the MUSE manager but I agree that we can just put an exception here.

etl-pipeline/services/sources/source_service.py
mitri-slory: That method could be added but I think the priority now should be to get this service merged in and implement the record ingestor in another PR. This method can be added in a separate ticket later on I believe. What are your thoughts?

etl-pipeline/services/sources/muse_service.py
mitri-slory: I can agree with this logic and makes the code logic more clear to see return True indicating a record should be skipped.

sarangj (approved):

kylevillegas93 (approved):

PR #701, opened 2025-04-16
Migrate gutenberg to RecordIngestor

sarangj:
Migrate to using the record ingestor for the gutenberg ingest. Tracing through this, I noticed that the mapping creates some record parts, but then the ingest clears them out and reworks them to add the s3 urls. Instead, I opted to just move everything into the ingest, but I can instead invert that and move it all back to the mapping.

## How to test
`python -m pytest tests/functional/processes/ingest/test_gutenberg_process.py`


etl-pipeline/mappings/gutenberg.py
```diff
@@ -1,14 +1,16 @@
 import json
 import re
 
+from digital_assets.utils.get_stored_file_url import get_stored_file_url
```
sarangj, line 4: remove

etl-pipeline/mappings/gutenberg.py
```diff
@@ -1,14 +1,16 @@
...
 from mappings.xml import XMLMapping
+from model import FileFlags, Part
```
sarangj, line 6: remove

etl-pipeline/processes/record_clusterer.py
```diff
@@ -17,7 +17,7 @@
 
 class RecordClusterer:
     """Clusters related bibliographic records and transforms them into FRBR model objects.
-    
+
```
sarangj, line 20: I'll remove these whitespace changes from this PR

etl-pipeline/processes/record_file_saver.py
```diff
@@ -61,10 +60,11 @@ def get_file_contents(self, file_url: str):
...
             raise e
-    
-    def generate_webpub(self, file_root: str):
-        file_path = get_stored_file_url(storage_name=self.file_bucket, file_path=f'{file_root}/META-INF/container.xml')
-        webpub_conversion_url = f'{RecordFileSaver.WEBPUB_CONVERSION_BASE_URL}/api/{quote_plus(file_path)}'
+
+    def generate_webpub(self, source_url: str):
+        webpub_conversion_url = (
+            f'{RecordFileSaver.WEBPUB_CONVERSION_BASE_URL}/api/{quote_plus(source_url)}'
+        )
```
sarangj, line 67: oh this makes no sense lol, I'll add some kind of fixture / mock webpub converter.

sarangj:
Functional test will still fail (we're dropping like 10 records somewhere), but otherwise this is ready to go

etl-pipeline/processes/ingest/gutenberg.py
```diff
@@ -117,16 +105,15 @@ def add_cover(self, gutenberg_record: GutenbergMapping):
             file_type = re.search(r'(\.[a-zA-Z0-9]+)$', cover_data.get('image_path')).group(1)
             cover_path = 'covers/gutenberg/{}{}'.format(gutenberg_id, file_type)
             cover_url = get_stored_file_url(self.file_bucket, cover_path)
+            cover_root = yaml_file.get('url').replace('ebooks', 'files')
+            cover_source_url = f"{cover_root}/{cover_data.get('image_path')}"
```
sarangj, line 109: This url 404s most of the time - I can add a HEAD request here to check for file existence. Alternatively, it looks getting the image from the github repo is more reliable, should I swap to doing that?

sarangj: Third alternative - the epub contains a cover which is saved to s3 when the epub is unzipped. We could potentially build something into the record file saver that handles that, but I think that would be a separate ticket that I can create.

sarangj:
Test passes locally 🤷

etl-pipeline/tests/conftest.py
```diff
@@ -340,3 +341,15 @@ def limited_access_record_uuid(db_manager):
...
+@pytest.fixture
+def mock_epub_to_webpub():
+    with open("tests/fixtures/webpub-manifest.json", "rb") as f:
+        response_content = f.read()
+
+    with requests_mock.Mocker() as m:
+        m.get(
+            "https://epub-to-webpub.vercel.app/api",
+            content=response_content,
+        )
```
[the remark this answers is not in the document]
kylevillegas93: ya i like that idea

etl-pipeline/processes/ingest/gutenberg.py
```diff
@@ -19,87 +23,73 @@ class GutenbergProcess():
...
-                    flags=part.flags
-                )))
-
-        gutenberg_record.record.has_part = epub_parts
+            count += 1
+            yield record_mapping.record
+
+        return count
+
+    def add_epub_links(self, record: Record):
```
kylevillegas93, line 57: Hmm this is interesting - do we do this already? I think we might just store the epub+zip file to download and then the manifest in the reader has all the links to the files within the zip?

kylevillegas93, line 57: We could consider moving this to the mapper

etl-pipeline/processes/ingest/gutenberg.py
```diff
@@ -19,87 +23,73 @@ class GutenbergProcess():
...
+                source="gutenberg",
+                url=get_stored_file_url(
+                    self.file_bucket,
+                    f"epubs/gutenberg/{gutenberg_id}_{extension}/manifest.json",
+                ),
+                file_type="application/webpub+json",
+                flags=str(FileFlags(reader=True)),
+            )))
 
     def add_cover(self, gutenberg_record: GutenbergMapping):
```
kylevillegas93, line 94: same as above - we could move this!

etl-pipeline/processes/ingest/gutenberg.py
```diff
@@ -117,16 +107,16 @@ def add_cover(self, gutenberg_record: GutenbergMapping):
             file_type = re.search(r'(\.[a-zA-Z0-9]+)$', cover_data.get('image_path')).group(1)
             cover_path = 'covers/gutenberg/{}{}'.format(gutenberg_id, file_type)
             cover_url = get_stored_file_url(self.file_bucket, cover_path)
+            cover_root = yaml_file.get('url').replace('ebooks', 'files')
+            cover_source_url = f"{cover_root}/{cover_data.get('image_path')}"
+            response = requests.head(cover_source_url, allow_redirects=True)
+            response.raise_for_status()
```
kylevillegas93, line 113: We could skip this and add this step in the record file saver step.

kylevillegas93: But i understand we may want to double check

etl-pipeline/processes/ingest/gutenberg.py
```diff
@@ -2,13 +2,17 @@
...
+
+import requests
 
 from digital_assets import get_stored_file_url
 from mappings.gutenberg import GutenbergMapping
 from managers import DBManager, RabbitMQManager
-from model import get_file_message, FileFlags, Part, Source
+from model import get_file_message, FileFlags, Part, Record, Source
 from logger import create_log
 from ..record_buffer import RecordBuffer
```
kylevillegas93, line 14: we may not need the record_buffer?

kylevillegas93:
It's looking good. 1 thing to consider is to move the has_part logic from the ingest process to the mapper. Then there may not need to be preprocessing between the ingest process and the record ingestor.

sarangj:
> It's looking good. 1 thing to consider is to move the has_part logic from the ingest process to the mapper. Then there may not need to be preprocessing between the ingest process and the record ingestor.

I'll do that - the reason I opted for this approach was to avoid coupling the mapping with stuff like the file bucket or fetching the cover image to set the has_part field, but I think it makes more sense to just do that to follow the new architecture better!

etl-pipeline/processes/ingest/gutenberg.py
sarangj: Yes, currently we add the epub links in the mapper: https://github.com/NYPL/digital-research-books/blob/4ce2de149901f4eebcf86f32e8fd751601ee3d38/etl-pipeline/mappings/gutenberg.py#L87-L105

Then we essentially explode the epub parts into download and read parts in the ingest: https://github.com/NYPL/digital-research-books/blob/4ce2de149901f4eebcf86f32e8fd751601ee3d38/etl-pipeline/processes/ingest/gutenberg.py#L61-L102

I opted to simplify by moving all this into one place in the ingest process, but simple enough to move it all into the mapping. 

On the other hand, I could handle this all in the file saver, thoughts?

sarangj: I'd just need to see whether other epub ingests are doing a similar thing

sarangj: done!

kylevillegas93: ah interesting - ya ive thought about that option. may simplify/consolidate things for epubs

etl-pipeline/processes/ingest/gutenberg.py
sarangj: Currently, the file saver will fail if any of the parts fail to save, which I think makes sense, as the new ETL is designed to not save the records without the files. I first went about changing the file saver to not fail specifically for cover images, but opted not to add that logic in favor of keeping it in the specific ingest, mainly to avoid making a decision *now* that affects other future ingests without more knowledge.

However, happy to implement it there!

sarangj:
and just to clarify 'preprocessing' - because we're using generators, there's not *extra* processing of anything in the first implementation I put up here - just a matter of where the code goes!

etl-pipeline/processes/ingest/gutenberg.py
sarangj: https://newyorkpubliclibrary.atlassian.net/browse/SFR-2664

kylevillegas93 (approved):
This looks good to me!

PR #702, opened 2025-04-18
Add an editorconfig

sarangj:
See https://editorconfig.org/

I opted to just have a single file in the root dir for simplicity, but I think you can nest these. Open to either implementation

### DIFF
--- a/etl-pipeline/.editorconfig
+++ b/etl-pipeline/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig is awesome: https://editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# Maybe this should be Python-specific under etl-pipeline?
+[*.py]
+indent_style = space
+indent_size = 4


etl-pipeline/.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# Maybe this should be Python-specific under etl-pipeline?
```
kylevillegas93, line 11: hmm - since this is inside the etl-pipeline? Should it only look at files within this folder?

kylevillegas93 (approved):
Looks good to me!

mitri-slory (approved):

etl-pipeline/.editorconfig
sarangj: d'oh yes, I meant to put this at the top level. Though I think the way this thing is read is relative to the dir it is in, so I think you don't need to specify `etl-pipeline`. I think I'll just leave this as is for now, and we can look into moving this up to the rootdir if someone wants to share whitespace rules with the web code.

PR #704, opened 2025-04-21
SFR-2629 sqs manager

thinkwebengineer:
## Describe your changes
This is an implementation of the SQS Manager, this implementation should allow for us to recover any last messages from rabbit mq and seamlessly move to SQS.  We will need 2 more PRs for this too. One to replace RabbitMQManager across all processes and another to remove rabbitmq completely after we have had a chance to clear all those messages.

## How to test
A new unit test has been added, the functional and integration tests will be expanded in SFR-2631


etl-pipeline/managers/sqs.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def send_message_to_queue(self, message: Union[str, dict]):
+        queue = self.get_queue()
+        if isinstance(message, dict):
+            message = json.dumps(message)
+
+        params = {"MessageAttributes": message}
+
+        try:
+            response = queue.send_message(**params)
```
kylevillegas93, line 72: We could also pass in the queue url here rather than making a call to get the queue.

etl-pipeline/managers/sqs.py
```diff
@@ -0,0 +1,147 @@
...
+            logger.error(f"Failed to send message to SQS: {e}")
+            # Reconnect and retry once
+            self.create_client()
+            queue = self.get_queue()
+            try:
+                response = queue.send_message(**params)
+                return response
+            except ClientError as e:
+                logger.error(f"Failed retry sending message to SQS: {e}")
+                raise
```
kylevillegas93, line 84: I know we had this logic in the previous rabbitmq implementation but what are the reasons there would be a ClientError thrown?

etl-pipeline/managers/sqs.py
```diff
@@ -0,0 +1,147 @@
...
+    def get_message_from_queue(self):
+        # Try to get messages from rabbitmq first
+        try:
+            self.rabbit_mq.create_connection()
+            self.rabbit_mq.create_or_connect_queue(self.queue_name, self.rabbit_route)
+            message = self.rabbit_mq.get_message_from_queue(self.rabbit_route)
+            if message != None:
+                return message
+        except (StreamLostError, ConnectionError):
+            pass  # Fall through to SQS
```
kylevillegas93, line 95: Let's just migrate to SQS. So we don't need to wrap the rabbitmq stuff here!

etl-pipeline/managers/sqs.py
```diff
@@ -0,0 +1,147 @@
...
+        
+        queue = self.get_queue()
+        try:
+            response = queue.receive_messages(
+                MessageAttributeNames=["All"],
+                MaxNumberOfMessages=self.max_receive_count,
+            )
```
kylevillegas93, line 102: Same as above we might be able to receive from the queue without getting it!

etl-pipeline/tests/unit/test_sqs.py
kylevillegas93, line 1: I think because we are wrapping the boto3 sqs client in the SQS manager it will be useful to write an integration or functional test for this rather than a unit test.

etl-pipeline/managers/sqs.py
```diff
@@ -0,0 +1,147 @@
...
+            self.rabbit_mq.create_or_connect_queue(self.queue_name, self.rabbit_route)
+            self.rabbit_mq.acknowledge_message_processed(message)
+            return
+        try:
+            message.delete()
+        except ClientError as e:
+            logger.error(f"Failed to delete/acknowledge message: {e}")
+            raise
+
+    def reject_message(self, message, requeue=False):
```
kylevillegas93, line 120: We may not need this either!

etl-pipeline/managers/sqs.py
thinkwebengineer: As per the SQS docs [here](https://docs.aws.amazon.com/code-library/latest/ug/python_3_sqs_code_examples.html)

etl-pipeline/managers/sqs.py
```diff
@@ -0,0 +1,109 @@
...
+        if not self.client:
+            self.create_client()
+        try:
+            response = self.client.receive_message(
+                QueueUrl=self.queue_url,
+                WaitTimeSeconds=self.wait_time_seconds,
+                MessageAttributeNames=["All"],
+                MaxNumberOfMessages=self.max_receive_count,
+            )
+            return response["Messages"][0] if "Messages" in response else None
```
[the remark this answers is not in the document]
sarangj: Actually, yea, I think it would make sense to just change this to `get_messages_from_queue`, and always return a list, even if there's just a single message.  Then the consumer can just read multiple messages at a time

PR #706, opened 2025-04-21
Use read replica for API

sarangj:
## Describe your changes
The API *generally* doesn't write data to the DB, so point it at the read replica so that things run smoothly if data ingests are doing a lot of writes.
https://newyorkpubliclibrary.atlassian.net/browse/SFR-2666

## How to test
`python -m main -p APIProcess -e local-{qa|production}` and hit localhost:5050

### DIFF
--- a/etl-pipeline/processes/api.py
+++ b/etl-pipeline/processes/api.py
@@ -1,3 +1,5 @@
+import os
+
 from api.app import FlaskAPI
 from logger import create_log
 from managers import DBManager, ElasticsearchManager, RedisManager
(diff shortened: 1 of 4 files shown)


etl-pipeline/processes/api.py
```diff
@@ -7,7 +7,8 @@
 
 class APIProcess:
     def __init__(self, *args):
-        self.db_manager = DBManager()
+        host = os.environ.get("POSTGRES_READ_HOST")
+        self.db_manager = DBManager(host=host)
```
kylevillegas93, line 11: Thoughts on falling back here to POSTGRES_HOST? Otherwise, let's set it in the other config files as well.

sarangj: This already happens by default in the DBManager!

kylevillegas93 (approved):

PR #707, opened 2025-04-22
Fix: local seed script

kristojorg:
## Describe your changes

- Introduces a healthcheck to the RabbitMQ container in the `docker-compose.setup.yml` and then uses that to make sure that RabbitMQ is ready before running the `drb_local_seed` container by depending on the health check. 
- Increases the HathiTrust TSV field size limit by 8x since I was getting errors about hitting the limit while seeding my DB. A slightly lower limit may work, but I also tried 4x and that didn't work. 

## How to test

- Run the setup process using `docker compose -f docker-compose.setup.yml up --abort-on-container-exit` and ensure the seed completes without errors. You might try wiping local data before doing that so you can be sure it actually added data

### DIFF
--- a/etl-pipeline/docker-compose.setup.yml
+++ b/etl-pipeline/docker-compose.setup.yml
@@ -18,6 +18,11 @@ services:
     extends:
       file: docker-compose.yml
       service: rabbitmq
+    healthcheck:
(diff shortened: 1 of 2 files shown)


kylevillegas93 (approved):
Looks good!

kylevillegas93:
You're good to go to merge these changes in - the LOC test will likely continue failing

PR #711, opened 2025-04-23
NO-REF: Fix python req cache

kylevillegas93:
## Describe your changes
- We weren't getting hits when caching which caused an extra minute per GH test run
- Fixes python dep caching

## How to test
Screenshot 2025-04-23 at 12 03 01

### DIFF
--- a/.github/workflows/etl-pipeline-tests.yaml
+++ b/.github/workflows/etl-pipeline-tests.yaml
@@ -3,19 +3,19 @@ name: ETL Pipeline Tests (Pull Request)
 on:
   pull_request:
     types: [opened, synchronize]
-    # paths:
-    #   - "etl-pipeline/api/**"
-    #   - "etl-pipeline/managers/**"
-    #   - "etl-pipeline/mappings/**"
-    #   - "etl-pipeline/model/**"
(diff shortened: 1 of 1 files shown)


.github/workflows/etl-pipeline-tests.yaml
```diff
@@ -39,13 +46,6 @@ jobs:
             etl-pipeline/requirements.txt
             etl-pipeline/dev-requirements.txt
 
-      - name: Install build essentials
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y gcc-12 g++-12
-          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-12 60
-          sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-12 60
-
```
mitri-slory, line 48: I wonder why moving the placement of this code improves the test time.

mitri-slory (approved):

.github/workflows/etl-pipeline-tests.yaml
sarangj: I'm assuming because the c packages that are installed here weren't getting cached (since this was happening after the pip install step which checks the cache.

PR #712, opened 2025-04-23
Write record ingest messages to SQS

sarangj:
## Describe your changes
To do a final check that the queue is configured properly in AWS, start
writing messages to SQS, so we can validate they can be written and the
message format is as expected. Once this is validated, we can purge the
queue, update the record pipeline to read from the queue, and remove the
RabbitMQ step

## How to test

Run `make functional`, then observe the messages in the queue in the localstack UI

### DIFF
--- a/etl-pipeline/processes/record_ingestor.py
+++ b/etl-pipeline/processes/record_ingestor.py
@@ -3,7 +3,7 @@ import os
 import typing
 
 from logger import create_log
-from managers import RabbitMQManager
+from managers import RabbitMQManager, SQSManager
 from model import Record
 
 logger = create_log(__name__)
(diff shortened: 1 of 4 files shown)


etl-pipeline/processes/record_ingestor.py
```diff
@@ -17,26 +17,35 @@ def __init__(self, source: str):
         self.records_queue = os.environ['RECORD_PIPELINE_QUEUE']
         self.records_route = os.environ['RECORD_PIPELINE_ROUTING_KEY']
```
sarangj, line 18: remove

mitri-slory (approved):
The code looks good. This is the link to my current log in the localstack after running the functional tests: [https://privatebin.nypl.org/?e327542a0d3a6254#mEMztsql9QlpXBtmgGVk1H6gxCT5VWJ1t2yfb83wA0Y=](url)

PR #713, opened 2025-04-23
Migrate record pipeline to SQS

sarangj:
Now that we've validated that the ingestor can successfully write to SQS, migrate the pipeline to read off the new queue.


kylevillegas93 (approved):
looks good !

PR #714, opened 2025-04-23
NO-REF: Implementing RedisLock for record clustering

kylevillegas93:
## Describe your changes
- basically, this change tries to skip clustering a record if it's already being clustered by another consumer
- uses redlock from pottery to lock a record id :)
- checks redis to see if any of the record ids that are part of the candidate pool are already locked 
- skips the record clustering step if it's there's overlap with another consumer

## How to test
`make up; make functional`


etl-pipeline/managers/redis.py
```diff
@@ -88,6 +88,11 @@ def set_key(
             ex=expiration_time,
         )
 
+    def any_locked(self, keys: list) -> bool:
+        keys = [f'redlock:{key}' for key in keys]
+        
+        return any(locks is not None for locks in self.client.mget(keys))
```
sarangj, line 94: could return the locked keys and log them?

etl-pipeline/processes/record_clusterer.py
```diff
@@ -40,26 +45,34 @@ def __init__(self, db_manager: DBManager):
...
+
         self.constants = get_constants()
 
     def cluster_record(self, record) -> list[Record]:
         """Clusters a single record and updates the database and Elasticsearch.
         """
         try:
-            work, stale_work_ids, records = self._get_clustered_work_and_records(record)
-            self._commit_changes()
+            record_lock = Redlock(key=str(record.id), masters={self.redis_manager.client}, auto_release_time=self.CLUSTER_TIMEOUT)
```
sarangj, line 56: would it make sense to add a prefix to the key, ie 'cluster_lock_{record_id}'. Just so there are no conflicts if we add more locks in the future?

sarangj (approved):
Consider my comments optional / potential TODOs, this looks good!

etl-pipeline/managers/redis.py
kylevillegas93: we could - but i think we want to return as quickly as possible here!

etl-pipeline/processes/record_clusterer.py
kylevillegas93: hmm yes !

etl-pipeline/processes/record_clusterer.py
```diff
@@ -30,8 +34,10 @@ class RecordClusterer:
     CLUSTER_SIZE_LIMIT = 10000
     # Regular expression to identify identifiers that should be matched
     IDENTIFIERS_TO_MATCH = r"\|(?:isbn|issn|oclc|lccn|owi)$"
+    CLUSTER_TIMEOUT = 60 * 60 # 1 hour
```
kristojorg, line 37: Just checking - this is the timeout to cluster a single record?? Seems like if we are reaching anything near that then we probably selected way too many records in the candidate pool identification step

kristojorg (approved):
Looks good! One question though - could we also implement this as a DB flag? For example changing the `cluster_status` to a string enum that can either be `unclustered | clustering | clustered` ? I don't have a problem with the redlock method but just curious if there is a reason not to do it that way?

mitri-slory (approved):
This looks good to me and hopefully it helps with speeding up the clustering process.

etl-pipeline/processes/record_clusterer.py
kylevillegas93: ya i think what we were seeing was a record with 2k+ records that represent the work

sarangj:
@kristojorg re: redlock, one of the issues I *think* was that the db was locking record rows to do some of the updates during clustering, so I think using the DB might be a no-go (correct me if I'm wrong @kylevillegas93 )

PR #716, opened 2025-04-24
Add `state` column to record table

sarangj:
This column will eventually supplant cluster_status and frbr_status, and the new `ingested` state will allow us to persist records during ingested to reduce SQS payload size

## How to test
ran `alembic upgrade head` locally, checked DB


etl-pipeline/migrations/versions/fabf5e5c3109_add_collections_table.py
```diff
@@ -50,7 +51,8 @@ def upgrade():
             sa.Integer,
             sa.ForeignKey('editions.id', ondelete='CASCADE'),
             index=True
-        )
+        ),
+        if_not_exists=True,
```
sarangj, line 55: remove

PR #717, opened 2025-04-24
Add `state` column to record table

sarangj:
This column will eventually supplant cluster_status and frbr_status, and the new ingested state will allow us to persist records during ingested to reduce SQS payload size

## How to test
`alembic upgrade head`, look at DB

### DIFF
--- a/etl-pipeline/migrations/versions/eb6ce7780436_add_record_state.py
+++ b/etl-pipeline/migrations/versions/eb6ce7780436_add_record_state.py
@@ -1,0 +1,37 @@
+"""Add record state
+
+Revision ID: eb6ce7780436
+Revises: 0cebb9a98a6f
+Create Date: 2025-04-24 11:09:19.838900
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import ENUM
+
+
+# revision identifiers, used by Alembic.
+revision = 'eb6ce7780436'
+down_revision = '0cebb9a98a6f'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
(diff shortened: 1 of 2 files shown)


etl-pipeline/model/postgres/record.py
```diff
@@ -114,6 +114,15 @@ class Record(Base, Core):
         index=True
     )
     cluster_status = Column(Boolean, default=False, nullable=False, index=True)
+    state = Column(
+        Unicode,
+        ENUM(
+            "ingested", "files_saved", "complete",
```
kylevillegas93, line 120: let's do ingested, files_saved, embellished, and clustered

etl-pipeline/model/postgres/record.py
```diff
@@ -114,6 +114,15 @@ class Record(Base, Core):
...
+            name="record_state", create_type=False,
+        ),
+        nullable=True,
+        index=True,
```
kylevillegas93, line 124: i think in the past we have failed to create indices on the database because there's too much going on :| 

what are your thoughts on not making this an index to start and then adding the index later?

etl-pipeline/migrations/versions/eb6ce7780436_add_record_state.py
```diff
@@ -0,0 +1,37 @@
...
+            "state",
+            sa.String,
+            nullable=True,
+            index=True,
+        ),
+    )
+
+
+def downgrade():
+    pass
```
kylevillegas93, line 37: Do we want to add the delete_column code here just in case?

kylevillegas93 (approved):

mitri-slory (approved):
This looks good to go.

PR #719, opened 2025-04-24
SFR-2678: Persist records and set state in ingestor

sarangj:
when ingesting records, persist them with a new `state` value of `ingested`, and then send a smaller message to SQS, just containing the record id to stay under the SQS message size limit.

Note, I adapted the record buffer a bit here so that the ingestor could consume the buffered records as they get flushed out - I *could* instead replicate / extract the upsert logic in the record buffer into the ingestor, but this way was fastest.

## How to test
Run the loc ingest `python -m main -p LOCProcess -i complete -l 5`, observe the new records in the DB with the correct state, and the new messages in localstack SQS

### DIFF
--- a/etl-pipeline/model/postgres/record.py
+++ b/etl-pipeline/model/postgres/record.py
@@ -92,7 +92,7 @@ class FRBRStatus(Enum):
 class RecordState(Enum):
     INGESTED = 'ingested'
     FILES_SAVED = 'files_saved'
-    EMBELLISED = 'embellished'
+    EMBELLISHED = 'embellished'
     CLUSTERED = 'clustered'
 
 
--- a/etl-pipeline/model/postgres/record.py
+++ b/etl-pipeline/model/postgres/record.py
@@ -114,6 +114,15 @@ class Record(Base, Core):
         index=True
     )
     cluster_status = Column(Boolean, default=False, nullable=False, index=True)
+    state = Column(
+        Unicode,
+        ENUM(
+            "ingested", "files_saved", "complete",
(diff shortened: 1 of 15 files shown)


sarangj:
I could also just commit the records 1 by 1, just didn't wanna execute a ton of transactions if we do a backfill or something

etl-pipeline/model/postgres/record.py
```diff
@@ -89,6 +89,13 @@ class FRBRStatus(Enum):
     COMPLETE = 'complete'
 
 
+class RecordState(Enum):
+    INGESTED = 'ingested'
+    FILES_SAVED = 'files_saved'
+    EMBELLISED = 'embellished'
```
sarangj, line 95: nit: EMBELLISHED?

etl-pipeline/processes/record_buffer.py
```diff
@@ -23,22 +25,16 @@ def add(self, record: Record):
...
-            self.deletion_count += 1
-            self.db_manager.session.commit()
+        return None
 
-    def flush(self):
+    def flush(self) -> Iterable[Record]:
         self.db_manager.bulk_save_objects(self.records)
         self.ingest_count += len(self.records)
+
+        yield iter(self.records)
```
sarangj, line 36: This this needs a `from`

etl-pipeline/model/postgres/record.py
kylevillegas93: oop

kylevillegas93 (approved):

PR #720, opened 2025-04-24
SFR-2679: Add OWI number to all records

mitri-slory:
## Describe your changes
This PR intends to add an OWI number to the identifiers attribute of all records during the embellishment process. This should help increase the success rate and accuracy of our candidate pool generation which thereby should increase cluster rates. Right now it's set up to where all the owi numbers present in the oclc bibs will be added to the records however I don't know if it suffices to just add the first instance of an owi number to a record which I would appreciate feedback on.

## How to test
-`make up` 
-running one of the processes locally 
-checking for OWI numbers in records table of local PSQL database


etl-pipeline/processes/record_embellisher.py
```diff
@@ -58,13 +58,22 @@ def _add_works_for_record(self, record: Record):
...
         if self.record_buffer.ingest_count == 0 and len(self.record_buffer.records) == 0 and author and title:
             search_query = self.oclc_catalog_manager.generate_search_query(author=author, title=title)
+            oclc_bibs = self.oclc_catalog_manager.query_bibs(query=search_query)
+            self._add_owi_to_record(record, oclc_bibs)
             self._add_works(self.oclc_catalog_manager.query_bibs(query=search_query))
 
+    def _add_owi_to_record(self, record: Record, oclc_bibs: dict):
+        for oclc_bib in oclc_bibs:
+            owi_number = oclc_bib.get('work', {}).get('id')
+            record.identifiers.append(f'{owi_number}|owi')
```
kylevillegas93, line 75: Looking good! 

We may want to map first the owis from the oclc bibs - filtering out any that are None! May make sense to use a set here to dedupe the owis! 

What are your thoughts?

mitri-slory: I completely agree with your suggestions and I'll implement them in the next commit.

kylevillegas93 (approved):
Looks good - you may just want to modify the test to ensure that the owi number is being added !

kristojorg (approved):
Looks good! Just the one question. I also would love to hold off on merging this until we merge #729 and get one round of analytics to see how this changes our clustering

etl-pipeline/processes/record_embellisher.py
```diff
@@ -31,13 +31,14 @@ def __init__(self, db_manager: DBManager, redis_manager: RedisManager):
         self.record_buffer = RecordBuffer(db_manager=self.db_manager)
 
     def embellish_record(self, record: Record) -> Record:
-        self._add_works_for_record(record=record)
+        record_new_identifiers_owi = self._add_works_for_record(record=record)
```
kylevillegas93, line 34: based our convo - let's return owi_identifiers here

etl-pipeline/processes/record_embellisher.py
```diff
@@ -31,13 +31,14 @@ def __init__(self, db_manager: DBManager, redis_manager: RedisManager):
...
 
         self.record_buffer.flush()
 
         # TODO: deprecate frbr_status
         record.frbr_status = 'complete'
         record.state = RecordState.EMBELLISHED.value
+        record.identifiers = record_new_identifiers_owi
```
kylevillegas93, line 41: here let's extend the identifiers with the owi_identifiers from above!

mitri-slory:
The changes here are reflected in this PR: [https://github.com/NYPL/digital-research-books/pull/739](url).

PR #724, opened 2025-04-25
NO-REF: Add SQS config

kylevillegas93:
## Describe your changes
- Adds SQS names to the yaml configs


sarangj:
Hmmm, I added these to the ECS task definitions, so *shouldn't* need this, but no harm!

sarangj (approved):

PR #729, opened 2025-04-28
Sfr 2635/cluster metrics

kristojorg:
## Describe your changes

Adds clustering metrics according to [this spec](https://docs.google.com/document/d/1fzfAzKfMlFEegdc4lAie-WMfRSjn5HoXX-TyOysOX3Y/edit?usp=sharing).

## How to test

- Set a `NEW_RELIC_LICENSE_KEY` in your shell env. DM me if you need one.
- Disable your adblocker! This tripped me up.
- Run the LOC ingest process to get some records into your queue: `python main.py -e local -p LOCProcess -i complete -l 5`
- Run the `RecordPipelineProcess`: `python main.py -e local -p RecordPipelineProcess -i complete` to trigger clustering
- Stop the process with control + c. This will make sure events are flushed to NR.
- Check the NR console using a query like 
```FROM `Cluster:WorkRecordsChosen` select *```

### CODE REVIEW

File: etl-pipeline/processes/candidate_record_finder.py [hunk 1]
```diff
@@ -3,7 +3,7 @@
 from typing import List, Set, Tuple, Optional, Any
 from sqlalchemy.exc import DataError
 from managers import DBManager, RedisManager
-
+import services.monitor as Monitor
```
sarangj, line 6: I'd import this `as monitor`, modules (and module aliases) are typically lower cased in python!

etl-pipeline/test_newrelic.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
+"""
+Test script for New Relic custom events
+"""
+import os
+import sys
+import time
+import newrelic.agent
+
+def main():
```
sarangj, line 10: I think we should throw this under `scripts`!

sarangj (commented):
looks good, just a couple small suggestions!

etl-pipeline/test_newrelic.py
kristojorg: Oh whoops I forgot to remove this. Do you think we want to include it?

sarangj (approved):

sarangj (changes requested):
One thing I noticed when testing my changes that are based on this branch!

PR #733, opened 2025-04-29
Add RecordPipeline success / failure events 

sarangj:
## Describe your changes
Building off @kristojorg's monitoring work, add a couple more events for
tracking successful and failed messages in the RecordPipeline.

I'm adding processing time separately - I'm still evaluating whether I should just add that to these events, set up a custom metric (which makes less and less sense the more I look into it) or use newrelic's provided function tracing (which seems to behave weirdly, but I'm looking into it).

https://newyorkpubliclibrary.atlassian.net/browse/SFR-2683

## How to test
Run the LOC process, run the record pipeline, and then ctrl-c to exit the pipeline to see the new events


sarangj:
I *only* added execution time to success events, but looks like you can query multiple events at once, so if I added the same attribute to failed message events, we could do a query like 
```
FROM `RecordPipeline:MessageSucceeded`, `RecordPipeline:MessageFailed`  SELECT max(execution_time)
```
to get a picture of execution time across successes / failures (or just query each event individually to get a picture of each specific case)

kylevillegas93 (approved):
Looks good to me :)

kylevillegas93:
I think just merge in with main and I think you should be good to go with the tests!

sarangj:
Kinda clear that it's flakiness, I'm gonna merge

PR #734, opened 2025-04-30
NO-REF: Add record state filter when generating candidate pool

kylevillegas93:
## Describe your changes
- We should only try to create candidate records if the state is past the ingestion step

## How to test
`make functional; make integration`


sarangj (approved):

kylevillegas93:
Bloop - ill update the tests!

PR #738, opened 2025-04-30
SFR-2662: Add GRIN oauth functionality

priya-pai:
## Describe your changes
GRIN's "API" currently uses a manual login flow in order to be accessed. However, when we make GRIN conversion/downloads their own ECS tasks, we'll want an oauth flow that is programmatic and specifically uses our service account. This PR makes it so that we use the GRIN service account's credentials in a traditional Google oauth flow. Note that I've added the GRIN service account credentials (as a JSON blob) to Parameter Store. 

Also, I would like to move some of this logic into a service soon, but I don't want to create that prematurely until I know what the conversion/download script will need. For now, I think it's ok for the code to live here.

## How to test
Run the grin-mirror script in a venv by doing: `python3 -m scripts.grin_mirror`
I would also add a `break` on line 103 so that only one file is downloaded from GRIN. One note is that your IAM user needs to have GetParameter access in their IAM policy. I had to create a ticket for DevOps to do so for me, but in the meantime decided to use the QA credentials.


Toxiapo (approved):
Tested locally, works great!

PR #739, opened 2025-04-30
NO-REF: Simplify embellish process

kylevillegas93:
## Describe your changes
- Adds owi number to embellished record
- Adds oec to catalog records
- Adds complete mapping from oclc bib to dc records
- Simplifies query

## How to test
`make up; make functional`


mitri-slory (approved):

etl-pipeline/mappings/oclc_bib.py
```diff
@@ -2,130 +2,189 @@
...
+
+
+def _get_parts(oclc_bib):
+    digital_access_locations = oclc_bib.get('digitalAccessAndLocations')
+
+    if not digital_access_locations:
+        return None
+    
+    return [
+        str(Part(index=1, url=uri, source='oclc', file_type='text/html', flags=str(FileFlags(embed=True))))
```
sarangj, line 100: Should these parts all have different indices?

etl-pipeline/mappings/oclc_bib.py
```diff
@@ -2,130 +2,189 @@
...
+            creators
+        )
+    )
+
+def _is_author(creator):
+    for role in set(map(lambda relator: relator.get('term', '').lower(), creator.get('relators', []))):
+        if 'author' in role.lower() or 'writer' in role.lower():
+            return True
+    
+    return False
```
sarangj, line 144: Might consider using the `any` function here (personal pref!).  And wrapping in a `set` is probably unnecessary since you're just short circuiting once you find a match?

sarangj (approved):
couple optional comments!

etl-pipeline/mappings/oclc_bib.py
kylevillegas93: i think so for the post part - they are all links to the same bib just maybe from different sources.

etl-pipeline/mappings/oclc_bib.py
kylevillegas93: bloop - ya this was when my python skills were not up to par haaaa - i will circle back to this :D

PR #740, opened 2025-04-30
SFR-2689: Create grin_status table

Toxiapo:
## Describe your changes

Define a `grin_status` table that tracks the status of GRIN records. Each record in `grin_status` will correspond to a record in the existing `record` table via a foreign key relationship using the `record.id`. The `grin_status` table will have a `barcode` column (Primary key) which stores the GRIN barcode. The existing `record` table should remain unchanged.

## How to test
Run `alembic upgrade head` to see the new table created in your local db


priya-pai (approved):

kylevillegas93 (approved):
this looks good to me

PR #742, opened 2025-05-05
SFR-2695/96: Add gtag tracking for Read Online and Download PDF buttons

jackiequach:
[SFR-2695](https://newyorkpubliclibrary.atlassian.net/browse/SFR-2695)
[SFR-2696](https://newyorkpubliclibrary.atlassian.net/browse/SFR-2696)

## Describe your changes

- Add Google Analytics tracking for Read Online and Download PDF buttons 

Note: It may be better to use React context instead of prop drilling to eliminate the need to pass the author(s) prop unnecessarily. This method was chosen due to simplicity and we are likely reworking the frontend in the near future.

## How to test
- This can be verified by Miguel through Google Tag Manager as per the Jira tickets.

[SFR-2695]: https://newyorkpubliclibrary.atlassian.net/browse/SFR-2695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SFR-2696]: https://newyorkpubliclibrary.atlassian.net/browse/SFR-2696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/src/components/CollectionItemCard/DownloadLink.tsx
```diff
@@ -5,8 +5,10 @@ import Link from "~/src/components/Link/Link";
 import { OpdsLink } from "~/src/types/OpdsModel";
 import { formatUrl } from "~/src/util/Util";
 import { trackCtaClick } from "~/src/lib/adobe/Analytics";
+import { trackEvent } from "~/src/lib/gtag/Analytics";
 
-const DownloadLink: React.FC<{ links: OpdsLink[]; title: string }> = ({
+const DownloadLink: React.FC<{ author: string; links: OpdsLink[]; title: string }> = ({
```
kylevillegas93, line 10: The author may be none here - do we want to make the explicit?

web/src/components/CollectionItemCard/ReadOnlineLink.tsx
```diff
@@ -3,9 +3,11 @@ import { OpdsLink } from "~/src/types/OpdsModel";
 import CollectionUtils from "~/src/util/CollectionUtils";
 import { formatUrl } from "~/src/util/Util";
 import Link from "~/src/components/Link/Link";
+import { trackEvent } from "~/src/lib/gtag/Analytics";
 
 // "Read Online" button should only show up if the link was flagged as "reader" or "embed"
-const ReadOnlineLink: React.FC<{ links: OpdsLink[]; title: string }> = ({
+const ReadOnlineLink: React.FC<{  author: string; links: OpdsLink[]; title: string; }> = ({
```
kylevillegas93, line 9: same as above

kylevillegas93 (changes requested):
Overall, it's looking good! Great work - just want to double check something on the types.

web/src/components/CollectionItemCard/ReadOnlineLink.tsx
```diff
@@ -17,13 +19,23 @@ const ReadOnlineLink: React.FC<{ links: OpdsLink[]; title: string }> = ({
 
   if (!readOnlineLink) return null;
 
+  const trackReadOnlineClick = () => {
+    trackEvent({
+      "event":  "digital_read_online",
+      "item_title": title,
+      "item_author": author,
```
Toxiapo, line 26: I wonder if it would makes more sense to enforce this to a specific type (make `item_author` either an array or a string)?  Might be easier to process in the dashboard consider GA might treat the two datatypes differently, but please ignore if you think this is not necessary!

jackiequach: Miguel said that both values, either array or string, are valid for item_author.

Toxiapo (approved):

PR #743, opened 2025-05-06
SFR-2685: Add initial scrape script

Toxiapo:
## Describe your changes
Adds a new script to scrape all currently available books from GRIN and mark them to be ready for conversion, and ready to be consumed by our daily GRIN process.

#### Open question:
1. Where should we run this script?  
   - I am thinking ECS but maybe there's a better place for these one-time scripts?

## How to test

`python -m scripts.GRIN_initial_scrape` , ensure you have AWS access and have already migrated to the latest DB schema from the previous PR.


etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+        host=os.environ.get("POSTGRES_HOST", None),
+        port=os.environ.get("POSTGRES_PORT", None),
+        db=os.environ.get("POSTGRES_NAME", None),
+    )
+    db_manager.create_session()
+
+    url = f"{args.url}/_all_books?book_state=NEW&book_state=PREVIOUSLY_DOWNLOADED&format=text"
+    logging.info(f"Scrapping url: {url}")
+
+    response = authed_session.get(url, timeout=600)
```
Toxiapo, line 72: I don't think the request will go beyond 600 seconds, average about 2minute.

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+    logging.info(f"Processing {len(barcodes)} barcodes")
+
+    for chuncked_barcodes in chunk(iter(barcodes), chunk_size):
+        new_records: List[Record] = []
+        for barcode in chuncked_barcodes:
+            _barcode_id = f"NYPL_{barcode}"
+            new_records.append(
+                Record(
+                    uuid=uuid4(),
+                    frbr_status=FRBRStatus.TODO.value,
```
Toxiapo, line 92: These two fields are required, I am using `to_do` as the default value but please LMK if I shouldn't be using this.

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+                    source_id=_barcode_id,
+                    grin_status=GRINStatus(
+                        barcode=_barcode_id,
+                        failed_download=0,
+                        state=GRINState.PENDING_CONVERSION.value, 
+                        date_created=datetime(1991, 8, 25)
```
Toxiapo, line 98: Using a very old date to mark these as _historical_ files

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+        creds = Credentials.from_service_account_info(
+            service_account_info, scopes=scopes)
+        return creds
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--url", default=DEFAULT_BASE_URL)
+    parser.add_argument("--size", default=DEFAULT_CHUNK_SIZE)
+    args = parser.parse_args()
```
Toxiapo, line 55: Not necessary, but thought it would be useful

etl-pipeline/scripts/__init__.py
```diff
@@ -15,6 +15,5 @@
 from .nyplLoginFlags import main as nyplFlags
 from .deleteUMPManifestLinks import main as deleteUMPManifests
 from .deleteProblemWorks import main as deleteWorks
-from .check_queue_size import main as checkQueueSize
```
priya-pai, line 18: Thanks for taking care of this!

etl-pipeline/scripts/GRIN_initial_scrape.py
priya-pai: I like the idea to make chunk_size configurable! But I am curious about the decision to make the URL configurable. When would the base URL change for this script, since the URL params are pretty specific to GRIN?

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+        user=os.environ.get("POSTGRES_USER", None),
+        pswd=os.environ.get("POSTGRES_PSWD", None),
+        host=os.environ.get("POSTGRES_HOST", None),
+        port=os.environ.get("POSTGRES_PORT", None),
+        db=os.environ.get("POSTGRES_NAME", None),
+    )
+    db_manager.create_session()
+
+    url = f"{args.url}/_all_books?book_state=NEW&book_state=PREVIOUSLY_DOWNLOADED&format=text"
+    logging.info(f"Scrapping url: {url}")
```
priya-pai, line 70:
```suggestion
    logging.info(f"Scraping url: {url}")
```

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+
+    response = authed_session.get(url, timeout=600)
+    response.raise_for_status()
+   
+    barcodes = response.content.decode("utf8").strip().split("\n")
+
+    if len(barcodes) > 0: 
+        insert_into_db(barcodes=barcodes, db_manager=db_manager, chunk_size=args.size)
+    else:
+        logging.info("Zero record found")
```
priya-pai, line 80:
```suggestion
        logging.info("No records found")
```

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+
+def insert_into_db(barcodes: List[str], db_manager: DBManager, chunk_size: int):
+    logging.info(f"Processing {len(barcodes)} barcodes")
+
+    for chuncked_barcodes in chunk(iter(barcodes), chunk_size):
```
priya-pai, line 85:
```suggestion
    for chunked_barcodes in chunk(iter(barcodes), chunk_size):
```

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+        new_records: List[Record] = []
+        for barcode in chuncked_barcodes:
+            _barcode_id = f"NYPL_{barcode}"
```
priya-pai, line 88: Looking at the source IDs in the records table, they seem to be source-specific - maybe we want this to be "{barcode}|grin" based off of how other sources are set up?

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+                        state=GRINState.PENDING_CONVERSION.value, 
+                        date_created=datetime(1991, 8, 25)
+                    )
+                )
+            )
+            
+        logging.info(f"Inserting {len(chuncked_barcodes)} barcodes into Record")
+
+        try:
+            db_manager.session.add_all(new_records)
```
priya-pai, line 106: Instead of `add_all`, I think we should use `bulk_save_objects`. It looks like `add_all` will execute an INSERT statement per object being inserted. Rather I think we want a single (or at least fewer) INSERT statement for multiple entries. I found this to be a [good read](https://stackoverflow.com/questions/58517491/sqlalchemy-bulk-save-objects-vs-add-all-underlying-logic-difference) to discuss what's happening under the hood. LMK what you think!

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+            db_manager.commit_changes()
+        except Exception:
+            logging.exception('Failed to add records')
+            raise
```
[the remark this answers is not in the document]
priya-pai: I don't think we want to roll back everything, but I do think we should log the barcodes of the records that fail to be inserted. Maybe we could do a txt file of the failed barcodes?

etl-pipeline/scripts/GRIN_initial_scrape.py
Toxiapo: ohh, checking now!

etl-pipeline/scripts/GRIN_initial_scrape.py
Toxiapo: For other sources that also hosted on GRIN in the future, e.g.  "https://books.google.com/libraries/other", but it's not needed for this!

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+DEFAULT_BASE_URL = "https://books.google.com/libraries/NYPL"
+DEFAULT_CHUNK_SIZE = 5000
+
+logging.basicConfig(
+    filename="GRIN_initial_scrape.log",
+    level=logging.INFO,  
+    format='%(asctime)s - %(levelname)s - %(message)s' 
+)
```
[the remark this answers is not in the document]
kylevillegas93: Feel free to update the main.py file to wrap a with clause around the scripts that are run!

etl-pipeline/scripts/GRIN_initial_scrape.py
```diff
@@ -0,0 +1,133 @@
...
+
+            break
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    except Exception as e:
+        logging.exception(e, exc_info=True)
+
```
kylevillegas93, line 133: oop - let's add a new line :)

etl-pipeline/scripts/__init__.py
kylevillegas93: yay

etl-pipeline/scripts/GRIN_initial_scrape.py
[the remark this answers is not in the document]
Toxiapo: Still learning new python~ so I might need a bit of help getting what you mean here!

etl-pipeline/scripts/GRIN_initial_scrape.py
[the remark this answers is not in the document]
Toxiapo: I am thinking more of in terms of inserting a duplicate barcodes but maybe I am overthinking this! 🤔

etl-pipeline/scripts/GRIN_initial_scrape.py
Toxiapo: I am going to redo the `ArgumentParser` and just hardcode these values because it was conflicting the run command when running the script as part of the process in ECS

priya-pai (approved):
This looks good to me - the only thing I'm unsure about is the main.py logic, but I think Kyle is a better person to check in with about that!

Toxiapo:
> This looks good to me - the only thing I'm unsure about is the main.py logic, but I think Kyle is a better person to check in with about that!

I removed that changes, we can figure that out later if we need it or not!

PR #754, opened 2025-05-13
SFR-2692: Download process

Toxiapo:
## Describe your changes
This PR handles the GRIN download process which includes

1. Getting all the converted books from DB
2. Download the books from GRIN
3. Upload downloaded books to the S3 Glacier bucket
4. Mark the downloaded book as DOWNLOADED

It handles the download of recently converted books, ordered by the modified date, and the ability to run a backfill process separately.

## How to test

Running `python -m services.grin.download` will execute the full process as long as you have a valid record with CONVERTED state in your local database.

@kylevillegas93 @sarangj I was getting the error `ERROR - Unable to store file grin/test.tar.gz.gpg in s3: An error occurred (AccessControlListNotSupported) when calling the PutObject operation: The bucket does not allow ACLs` when calling the S3Manager.put_object method, the error seem to go away if I put `bucket_permissions` as None, should this something we should concern about? Or was there's a bucket setting somewhere? (`drb-grin-files-test` is the bucket I created for test that triggered this error.)


etl-pipeline/services/sources/grin/download.py
```diff
@@ -0,0 +1,84 @@
...
+                    self.s3_client.put_object(
+                        object=content, s3_key=s3_key, bucket=S3_BUCKET_NAME
+                    )
+                    self._update_state(book, GRINState.DOWNLOADED.value)
+                except Exception as e:
+                    logger.exception(f"Error uploading to s3 for {book}")
+            except:
+                logger.exception(f"Error downloading content for {book}")
+                book.grin_status.failed_download += 1
+                self.db_manager.commit_changes()
```
priya-pai, line 44: What do you think about wrapping this specific try/except only around line 32 `content = self._download(file_name)`? That way the error downloading the book is separate from the error uploading to s3?

etl-pipeline/services/sources/grin/download.py
```diff
@@ -0,0 +1,84 @@
...
+    def run_process(self, batch_size=BATCH_SIZE_LIMIT, backfill=False):
+        books: List[Record] = self._get_converted_books(batch_size, backfill)
+
+        for book in books:
+            try:
+                barcode = book.source_id.split("|")[0]
+                file_name = f"{barcode}.tar.gz.gpg"
+                s3_key = f"grin/{file_name}"
+
+                content = self._download(file_name)
```
priya-pai, line 32: Right now, I believe there won't be an exception that is thrown when the file can't be found in GRIN. A 404 response will be returned. I think a successful download is only when the response status code is 200, so maybe we should incorporate that logic? But let me know if I'm missing something!

etl-pipeline/services/sources/grin/download.py
```diff
@@ -0,0 +1,84 @@
...
+        response = self.client.download(file_name)
+        return response
+
+    def _get_converted_books(self, batch_size: int, backfill: bool):
+        """Should return the DB object so that we can update the state directly"""
+        query = (
+            select(Record)
+            .join(Record.grin_status)
+            .where(GRINStatus.state != GRINState.DOWNLOADED.value)
+            .where(GRINStatus.state == GRINState.CONVERTED.value)
```
priya-pai, line 66: Do we need both `where` statements? It should only be if the state is equal to CONVERTED, right? That would filter out the DOWNLOADED books anyways

etl-pipeline/services/sources/grin/download.py
```diff
@@ -0,0 +1,84 @@
...
+
+class GRINDownload:
+    def __init__(self):
+        self.s3_client = S3Manager()
+        self.db_manager = DBManager()
+        self.db_manager.create_session()
+
+        self.client = GRINClient()
+
+    def run_process(self, batch_size=BATCH_SIZE_LIMIT, backfill=False):
```
priya-pai, line 23: I like this backfill variable idea!

etl-pipeline/services/grin/download.py
```diff
@@ -0,0 +1,92 @@
...
+
+        if len(books_done_processing) > 0:
+            self._update_states(books_done_processing, GRINState.DOWNLOADED.value)
+
+        self.db_manager.session.close()
+
+    def _update_states(self, books: List[Record], state: str):
+        try:
+            for book in books:
+                book.grin_status.state = state
```
priya-pai, line 59: You may want to set the state upon the first iteration - so in line 48 (when download and upload have both succeeded), and then you can just do a bulk save on the records. That way, you don't have to iterate over the books twice!

Toxiapo: I like that!

etl-pipeline/services/grin/download.py
```diff
@@ -0,0 +1,98 @@
...
+            select(Record)
+            .join(Record.grin_status)
+            .where(GRINStatus.state == GRINState.CONVERTED.value)
+            .where(GRINStatus.date_modified >= yesterday)
+            .order_by(desc(Record.date_modified))
+        )
+
+        if backfill:
+            query = query.where(GRINStatus.date_created <= datetime(1991, 8, 25))
+            query = query.limit(batch_size)
```
Toxiapo, line 86: I am using the same batch_size variable for the query limit for backfills, but lmk if this should be on it's own variable! e.g. `backfll_batch_size` @priya-pai

priya-pai (approved):

PR #755, opened 2025-05-13
SFR-2690: Add an optional "storage_class" argument to the put_object method

Toxiapo:
## Describe your changes
Adds an optional "storage_class" argument to the `put_object` method so that the GRIN script can choose a different storage class option (glacier IR)

## How to test
Run the unit test `make unit`


priya-pai (approved):
Nice 🙌🏽

PR #756, opened 2025-05-13
SFR-2687 Initialize conversion for new books

priya-pai:
## Describe your changes
This PR initializes conversion for all new GRIN books acquired in the past day. I changed the flow in the conversion script to kick off conversion immediately for all new books. I also refactored the `please_process` method to be `convert` and added some batch processing since there is around a 1000 batch size limit for conversion in GRIN.

### DIFF
--- a/etl-pipeline/services/grin/grin_client.py
+++ b/etl-pipeline/services/grin/grin_client.py
@@ -12,6 +12,7 @@ import json
 from ..ssm_service import SSMService
 from pdb import set_trace
 
+BATCH_LIMIT = 1000
 
 class GRINClient(object):
     def __init__(self):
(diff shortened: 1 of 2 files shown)


etl-pipeline/services/grin/grin_client.py
```diff
@@ -44,6 +46,35 @@ def get(self, fragment):
...
+        # "Success", "Already being converted" "Other error", "Not allowed to be downloaded"
+        response = []
+
+        if len(barcodes) >= BATCH_LIMIT:
+            chunked_barcodes = [
+                barcodes[i : i + BATCH_LIMIT]
+                for i in range(0, len(barcodes), BATCH_LIMIT)
+            ]
+        else:
+            chunked_barcodes = [barcodes]
```
Toxiapo, line 62: hmm, should the batch limit handled by the consuming script? I feel like we should keep these client methods small and pure.

Toxiapo: I think this is fine for now!

Toxiapo (approved):
Looks good!

PR #758, opened 2025-05-14
SFR-2702 Initiate conversion for backfilled books

priya-pai:
## Describe your changes
This PR adds converting a batch of backfilled books to the daily conversion script. Specifically, it grabs a batch of relevant backfilled books, kicks off conversion for those, and updates the records in the database.

## How to test
You can test this by having some barcodes that are pending_conversion in your DB with a date of 8-24-1991. Comment out line 26 and 28 to isolate the backfill functionality. Run the file and you can check that conversion is kicked off in GRIN for those barcodes and the DB has been updated accordingly.

### DIFF
--- a/etl-pipeline/services/grin/grin_client.py
+++ b/etl-pipeline/services/grin/grin_client.py
@@ -93,30 +93,23 @@ class GRINClient(object):
         return data
 
     def _for_state(self, state, *args, **kwargs):
-        # Which books are in the given state?
         data = self.get("_%s?format=text" % state, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


etl-pipeline/services/grin/grin_client.py
```diff
@@ -93,30 +93,23 @@ def acquired_today(self, *args, **kwargs):
         return data
```
priya-pai, line 94: Thought I would clean up some of these repetitive comments :)

Toxiapo (approved):

PR #759, opened 2025-05-15
NO-REF: Remove file extension

Toxiapo:
## Describe your changes

The converted API returns a list of barcodes with file extension, we need to strip that since we only store the barcodes from our other processes

### DIFF
--- a/etl-pipeline/services/grin/conversion.py
+++ b/etl-pipeline/services/grin/conversion.py
@@ -1,6 +1,7 @@
 # Script run daily to scrape and initialize conversion for GRIN books acquired in the previous day
 # Temporarily, script will also intialize conversion for backfilled books
 
+from sqlalchemy import update
 from .grin_client import GRINClient
 import pandas as pd
 from model import GRINState, GRINStatus, Record, FRBRStatus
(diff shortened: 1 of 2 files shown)


etl-pipeline/services/grin/conversion.py
```diff
@@ -72,6 +72,7 @@ def process_converted_books(self):
         if len(converted_barcodes) > 0:
             for barcode in converted_barcodes:
                 try:
+                    barcode = barcode.split(".", 1)[0] # converted file name has the following pattern 1234.tar.gz.gpg
```
priya-pai, line 75: This worked for me! 2 thoughts:
1. I think we should change the method `converted` to `converted_filenames` in the grin_client to make it clear what that method is returning. OR we should have that method do this data formatting. It is a little weird to expect a list of barcodes and be returned filenames instead, so renaming might help. What do you think?
2. I think though we chatted earlier about refactoring this method to implement a bulk update instead of doing a query per barcode. Thoughts?

Toxiapo: 1. i like it, updated
2. I think we chatted about the download process but yea i think bulk update makes sense here as well. I updated this to use the method you showed me, LMK what you think!

etl-pipeline/services/grin/conversion.py
```diff
@@ -67,16 +68,22 @@ def save_barcodes(self, barcodes, state):
...
-                    self.db_manager.session.query(GRINStatus).filter(
-                        GRINStatus.barcode == f"{barcode}",
-                        GRINStatus.state != GRINState.DOWNLOADED.value,
-                        GRINStatus.state != GRINState.CONVERTED.value,
-                    ).update({"state": GRINState.CONVERTED.value})
+                    update_barcodes = (
+                        update(GRINStatus)
+                        .filter(GRINStatus.barcode.in_(stripped_barcodes))
+                        .values(state=GRINState.CONVERTED.value)
+                    )
```
priya-pai, line 85: Love this bulk update! I think we might need one more filter to check if a converted barcode has a state of DOWNLOADED, because we don't want to overwrite those entries in the DB and potentially re-download them

Toxiapo: Good catch! Thank you

priya-pai (approved):

PR #760, opened 2025-05-19
GRIN acquisition and ingestion

priya-pai:
## Describe your changes
This PR adds the functionality to acquire and ingest GRIN books when we are ready to hit the green light. 
First, it adds a conversion script which we imagine will run daily on ECS. This conversion script:
- scrape all of the new books added to GRIN in the past day
- initiate conversion for these ^ books and add these books to our DB
- initiate conversion for a batch of backfilled books per day until we are done with the backfill
- check if there are books that have finished conversion and update them in our DB

Second, it adds a download script which will also run daily on ECS. This download script:
- gets all of the books that have been converted in the past day
- downloads this subset of books from GRIN and uploads them to s3 Glacier
- it also downloads and uploads a batch of backfilled books per day until we are done with the backfill

## How to test
The best way to test the conversion script is to go to GRIN's summary reports by month (https://books.google.com/libraries/NYPL/_reports) and go into a month and find a day that has just a couple of NEW books (just so you don't end up initiating conversion for a bunch of books). Hardcode that day in the `acquired_today` method in the `grin_client` - line 81 - by doing `datetime(2025, 5, 3)`. Then, you can run `python3 -m services.grin.conversion`

The best way to test the download script is to have a few entries in your `grin_statuses` table with the status `converted`. Then you can run: `python3 -m services.grin.download`


etl-pipeline/model/postgres/grinStatus.py
kylevillegas93, line 1: let's rename this file to grin_status.py to be more pythonic!

etl-pipeline/model/postgres/grinStatus.py
```diff
@@ -34,6 +35,9 @@ class GRINStatus(Base, Core):
     )
     record = relationship("Record", back_populates="grin_status")
 
+    def historical_timestamp():
```
kylevillegas93, line 38: Is this the first known ingestion time of a book in GRIN?

etl-pipeline/services/grin/util.py
```diff
@@ -0,0 +1,15 @@
+from typing import Iterator
+
+
+def chunk(xs: Iterator, size: int) -> Iterator[list]:
```
kylevillegas93, line 4: you may be able to use `islice` from `itertools`

etl-pipeline/services/grin/conversion.py
```diff
@@ -0,0 +1,141 @@
...
+from .util import chunk
+
+CHUNK_SIZE = 1000
+
+
+class GRINConversion:
+    def __init__(self):
+        self.client = GRINClient()
+        self.db_manager = DBManager()
+        self.logger = create_log(__name__)
```
kylevillegas93, line 22: nice - i like the logger initialized like this!

etl-pipeline/model/postgres/grinStatus.py
priya-pai: This is the timestamp that all books in the backfill will be marked with. This timestamp helps us figure out if a book is a backfilled title or a new title (we had debated whether to add a new field to the DB or not, but settled on using the timestamp instead). I can change the name of this to be `backfill_timestamp` to make that clearer?

etl-pipeline/services/grin/conversion.py
```diff
@@ -0,0 +1,141 @@
...
+                )
+
+    def convert_barcodes(self, barcodes):
+        converted_data = self.client.convert(barcodes)
+        converted_df = self.transform_scraped_data(converted_data)
+        converted_barcodes = converted_df.query('Status == "Success"')
+        return converted_barcodes["Barcode"]
+
+    def save_barcodes(self, barcodes, state):
+        if len(barcodes) > 0:
```
kylevillegas93, line 82: Could do `if not barcodes: return` so that the rest of the code is on the first indent of the method block.

etl-pipeline/services/grin/conversion.py
```diff
@@ -0,0 +1,141 @@
...
+            for chunked_barcodes in chunk(iter(barcodes), CHUNK_SIZE):
+                records: List[Record] = []
+                for barcode in chunked_barcodes:
+                    records.append(
+                        Record(
```
kylevillegas93, line 87: For some reason I am not sure if the date_modified and date_created dates will be added to this even tho there is a default. Just recommend double checking !

etl-pipeline/services/grin/conversion.py
```diff
@@ -0,0 +1,141 @@
...
+            except Exception:
+                self.db_manager.session.rollback()
+                self.logger.exception(
+                    f"Failed to update the following records: {chunked_barcodes}"
+                )
+
+    def process_converted_books(self):
+        converted_barcodes = self.client.converted_filenames()
+
+        if len(converted_barcodes) > 0:
```
kylevillegas93, line 108: same as above!

etl-pipeline/services/grin/download.py
```diff
@@ -0,0 +1,106 @@
...
+from model import GRINState, GRINStatus, Record
+from managers import DBManager, S3Manager
+from logger import create_log
+from sqlalchemy import select, desc
+from .util import chunk
+from .grin_client import GRINClient
+
+logger = create_log(__name__)
+
+S3_BUCKET_NAME = "drb-grin-files-test"
```
kylevillegas93, line 14: we should consider using the limited files bucket which is private or another private bucket!

etl-pipeline/services/grin/download.py
```diff
@@ -0,0 +1,106 @@
...
+BATCH_SIZE_LIMIT = 1000
+
+
+class GRINDownload:
+    def __init__(self):
+        self.s3_client = S3Manager()
```
kylevillegas93, line 20: recommend changing this to store_manager or s3_manager since there is a client which is the S3 client that can be referenced from this object

etl-pipeline/services/grin/download.py
```diff
@@ -0,0 +1,106 @@
...
+                    book.grin_status.failed_download += 1
+                    self.db_manager.commit_changes()
+                    continue
+
+                try:
+                    self.s3_client.put_object(
+                        object=content,
+                        key=s3_key,
+                        bucket=S3_BUCKET_NAME,
+                        storage_class="GLACIER_IR",
```
kylevillegas93, line 58: I think this is fine for now but when we get to conversion of these OCR files to PDFs will it cost more to get the files from Glacier from conversion?

kylevillegas93 (commented):
Overall, it's looking really good - just a few minor comments!

etl-pipeline/model/postgres/grinStatus.py
kylevillegas93: backfill_timestamp sounds good!

priya-pai: Updated here: https://github.com/NYPL/digital-research-books/pull/760/commits/4f76f2032f105586f4c98286f1e0789f7bc575b2

etl-pipeline/services/grin/conversion.py
priya-pai: Yes, great call to use a guard clause - implemented here: [8a8ddae](https://github.com/NYPL/digital-research-books/pull/760/commits/8a8ddae25c5455d3a5e78939095be1e16fa33f67)

etl-pipeline/services/grin/conversion.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class GRINConversion:
+    def __init__(self):
+        self.client = GRINClient()
+        self.db_manager = DBManager()
+        self.logger = create_log(__name__)
+
+    def run_process(self, backfill=False):
+        self.db_manager.create_session()
```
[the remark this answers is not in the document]
priya-pai: Thanks for this note! Is this what you're imagining? Sorry, I'm newer to using these with clauses
```suggestion
        def run_process(self, backfill=False):
            with self.db_manager.__enter__() as connection:
                    self.acquire_and_convert_new_books()

                    self.process_converted_books()

                    if backfill:
                        self.convert_backfills(CHUNK_SIZE)

            self.db_manager.session.__exit__()
```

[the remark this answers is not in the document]
kylevillegas93: hmm! i think what you can do is 

`with DBManager() as self.db_manager:`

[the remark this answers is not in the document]
kylevillegas93: im also down to leave it as is for now - i think this over abstraction makes it somewhat difficult to work with sqlalchemy as intended

etl-pipeline/services/grin/download.py
priya-pai: Changed it to just use the file bucket in the environment file! https://github.com/NYPL/digital-research-books/pull/760/commits/8b4d40a809097207295eb79fccd1cdaac4dc8187

kylevillegas93 (approved):
lets ship it! great work!

etl-pipeline/services/grin/conversion.py
priya-pai: Checked and I believe it's working correctly!

etl-pipeline/services/grin/download.py
priya-pai: Yes, it does cost more to get files from Glacier ($0.01 per GET request) vs. s3 standard ($0.0004 per GET request)! Our plan was to store the TAR files in s3 Glacier and the unpacked files in s3 standard. So, these files are just the TAR files though - we aren't unpacking the tar files in this script. We *could* unpack them in this script and upload the files to s3 standard. However, I think the data retrieval cost might be more (there's around 1000 files in each TAR, so that would 0.0004 * 1000 files = 0.40 to retrieve each of OCR/METs files)? It will probably depend on the new PDF pipeline we build and how it works with the TAR files though!

etl-pipeline/services/grin/conversion.py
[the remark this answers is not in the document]
priya-pai: Thanks for noting this, I learned something new :) done here: https://github.com/NYPL/digital-research-books/pull/760/commits/46dbb1afe026da54a9537f801a18ef613d286a13

etl-pipeline/services/grin/download.py
```diff
@@ -0,0 +1,106 @@
...
+        if backfill:
+            backfilled_books: List[Record] = self._get_converted_books(
+                batch_size, backfill
+            )
+            self.download_and_upload_books(backfilled_books, batch_size)
+
+        daily_converted_books: List[Record] = self._get_converted_books()
+        self.download_and_upload_books(daily_converted_books, batch_size)
+
+        self.db_manager.session.close()
```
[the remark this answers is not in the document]
priya-pai: https://github.com/NYPL/digital-research-books/pull/760/commits/46dbb1afe026da54a9537f801a18ef613d286a13

etl-pipeline/services/grin/util.py
priya-pai: Leaving this for now, but noted for a future optimization!

etl-pipeline/services/grin/conversion.py
```diff
@@ -0,0 +1,141 @@
...
+    def convert_barcodes(self, barcodes):
+        converted_data = self.client.convert(barcodes)
+        converted_df = self.transform_scraped_data(converted_data)
+        converted_barcodes = converted_df.query('Status == "Success"')
+        return converted_barcodes["Barcode"]
+
+    def save_barcodes(self, barcodes, state):
+        if len(barcodes) > 0:
+            for chunked_barcodes in chunk(iter(barcodes), CHUNK_SIZE):
+                records: List[Record] = []
```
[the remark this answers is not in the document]
priya-pai: Leaving this for now, but noted for a future optimization!

etl-pipeline/services/grin/download.py
kylevillegas93: sounds good ! i think with the initial pull - we will just start small!

etl-pipeline/services/grin/download.py
```diff
@@ -0,0 +1,107 @@
...
+from sqlalchemy import select, desc
+from .util import chunk
+from .grin_client import GRINClient
+import os
+
+S3_BUCKET = (
+    "drb-files-limited-production"
+    if os.environ.get("ENVIRONMENT", "qa") == "production"
+    else "drb-files-limited-qa"
+)
```
kylevillegas93, line 17: recommend setting the storage location when we initialize! the ENVIRONMENT may not be set at that point!

priya-pai: oh thanks for the heads up! https://github.com/NYPL/digital-research-books/pull/760/commits/b53600669f7ab69ec84608069d74d5d80ad93a6c

PR #762, opened 2025-05-22
GRIN initial scrape script update

priya-pai:
## Describe your changes
The GRIN initial scrape script grabs all of the barcodes that exist in GRIN and inserts them into the DB. Upon testing locally, I found that the script failed when a record with the same barcode already exists in the DB - instead, we want it to just skip over that barcode and continue inserting. This PR adds logic for that as well as cleans up the scrape script to use the GrinClient and makes the batch size an argument, so that we can configure this more easily. Because we'd like to insert a smaller subset of books in QA and Prod, I added a break in the loop to finish after one batch of books (will remove this when we want to run the script in its entirety).

Lastly, I made the batch size configurable as a command line argument in the conversion and download script as well. 

## How to test
Run `python3 -m scripts.GRIN_initial_scrape --batch_limit=5` which should insert 5 books from GRIN's site into your DB locally (in the Records table and in the GRINStatus table). If you run it again, the script should succeed but just not insert anything into your DB.

### DIFF
--- a/etl-pipeline/services/grin/conversion.py
+++ b/etl-pipeline/services/grin/conversion.py
@@ -143,4 +143,4 @@ if __name__ == "__main__":
     batch_limit = args.batch_limit
 
     grin_conversion = GRINConversion(batch_limit)
-    # grin_conversion.run_process(backfill=True)
+    grin_conversion.run_process(backfill=True)
--- a/etl-pipeline/services/grin/conversion.py
+++ b/etl-pipeline/services/grin/conversion.py
@@ -86,6 +86,7 @@ class GRINConversion:
                         uuid=uuid4(),
                         frbr_status=FRBRStatus.TODO.value,
                         source_id=f"{barcode}|grin",
+                        source="grin",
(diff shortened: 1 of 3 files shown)


etl-pipeline/services/grin/conversion.py
```diff
@@ -78,14 +78,15 @@ def save_barcodes(self, barcodes, state):
...
-        for chunked_barcodes in chunk(iter(barcodes), CHUNK_SIZE):
+        for chunked_barcodes in chunk(iter(barcodes), self.batch_limit):
             records: List[Record] = []
             for barcode in chunked_barcodes:
                 records.append(
                     Record(
                         uuid=uuid4(),
                         frbr_status=FRBRStatus.TODO.value,
                         source_id=f"{barcode}|grin",
+                        source="grin",
```
priya-pai, line 89: This field was null before, but added this since it will be helpful in the future!

kylevillegas93 (approved):
Looks good!

PR #764, opened 2025-05-27
SFR-2705: Deprecate cover process

kylevillegas93:
## Describe your changes
- Our cover process does not work as intended
- This change deprecates the cover process

## How to test
`make unit`


priya-pai (approved):
Looks good - yeah echoing @sarangj I think you can remove the `CoverManager` class too, it seems unused now

kylevillegas93:
Ah good shout!

PR #768, opened 2025-05-28
Move GRIN conversion + download to processes

priya-pai:
## Describe your changes
This PR moves the GRIN conversion and download script to processes and also adds some logging info messages to help with traceability when we run these scripts on ECS. Also, I had to implement a small change to the arguments in these scripts for them to work right with how we have `main.py` set up.

## How to test
`python main.py --process=GRINDownload --e local`
`python main.py --process=GRINConversion --e local`


etl-pipeline/processes/grin/conversion.py
```diff
@@ -15,12 +15,12 @@
...
 
 class GRINConversion:
-    def __init__(self, batch_limit=1000):
+    def __init__(self, *args, batch_limit=1000):
         self.client = GRINClient()
         self.logger = create_log(__name__)
         self.batch_limit = batch_limit
 
-    def run_process(self, backfill=False):
+    def runProcess(self, backfill=True):
```
priya-pai, line 23: changed this so that it would work right with `main.py` I know it's against Python convention :(

kylevillegas93: no worries! we should make this change soon - should be easier once the ingest processes are deprecated!

PR #771, opened 2025-06-02
[SCHOL-4] GRIN ingest process

kylevillegas93:
## Describe your changes
- Implementing the GRIN ingest process
- Need to work on determining the rights next which could get interesting 

## How to test
`make functional`


etl-pipeline/model/source.py
```diff
@@ -5,6 +5,7 @@ class Source(Enum):
     CHICACO_ISAC = "isac"
     CLACSO = "clacso"
     DOAB = "doab"
+    GRIN = "grin"
```
kylevillegas93, line 8: I think this is fine for now @priya-pai but we might want to consider maybe determining the source depending on the OCR package name or the like? maybe we should be using nypl as the source even if it's coming from GRIN. need to flush that out but we shall see

priya-pai (approved):

PR #772, opened 2025-06-03
[SCHOL-11] Add metric to track ingest count per source

kylevillegas93:
## Describe your changes
- Adds metric to track ingest count per source

## How to test
`make functional`

### DIFF
--- a/etl-pipeline/services/monitor.py
+++ b/etl-pipeline/services/monitor.py
@@ -1,5 +1,5 @@
 import newrelic.agent
-from model import Record, Source
+from model import Record
 
 
 def record_event(event_name: str, data: dict):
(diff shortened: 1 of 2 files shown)


etl-pipeline/services/monitor.py
```diff
@@ -77,3 +77,10 @@ def track_record_pipeline_message_failed(execution_time: float, message_body: st
         "execution_time": execution_time,
     }
     record_event(event_name, data=data)
+
+
+def track_records_ingested(number_of_records: str, source: str):
```
priya-pai, line 82: curious why the `number_of_records` is a `str` instead of an `int`?

kylevillegas93: oop

PR #773, opened 2025-06-05
Download and upload lambda

priya-pai:

### DIFF
--- a/etl-pipeline/processes/grin/bulk_download.py
+++ b/etl-pipeline/processes/grin/bulk_download.py
@@ -1,0 +1,113 @@
+# Script run daily to download GRIN books that were converted in the past day
+
+from datetime import datetime, timedelta
+from typing import List
+from model import GRINState, GRINStatus, Record
+from managers import DBManager, S3Manager
+from logger import create_log
+from sqlalchemy import select, desc
+from .util import chunk
+from .grin_client import GRINClient
+import os
+import argparse
+
+
+class GRINBulkDownload:
+    def __init__(self, *args, batch_limit=1000):
+        self.s3_manager = S3Manager()
+        self.client = GRINClient()
+        self.logger = create_log(__name__)
+        self.bucket = (
+            "drb-files-limited-production"
+            if os.environ.get("ENVIRONMENT", "qa") == "production"
+            else "drb-files-limited-qa"
+        )
+        self.batch_limit = batch_limit
+
+    def runProcess(self, backfill=True):
+        with DBManager() as self.db_manager:
+            if backfill:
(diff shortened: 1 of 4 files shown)


etl-pipeline/processes/grin/bulk_download.py
```diff
@@ -0,0 +1,113 @@
+# Script run daily to download all GRIN books that were converted in the past day
+
```
priya-pai, line 2: heads up I basically renamed the previous `download.py` to be `bulk_download.py`, so all of the changes in this file are not new. We might want to eventually remove this file because I don't foresee using it, but didn't want to remove it just yet

etl-pipeline/scripts/grin_file_acquisition.py
```diff
@@ -0,0 +1,16 @@
+from processes.grin.download import GRINDownload
+import logging
+
+logger = logging.getLogger()
+
+
+def lambda_handler(event, context):
+    # TODO: Add parsing of the actual StepFunction input to retrieve barcode correctly
+    barcode = 33433115534525
```
priya-pai, line 9: This will be addressed in a follow up PR when we have the stepfunction created and it's clear what will be sent to the lambda

kylevillegas93: We could consider leveraging the existing API to start the workflow to convert the OCR package: https://github.com/NYPL/pdf-pipeline/blob/main/src/api/workflow.py

I think would just need to changeup the params :)

etl-pipeline/processes/grin/bulk_download.py
kylevillegas93: Sounds good! Plan is still to add the download/upload steps in the step function?

priya-pai: Yes! Sorry this is just clarifying that `download.py` below is the file with the correct upload/download flow that will download and upload a barcode's TAR file and then unpack and upload its respective parts. 

This `bulk_download` file is to download/upload all books that were converted in the past day - what we had initially built for with the GRIN work. For now, I just renamed it and I think let's remove it from the codebase when we have things working end to end.

etl-pipeline/scripts/grin_file_acquisition.py
priya-pai: @kylevillegas93 Because we're trying to deprecate the `pdf-pipeline` (right?), I figure that we can kick off the ECS task directly to convert the OCR package. We'll probably need to have the new Step Function configured in s3 and the process for the ECS task written in order to kick off the right thing. LMK if that sounds ok!

kylevillegas93 (approved):

PR #777, opened 2025-06-25
NO-REF: Remove Union and Optional typing

kylevillegas93:
## Describe your changes
- Removes Optional and Union typing: https://docs.python.org/3.10/library/stdtypes.html#types-union

## How to test
`make unit`


priya-pai (approved):
🙌🏽

PR #785, opened 2025-07-01
Send SQS messages in Conversion script

priya-pai:


etl-pipeline/processes/grin/conversion.py
```diff
@@ -195,6 +200,12 @@ def transform_scraped_data(self, data):
 
         return pd.DataFrame(rows, columns=headers)
 
+    def send_sqs_messages(self, converted_barcodes):
+        sqs_manager = SQSManager("drb-grin-ingest-queue-qa-tf")
```
jackiequach, line 204: this should eventually be added to the env var files

etl-pipeline/processes/grin/download.py
```diff
@@ -79,7 +79,7 @@ def unpack_and_upload_ocr_files(self, file_content):
                 for file in tar_file:
                     self.s3_manager.put_object(
                         object=tar_file.extractfile(file).read(),
-                        key=self.ocr_dir + f"{file.name}",
+                        key=self.ocr_dir + str(file.name),
```
jackiequach, line 82: I think `file.name` will always be a string here and this could just be `self.ocr_dir + file.name`

etl-pipeline/processes/grin/conversion.py
```diff
@@ -24,7 +24,9 @@ def runProcess(self, backfill=True):
         with DBManager() as self.db_manager:
             self.acquire_and_convert_new_books()
 
-            self.process_converted_books()
+            successfully_converted_books = self.process_converted_books()
+
+            self.send_sqs_messages(successfully_converted_books)
 
             if backfill:
                 self.convert_backfills()
```
jackiequach, line 32: messages should be sent to the queue for backfills

etl-pipeline/processes/grin/conversion.py
kylevillegas93: +1 - let's add these to the env variables.

kylevillegas93 (approved):
Looking good!

etl-pipeline/processes/grin/conversion.py
priya-pai: Added here: https://github.com/NYPL/digital-research-books/pull/785/commits/8ad7d434bc59e2bf0d34f27ae82be14c95ab98d6

PR #786, opened 2025-07-02
SCHOL-43: Prep Grin record for ingest

jackiequach:
## Describe your changes

- Updates the MARC record mapping to handle more fields that may be included in the Grin metadata file
- Updates the MARC record test to verify values are correct
- Adds commented out section in GrinIngestProcess for ingesting a mapped record. This should be uncommented once we put in the logic to determine if a book is out of copyright.

## How to test
`make functional`


priya-pai (approved):
Great work!!

PR #715, opened 2025-04-24
SFR-2677: Set up frontend Playwright tests in BrowserStack Test Observability

clarissarichard:
## Describe your changes
- Runs Playwright tests for DRB Web on BrowserStack Test Observability on a schedule

## How to test
- Observe build runs [here](https://observability.browserstack.com/projects/NYPL+DRB/builds) in BrowserStack Test Observability

PR #790, opened 2025-07-07
Remove Browserstack

samanthaandrews:
## Describe your changes
Remove browserstack reporting since we've made the decision not to use it moving forward.

## How to test
I ran `npm run build`, `npm run start`, and `npm run test` without error.


clarissarichard (approved):
Confirmed this reverts changes from [this](https://github.com/NYPL/digital-research-books/pull/715/files#diff-3ebf69f247f3231fd796e60555489b4a1ed684e3ad4cb0ace460a2ed07d53a95) PR. Not sure if you wanted to address it in this PR, but `.github/workflows/web-playwright-schedule.yml` also can be removed!

PR #793, opened 2025-07-07
[SCHOL-51] Improve API app unit tests

kylevillegas93:
## Describe your changes
- Using TDD to improve source code and test code

## How to test
`make unit`

### DIFF
--- a/etl-pipeline/tests/unit/processes/test_api_process.py
+++ b/etl-pipeline/tests/unit/processes/test_api_process.py
@@ -25,16 +25,16 @@ class TestAPIProcess:
             return APIProcess()
 
     def test_run_process_success(self, api_process: APIProcess):
-        with patch("processes.api.API") as MockFlaskAPI:
-            mock_flask_api = MockFlaskAPI.return_value
+        with patch("processes.api.API") as MockAPI:
+            mock_api = MockAPI.return_value
 
             api_process.runProcess()
 
             api_process.db_manager.generate_engine.assert_called_once()
             api_process.redis_manager.create_client.assert_called_once()
             api_process.elastic_search_manager.create_elastic_connection.assert_called_once()
-            mock_flask_api.create_error_responses.assert_called_once()
-            mock_flask_api.run.assert_called_once()
+            mock_api.create_error_responses.assert_called_once()
+            mock_api.run.assert_called_once()
 
     def test_run_process_failure(self, api_process: APIProcess):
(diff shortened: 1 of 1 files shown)


etl-pipeline/api/app.py
```diff
@@ -63,27 +71,25 @@ def run(self):
         else:
             serve(self.app, host="0.0.0.0", port=80)
 
-    def createErrorResponses(self):
+    def create_error_responses(self):
```
kylevillegas93, line 74: Migrating over to snake case

etl-pipeline/api/app.py
```diff
@@ -63,27 +71,25 @@ def run(self):
...
         @self.app.errorhandler(404)
-        def pageNotFound(error):
-            logger.warning("Page not found")
-            logger.debug(error)
+        def page_not_found(error):
+            logger.exception("Page not found")
```
kylevillegas93, line 77: logger.exception also logs the error. 

we don't usually use debug logs - removing

etl-pipeline/api/app.py
```diff
@@ -28,32 +28,40 @@
...
+    collection,
+    collections,
+    citation,
+    fulfill,
+]
 
-class FlaskAPI:
-    def __init__(self, dbEngine, redisClient):
+
+class API:
```
kylevillegas93, line 49: It's not good practice to use the underlying api framework in the naming. We could switch to a different framework, which now will require fewer changes.

etl-pipeline/processes/api.py
```diff
@@ -22,9 +22,9 @@ def runProcess(self):
             redis_client = self.redis_manager.create_client()
             self.elastic_search_manager.create_elastic_connection()
 
-            api = FlaskAPI(db_engine, redis_client)
+            api = API(db_engine, redis_client)
 
-            api.createErrorResponses()
+            api.create_error_responses()
```
kylevillegas93, line 27: moving to snake case here

etl-pipeline/tests/unit/api/test_api_app.py
```diff
@@ -0,0 +1,30 @@
+import os
+import pytest
+
+from api.app import API
+
+
+class TestAPI:
```
kylevillegas93, line 7: We could unit test the init, _register_blueprints, and create_error_responses but we start losing value here with the number of lines of codes required to mock everything.

PR #797, opened 2025-07-08
Update date range for acquired_today

jackiequach:
## Describe your changes
- Updates the acquired_today method to query from yesterday to today (previous day) instead of today to tomorrow (current day)
- Add `book_state` query param to remove the manual check of state


kylevillegas93 (approved):
Looks good to me

PR #799, opened 2025-07-09
Change tests with same name

roseb89:
## Describe your changes
There were two test files called test_redis_manager.py in the unit and functional folders, this was causing the command to run all tests pytest or python -m pytest, to not work. So, I changed the file names 

Screenshot 2025-07-09 at 9 24 15 AM

## How to test
run pytest or python -m pytest


kylevillegas93:
good call! 

we can use `make unit` or `make functional` or `make integration` to run the tests separately, so having the same test names shouldn't be an issue.

PR #805, opened 2025-07-11
SCHOL(74) GRIN client integration test

roseb89:
## Describe your changes
Integration tests for GRIN client.  
## How to test
python -m pytest tests/functional/test_grin_client.py


etl-pipeline/tests/conftest.py
```diff
@@ -448,3 +451,13 @@ def mock_sqs_manager():
...
+
+
+@pytest.fixture(scope="module")
+def grin_client():
+    client = GRINClient()
+    assert client is not None
+    assert isinstance(client.creds, Credentials)
+    assert client.session is not None
+    assert isinstance(client.session, AuthorizedSession)
+    yield client
```
kylevillegas93, line 463: quick point on readability - i like the assertions here to make sure things are solidly setup in the client. I recommend to use new lines to logically break up the code. So below we are using the new lines to break instantiations, assertions, yield/return statement. 

```
client = GRINClient()

assert client is not None
assert isinstance(client.creds, Credentials)
assert client.session is not None
assert isinstance(client.session, AuthorizedSession)

yield client
```

etl-pipeline/tests/integration/test_grin_client.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+
+@pytest.mark.parametrize(
+    "fragment", ["_available", "_in_process", "_converted", "_failed", "_all_books"]
+)
+class TestGRINClient:
+    def test_url(self, grin_client, fragment):
```
kylevillegas93, line 8: i think if you want we could bump this to a functional or unit test and focus these tests on just ensuring the requests to GRIN are solid.

PR #807, opened 2025-07-11
[NO-REF] Add timeout to integration test API calls

kylevillegas93:
## Describe your changes
- Adding timeout to integration test API calls
- We really shouldn't be waiting that long for the request to pass.


priya-pai (approved):
Good thinking to separate these!

PR #817, opened 2025-07-22
Grin conversion optimizations

priya-pai:
## Describe your changes
This PR does the following for the GRIN conversion script:
- Increases the default conversion batch limit to 5000
- Removes SQS message send entirely
- Removes limiting in the backfill query, so all of the books that are pending conversion in the DB will be kicked off for conversion and batch by batch, will be updated in the DB.


kylevillegas93 (approved):
looks good :)

PR #821, opened 2025-07-24
SCHOL-77: Test GRIN download

clarissarichard:
## Describe your changes
This PR adds functional tests for the GRIN download process.
- Adds S3 bucket access validation tests (read and write access)
- Confirms the download process uploads files to the correct bucket and updates the database accordingly
- Fixes regex pattern for cross-platform path compatibility and corrects AWS environment variable names

## How to test
`pytest -s tests/integration/test_grin_download.py`

If the barcode is already "Downloaded", add `_set_grin_status(db_manager, TEST_BARCODE, GRINState.CONVERTED)` to the beginning of `test_download_process` and then remove it for following test runs.


etl-pipeline/tests/integration/test_grin_download.py
```diff
@@ -0,0 +1,83 @@
...
+
+    print("DB connection string:", os.environ.get("DATABASE_URL"))
+
+    # confirm bucket exists
+    service = GRINDownloadService(TEST_BUCKET)
+    assert service.bucket == TEST_BUCKET, (
+        f"Expected bucket name '{TEST_BUCKET}', but received {service.bucket}"
+    )
+
+    download_result = service.download_barcode(TEST_BARCODE)
```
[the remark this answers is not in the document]
clarissarichard: After taking a look at @priya-pai's latest PR, I added `bucket = os.environ["PRIVATE_FILE_BUCKET"]` and the tests now all pass! Just not sure I entirely understand why and unsure where the `drb-files-limited-local` bucket can be found.

clarissarichard:
> One test passes and another tests fails for me. The failing test is the test_download_process Do we need to install gpg? Screenshot 2025-08-01 at 1 03 31 PM

@roseb89 hm perhaps you might need to! Just to double check, are the requirements.txt and dev-requirements.txt installed?

clarissarichard:
> @clarissarichard I had to do the brew install gpg
> 
> I also received this new test error about the converted barcode now being downloaded. I wonder if you would have to reset the barcode to converted after each test run Screenshot 2025-08-01 at 2 32 54 PM

@roseb89 I'm also seeing that error in my other PR, but that's not my test file. @kylevillegas93 since there's a little bit of overlap between our tests, should I merge them and delete one of the files so there is no duplication?

clarissarichard:
> > > @clarissarichard I think @priya-pai made a download-test.py and that's what's breaking. I was trying to fix it because it looks like it just needs a different barcode.
> > > I think since we're using hardcode barcodes the tests are going to be flaky because we have to constantly updated them. I use random barcodes.
> > 
> > 
> > @roseb89 I added a quick if statement at the end of the test. Does it work for you? You may need to add `_set_grin_status` at the beginning of the test to reset the state. I've been manually changing it from downloaded back to converted in my local db, but this definitely makes it easier!
> 
> Maybe you could try a setup and teardown were you can try to revert it back to converted.That way you won't have to keep manually changing it.

@roseb89 With the if statement that I just added in my last commit, I actually no longer need to manually change it. I do like the idea of setup and teardown though. Would it be alright if I try that out in another PR since I don't know how to do that yet?

roseb89:
> @roseb89 With the if statement that I just added in my last commit, I actually no longer need to manually change it. I do like the idea of setup and teardown though. Would it be alright if I try that out in another PR since I don't know how to do that yet?


Ok, understood. 

I've pulled the latest changes, but the 'downloaded' error persists. Is it necessary to revert the downloaded status back to converted on my local machine first? If so, could we add those steps to the 'how to test' section?

clarissarichard:
> > @roseb89 With the if statement that I just added in my last commit, I actually no longer need to manually change it. I do like the idea of setup and teardown though. Would it be alright if I try that out in another PR since I don't know how to do that yet?
> 
> Ok, understood.
> 
> I've pulled the latest changes, but the 'downloaded' error persists. Is it necessary to revert the downloaded status back to converted on my local machine first? If so, could we add those steps to the 'how to test' section?

@roseb89 Yes, you can add `_set_grin_status(db_manager, TEST_BARCODE, GRINState.CONVERTED)` to the beginning of `test_download_process`

.github/workflows/etl-pipeline-tests.yaml
```diff
@@ -23,8 +23,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       IS_CI: "true"
-      AWS_ACCESS: ${{ secrets.AWS_ACCESS_KEY_ID }}
-      AWS_SECRET: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
kylevillegas93, line 27: These are hardcoded env variables so let's change this back!

```
self.aws_access_key_id = os.environ.get("AWS_ACCESS", None)
self.aws_secret_access_key = os.environ.get("AWS_SECRET", None)
```

etl-pipeline/processes/grin/download.py
```diff
@@ -52,7 +52,7 @@ def download_barcode(self, barcode):
             decrypted_ocr_package = self.decrypt_ocr_package(
                 barcode, tmp_ocr_package, tmp_dir
             )
-            self.upload_unpacked_ocr_files(barcode, ocr_dir, decrypted_ocr_package)
+            self.unpack_and_upload_mets_file(barcode, ocr_dir, decrypted_ocr_package)
```
kylevillegas93, line 55: Does this need a merge with main? I think these are changes from last week right?

etl-pipeline/tests/integration/test_grin_download.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # confirm the METS file exists
+    mets_key = f"grin/{TEST_BARCODE}/NYPL_{TEST_BARCODE}.xml"
+    try:
+        s3_client.head_object(Bucket=test_bucket, Key=mets_key)
+        assert mets_file == mets_key, (
+            f"METS file path mismatch: expected {mets_key}, got {mets_file}"
+        )
+    except ClientError as e:
+        assert False, f"METS file {mets_key} does not exist in S3: {e}"
```
kylevillegas93, line 124: feel free to break this out into a separate function: 

```
assert_mets_file_uploaded(...
````

This way the comment is not necessary as a way to breakup the code.

File: etl-pipeline/tests/integration/test_grin_download.py [hunk 5]
```diff
@@ -0,0 +1,133 @@
...
+
+    # confirm barcode has DOWNLOADED state
```
kylevillegas93, line 126: Comment is redundant - feel free to remove.

etl-pipeline/tests/integration/test_grin_download.py
```diff
@@ -0,0 +1,144 @@
...
+        return False
+
+
+def _localstack_bucket_write_access(s3, bucket_name):
+    test_key = "test-write-access.txt"
+    try:
+        s3.put_object(Bucket=bucket_name, Key=test_key, Body=b"test")
+        logger.info(f"Confirmed write access to bucket '{bucket_name}'.")
+        s3.delete_object(Bucket=bucket_name, Key=test_key)  # deletes the test file
+        return True
```
kylevillegas93, line 94: let's use the s3_manager instead of the s3_client fixture

You can access the s3 client via `s3_manager.client`

kylevillegas93 (approved):
Looks good - nice work!

PR #822, opened 2025-07-24
[SCHOL-75] Continuously convert barcodes

kylevillegas93:
## Describe your changes
- Continuously converts barcodes that are pending conversion
- If the process is daily, just check for newly added barcodes

### DIFF
--- a/etl-pipeline/processes/grin/conversion.py
+++ b/etl-pipeline/processes/grin/conversion.py
@@ -13,7 +13,7 @@ from .. import utils
 
 
 class GRINConversion:
-    def __init__(self, *args, batch_limit=5000):
+    def __init__(self, *args, batch_limit=1000):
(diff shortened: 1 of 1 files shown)


etl-pipeline/processes/grin/conversion.py
```diff
@@ -1,109 +1,128 @@
...
+        self.params = utils.parse_process_args(*args)
         self.client = GRINClient()
         self.logger = create_log(__name__)
         self.batch_limit = batch_limit
 
-    def runProcess(self, backfill=True):
-        with DBManager() as self.db_manager:
-            try:
-                self.acquire_and_convert_new_books()
+    def runProcess(self):
```
priya-pai, line 22: Why not wrap everything in this method in just one `with DBManager() as self.db_manager`? Or LMK if that's frowned upon.

PR #823, opened 2025-07-24
[NO-REF] Change runProcess to run

kylevillegas93:
## Describe your changes
- Method name change from runProcess to run

### DIFF
--- a/etl-pipeline/main.py
+++ b/etl-pipeline/main.py
@@ -52,17 +52,17 @@ def main(args):
             source,
             options,
         )
-    except:
+    except Exception:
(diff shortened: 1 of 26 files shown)


priya-pai (approved):
woo 🥳🐍

roseb89 (commented):
Is this to make  runProcess more readable? If so how?  Python has a built-in run function.

kylevillegas93:
@roseb89 `runProcess` is redundant because the classes are processes. Moreover, `runProcess` uses camelcase and not snake case.

kylevillegas93:
also im not sure if python has a built-in run function: https://docs.python.org/3/library/functions.html

PR #825, opened 2025-07-24
[NO-REF] Fix hathi trust integration test

kylevillegas93:
## Describe your changes
- Increase field size limit by factor of 2

## How to test
`pytest tests/integration/sources/test_hathi_trust_service.py`


roseb89 (commented):
I noticed this test failure on my branch. So the field size limit can randomly change? Does this happen often? These sort of test failures prevent other branches from merging into main and if these can randomly change, then we must make a note of the test failures that happen often and why. So, I can at least to expect to update some field sizes and maybe right changes?

priya-pai (approved):

PR #836, opened 2025-07-29
Add unavailable barcode GRIN state and logic

priya-pai:
## Describe your changes
This PR parses books that are not allowed to be downloaded from GRIN. This could happen if availability changes at some point and we are not able to convert the book anymore. This PR:
- adds a new UNAVAILABLE GRIN State to the model
- adds logic to mark a book as UNAVAILABLE when trying to convert and we receive that status
- adds logic to grab the correct barcodes to send off for conversion

## How to test
I tested this by trying to run the script on a few barcodes I knew were not available for conversion (33433110152265, 33433110137266) - you can see these on https://books.google.com/libraries/NYPL/_all_books - and made sure they were marked correctly. I would do the same if testing again.

### DIFF
--- a/etl-pipeline/model/postgres/grin_status.py
+++ b/etl-pipeline/model/postgres/grin_status.py
@@ -12,6 +12,7 @@ class GRINState(Enum):
     CONVERTING = "converting"
     CONVERTED = "converted"
     DOWNLOADED = "downloaded"
+    UNAVAILABLE = "unavailable"
 
 
 class GRINStatus(Base, Core):
@@ -28,6 +29,7 @@ class GRINStatus(Base, Core):
             "converting",
             "converted",
             "downloaded",
+            "unavailable",
             name="state",
             create_type=False,
         ),
--- a/etl-pipeline/processes/grin/conversion.py
+++ b/etl-pipeline/processes/grin/conversion.py
@@ -12,7 +12,7 @@ from utils.chunker import chunk
 from .. import utils
 
 IN_PROCESS_LIMIT = 50000
-DELAY_IN_SECONDS = 120
+DELAY_IN_SECONDS = 60
 
 
 class GRINConversion:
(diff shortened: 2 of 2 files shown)


etl-pipeline/processes/grin/conversion.py
```diff
@@ -121,8 +126,12 @@ def sync_converted_books(self) -> set:
...
                         select(GRINStatus.barcode)
                         .filter(GRINStatus.barcode.in_(list(converted_barcodes)))
-                        .filter(GRINStatus.state != GRINState.DOWNLOADED.value)
-                        .filter(GRINStatus.state != GRINState.CONVERTED.value)
+                        .filter(
+                            or_(
+                                GRINStatus.state == GRINState.PENDING_CONVERSION.value,
+                                GRINStatus.state == GRINState.CONVERTING.value,
+                            )
+                        )
```
priya-pai, line 134: I didn't want to add another != statement, so I thought this positive OR logic would be easier to read - lmk if anyone disagrees!

etl-pipeline/model/postgres/grin_status.py
```diff
@@ -28,6 +29,7 @@ class GRINStatus(Base, Core):
             "converting",
             "converted",
             "downloaded",
+            "unavailable",
```
kylevillegas93, line 32: do we need a db migration for this?

etl-pipeline/processes/grin/conversion.py
```diff
@@ -13,7 +12,7 @@
 from .. import utils
 
 IN_PROCESS_LIMIT = 50000
-DELAY_IN_SECONDS = 120
+DELAY_IN_SECONDS = 60
```
kylevillegas93, line 15: THANK YOU :D

kylevillegas93 (approved):
Looks good :)

etl-pipeline/model/postgres/grin_status.py
priya-pai: Discussed via Slack, but b/c the field already exists and is an ENUM, no migration is needed! DB entries were able to be created and updated just fine.

PR #837, opened 2025-07-29
Add pagination to vra

jackiequach:
## Describe your changes
- Adds pagination and results header to reflect page for VRA results panel
- Updates `/chats` to return search parameters to be used with the `/search` endpoint on page change to update the results

image
image


kylevillegas93 (approved):
Looks good!

PR #841, opened 2025-08-04
[NO-REF] Set conversion limit

kylevillegas93:
## Describe your changes
- The conversion script looks like it stopped a few days ago. 
- Adding a conversion limit of 5k to prevent trying to convert 10s of thousands of book at once.


jackiequach (approved):
looks good

PR #842, opened 2025-08-04
Add reader to VRA

jackiequach:
## Describe your changes
**Note**: The `/chats` API is not returning the readable link, so clicking on the "Read online" button will show the embeddable reader.

- Adds reader layout to VRA when "Read online" button is clicked. A React context was added to handle the behavior of the "Read online" button (linking to `/read/{id}` page is default for all pages, excluding the research assistant)
- Adds temporary "Close reader" button at top right to go back to search results (this will be updated in the future to be within the WebReader back button)
- Moves research assistant navbar to a separate component

image
image


kylevillegas93 (approved):
looks good - i think the failing test was fixed yesterday !

PR #847, opened 2025-08-06
Moved test files out of pycache

roseb89:
## Describe your changes
Screenshot 2025-08-05 at 3 19 52 PM
Moved  test files out of pycache

## How to test
Make sure there are no test files in pycache

NOTE: Currently there is only one test file in pycache the name of the file is test_api_app.py. There are two files called test_api_app.py. Before I move it out of pycache I have to know if we're still using this file, if not then it will be deleted since it's a duplicate. Yet, if we're still using both files with the same name, then one of them have to be renamed.


kylevillegas93:
Looks good ! I fixed the tests - that's my fault :}}

kylevillegas93:
@roseb89 - gonna merge these changes in :)

PR #864, opened 2025-10-10
NOREF Update GHA to use OIDC

mwbenowitz:
Currently this repo uses an old approach referencing stored secrets to authenticate the GHA tasks to deploy to AWS. These keys can be compromised and be otherwise brittle. This updates the approach to use OIDC to create short lived tokens that can validate the access


bantucaravan (approved):
Cool! sorry for the delay, I didn't see that my review was requested. I need to figure out my notifications.

PR #870, opened 2025-10-27
NOREF/quickstart-local-dev

bantucaravan:
## Describe your changes
Documentation improvements and fixes necessary to quickstart a new dev (Noah)

## How to test
run against the existing integration tests?
maybe a unit test for raise_on_error in SSMService.get_parameter()


alea12:
Hi @bantucaravan! Yes I think [`get_parameter`](https://github.com/NYPL/digital-research-books/blob/c43f1c68cccaba1bdd2a9548798403d301eb451e/etl-pipeline/services/ssm_service.py#L24-L35) should throw an error if `response["Parameter"]["Value"]` is `None`. Don't have my AWS account yet so hard to tell, could this cause issues elsewhere?

jackiequach:
> Also it doesn't explain why the AWS SDK is not authenticating after I merged Mike's PR for AWS auth in GHA. Do you have any ideas?

I think `etl-pipeline-tests.yaml` needs to be updated to add a step for `aws-actions/configure-aws-credentials@v4` so the env vars are available for the tests.

bantucaravan:
> I think `etl-pipeline-tests.yaml` needs to be updated to add a step for `aws-actions/configure-aws-credentials@v4` so the env vars are available for the tests.

ah thank you!
and it seems like "etl-pipeline-ci.yaml" and "etl-pipeline-ci-self-hosted.yaml" need that too

etl-pipeline/requirements.txt
alea12: @bantucaravan could we start sorting `requirements.txt` by default? Ideally this would be enforced by a linter, but for now we can sort manually (`Cmd + Shift + P` → “Sort Lines Ascending” in VS Code).

bantucaravan: ok, do you know if ruff handles that?

.github/workflows/etl-pipeline-tests.yaml
```diff
@@ -18,19 +18,27 @@ on:
...
       DOCKER_BUILDKIT: 1
       COMPOSE_DOCKER_CLI_BUILD: 1
 
     steps:
       - uses: actions/checkout@v4
 
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::946183545209:role/GithubActionsRunnerRole
```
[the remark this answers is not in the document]
bantucaravan: I just realized this :) I changed deployer to runner

etl-pipeline/requirements.txt
alea12: Ruff doesn't handle that as far as I know... We could just format it manually pre-commit for now, and discuss automating in a separate PR.

refs: https://github.com/astral-sh/ruff/issues/8418

etl-pipeline/Dockerfile
```diff
@@ -8,8 +8,6 @@ RUN pip install -r requirements.txt
 
 RUN pip install --no-cache-dir newrelic
 
-RUN pip install elasticsearch>8.0.0
-
```
bantucaravan: @mwbenowitz this was contradicting the reqs.txt
(and out of line with the deployed ES version which was 7.x) just want to check that this is ok to remove and there are no known untested negative implications of this.

PR #887, opened 2025-12-15
SCHOL-168: Update addHasPartToGRINRecords script to include existing records

jackiequach:
## Describe your changes
- Updates addHasPartToGRINRecords to query for records that are public domain, with `grin_status=downloaded`, and record `state=ingested` (has not been through the record pipeline). See SQL query below
- Handles if the record has already been updated with the `first_page_url` and updates to append the `Part` string instead of overwriting the `record.has_part`.
- Adds additional logging

## How to test
`docker compose up`
`python main.py -sc addHasPartToGRINRecords -e local`

```
SELECT COUNT(*)
FROM records r
JOIN grin_statuses gs ON gs.record_id = r.id
WHERE r.source = 'grin'
  AND split_part(r.rights, '|', 2) = 'public_domain'
  AND gs.state = 'downloaded'
  AND r.state = 'ingested'
```

This should result in 302,353

### DIFF
--- a/etl-pipeline/scripts/addHasPartToGRINRecords.py
+++ b/etl-pipeline/scripts/addHasPartToGRINRecords.py
@@ -50,6 +50,7 @@ def main():
     record_pipeline_queue = os.environ["RECORD_PIPELINE_SQS_QUEUE"]
     sqs_manager = SQSManager(record_pipeline_queue)
     s3_client = boto3.client("s3")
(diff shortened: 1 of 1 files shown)


alea12:
Note: This change is to rescue records that was ingested but never went through the entire pipeline.

alea12 (approved):
Looks good as discussed!

PR #890, opened 2025-12-23
SCHOL 189: secrets refactor

bantucaravan:
# Changes

Although a lot of files are touched, there are actually only a few core changes.

etl-pipeline/utils/load_env.py -> This is the core addition. creates a new load env function. Simplifies the task to reading a single secret and a single non-secret env var file. Uses the well maintained python-dotenv library

etl-pipeline/config/ -> replaces all the .yaml formatted env files into files supported by the new load_env() function, maintaining all the same variable values.
- add some values previously in untracked env files (i.e from local-qa.yaml) to the  public env files, including some AWS private subnet hostnames.

etl-pipeline/Makefile -> creates alias for recommended command from README for running local dev docker container set up (`make seed`).

etl-pipeline/docker-compose-setup.yml -> Ensures that `make seed` waits until data seeding is finished before exiting. Previously, the set up command occasionally exited after DB tables have been created but before seeding was finished (due to `--abort-on-container-exit` flag)

etl-pipeline/services/ssm_service.py -> removed because all access to AWS SSM parameter store is centralized in load_env()

all other changes -> 
- replace load_env_file() with load_env() 
- use read_env() to access env vars with an informative error message if they are unavailable.

# Documentation

### Semantics of “environment” 
- "environment" refers to the env vars present in the a running application. 
-  The architecture implemented here intends to entirely describe the "environment" with a pair of  files, one for secret env var values and one for non-secret env var values, and further intends to specify those 2 files with a single string (e.g. "production")

### concept 
- all env vars (even secrets) are specified in files, and those files are all tracked in version control
- secrets are stored in a secrets manager (AWS SSM parameter store) and referenced by id. 
- The idea of separating secret from non-secret env vars and referencing secrets by their secrets manager id is inspired by the ECS task definition “environment” and “secrets” fields (see: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters_ec2.html) 
-specification in source controlled files allow easy change and customization of compute environments
- Separate env vars and secret env vars explicitly
- No secrets are specified in plain text locally
- This architecture clearly separates errors associated with loading secrets from SSM parameter store from errors related to failure to configure an environment variable, by centralizing dependence on reads from parameter store in the load_env() and using read_env() to raise informative errors at the time of accessing env var values elsewhere in the codebase.

### `load_env()` details
- (IMPORTANT) loading secrets depends on AWS authentication already being pre-configured the compute environment. This is achieved in different ways for each of our usual environments: local machine, github actions, AWS ECS
- Env File naming -> by default `load_env()` expects the name of the secret env var file to be  "<public_env_file>.secrets" 
- Env file contents -> env files are read by [python dotenv](https://saurabh-kumar.com/python-dotenv/). For secrets, the env value should be an AWS SSM parameter store parameter ARN or full parameter name (starting with `/`). If parameter name is used AWS account and region will be inferred from the local AWS auth profile.
- by default, env vars set before `load_env()` is run are not overriden by configured values.
-  by default,  if the env config files don’t exist, no error will arise
- load_env()  errors if a AWS SSM parameter is configured in a secrets file but does not exist in SSM parameter store.

# Alt. Proposal

Proposal ->  Configure secrets in each execution environment separately and do not store ANY environment configuration files (secret or public) in source control. 
- AWS ECS -> use the ECS task definition "environment" and "secrets" keys as referenced above.
- github actions -> configure public env vars in github actions workflow files and secrets in the repos github actions secrets.
- local machine -> use untracked (!) .env files that contain both public and secret env vars, but use a load_env() function (like python-dotenv) that does not error if the env file is not present (as it will be absent in GHA and AWS ECS).

#### Pros
* Easy export env to ECS task definition format is needed 
* Separates env vars and secrets explicitly
* flexibility; you can run the  "production" environment file from your local computer

##### Cons
* Secrets rely on AWS authentication already preconfigured in the compute environment
* you can’t specify a secret as plain text locally
* dev require write access to SSM parameter store
* potential confusion because you can run the "production" environment file from your local computer

# FUTURE TODO
- rename all env files to not start with .dot because …. they are supposed to be tracked and read (.secrets.env and .env)
- Update readme with new secrets details
- Define and document parameter store variable naming convention
- Define and document criteria for what is and what is not a secret (My initial proposal is: anything that you can charge money or access data by using in conjunction with all (public) env vars.
    - Decide if the elastic and other host names should be secret
- remove “hardcoded” envvars from cloud deployment environments
- identify all SSM parameters not used in env.secrets files and duplicates of other SSM parameters and delete them (clean up )
- create docker compose health check to ensure local resources are running before tests start (this is to address the test issue)
- clarify semantics of env handling in tests (particularly the handling of existing ENV vs ENVIRONMENT env var)
- define semantics of conflicts between existing ENVIRONMENT env var and named env (file) in main.py. specifically define the supremacy of the ENVIRONMENT env var and main.py's role as setting it based on the passed arguments
- fix newrelic ENVIRONMENT use in main.py
- Jira ticket: combine LocalDevelopmentSetupProcess and SeedLocalProcess into a single idempotent process and move it out of the main docker compose yaml (and if the idempotency check is fast enough maybe leave it in)

# Open Questions
- Is it better to separate secrets and env files into different folders rather than distinguish them by filename only (as is implemented in this PR)
- Should we put env/ secrets files in a different folder than config (for example logstash config is in there)? 
- Should local-compose (env inside the local dev docker image) passwords become secrets (they are currently public in the repo so I left them that way)?

# How to test

All existing test should pass
The only test code changed are:
- removed tests for a unused and removed `DBManager.decrypt_env_var()`
- added unit tests for read_env()
- refactored `setup_env` pytest fixture to use the new `load_env()` and required other fixtures to depend on this fixture.

Note: the test mentioned in [this ticket](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-207) failed like 5 times before passing with no code changes. So I may need to fix that to really be done with this. My diagnosis is that it has to do with the varying start up times for the local dev docker services (there is a fixed 90 second wait in the CI/CD workflow). My proposed fix is to add a docker compose health check that waits till the services are actually ready (with some longer timeout) before starting the tests.

### DIFF
--- a/etl-pipeline/config/.env.local-qa.secrets
+++ b/etl-pipeline/config/.env.local-qa.secrets
@@ -1,0 +1,28 @@
+# copied from .env.qa.secrets
+
+# CONTENT_CAFE_USER=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/contentcafe/user
+# CONTENT_CAFE_PSWD=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/contentcafe/pswd
+# ELASTICSEARCH_PSWD=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/elasticsearch/pswd
+# ELASTICSEARCH_USER=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/elasticsearch/user
+# GITHUB_API_KEY=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/github-key
+# GOOGLE_BOOKS_KEY=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/google-books/api-key
+# HATHI_API_KEY=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/hathitrust/api-key
+# HATHI_API_SECRET=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/hathitrust/api-secret
+# NEW_RELIC_LICENSE_KEY=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/newrelic/key
+# NYPL_API_CLIENT_ID=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/nypl-api/client-id
+# NYPL_API_CLIENT_PUBLIC_KEY=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/nypl-api/public-key
+# NYPL_API_CLIENT_SECRET=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/nypl-api/client-secret
+# NYPL_BIB_PSWD=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/postgres/nypl-pswd
+# NYPL_BIB_USER=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/postgres/nypl-user
+# OCLC_METADATA_ID=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/oclc-metadata-clientid
+# OCLC_METADATA_SECRET=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/oclc-metadata-secret
+# OCLC_CLIENT_ID=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/oclc-search-clientid
+# OCLC_CLIENT_SECRET=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/oclc-search-secret
+POSTGRES_PSWD=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/postgres/pswd
+POSTGRES_USER=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/postgres/user
+# GOOGLE_API_KEY=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/google-ai-api-key
+# API_KEY=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/api-key
+# GRIN_ACCESS_KEY=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/grin-access-key
+# GRIN_AUTH=arn:aws:ssm:us-east-1:946183545209:parameter/drb/qa/grin-auth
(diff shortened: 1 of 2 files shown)


etl-pipeline/config/.env.example
```diff
@@ -0,0 +1,46 @@
...
+NYPL_LOCATIONS_BY_CODE=xxx
+
+# NYPL API Credentials
+NYPL_API_CLIENT_TOKEN_URL=xxx
+
+# Bardo CCE API URL
+BARDO_CCE_API=xxx
+
+# Allowed sources of CORS requests to proxy endpoint
+API_PROXY_CORS_ALLOWED=(?:Source1:Source2)
```
alea12: If this is expecting a regex, could we rename this to `API_PROXY_CORS_ALLOWED_REGEX` to be clear?

bantucaravan:
I want to add a bunch of PR level explanation documentation, there is plenty of context and questions I have that I would like to get feedback from everyone on, but I am putting that off until I can push the backend e2e (which, as I noted in passing last week, does not actually depend on this to work if a parameter store change is made) as that seems greater urgency

bantucaravan:
here is the ticket https://newyorkpubliclibrary.atlassian.net/browse/DOPS-1869

etl-pipeline/main.py
bantucaravan: why is this outside of main().... move inside

etl-pipeline/config/.env.local-qa
```diff
@@ -0,0 +1,70 @@
...
+RECORD_PIPELINE_SQS_QUEUE=drb-record-pipeline-queue-qa-tf
+GRIN_INGEST_SQS_QUEUE=drb-grin-ingest-queue-qa-tf
+
+# OCLC CONFIGURATION
+# OCLC API keys must be configured in secrets file
+OCLC_QUERY_LIMIT=390000
+
+# AWS CONFIGURATION
+AWS_REGION=us-east-1
+# AWS_ACCOUNT_ID: "946183545209" # nypl-digital-dev # currently unused
```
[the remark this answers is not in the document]
bantucaravan: removed.

etl-pipeline/config/.env.example
alea12: I see, in that case let's just leave it as is!

PR #897, opened 2026-01-05
NOREF addHasPartToGRINRecords batch commits to improve memory management

jackiequach:
## Describe your changes

- Commits every 1000 records to avoid holding too many objects in the session
- Updates to use DBManager commit_changes to handle deadlocks

## How to test
```
docker compose up
python main.py -sc addHasPartToGRINRecords -e local
```


bantucaravan:
I see the semantics are "is this a retry" as opposed to (how i was interpreting) an instruction to "retry this execution". Ill add a doc str to clarify

PR #1036, opened 2026-04-01
NOREF: require docker compose up for local tests

bantucaravan:
## Describe your changes

* make api service dependent on devsetup. The DB and ES index creation done in devsetup is in fact a requirement for the api to work. This allows the assumption that "when the api service is up and "healthy", everything is ready for use of the API" to be correct.

* update the docker compose health check script, to focus on ensuring that all declared services with defined healthchecks are "healthy". This way, if all services that are background services have defined healthchecks and accurate dependencies, the healthcheck script will be an accurate check for tests or development use being ready to go.

* add an early exit of non-unit tests in the "local" env are attempted when the docker compose services are NOT runnning

* add early exit in tests if loading the environment fails

* replace docker-compose.setup.yaml with a short simple script. If the reviewer is comfortable with this replacement I will update the README.

## How to test
try (for example) `uv run pytest tests/functional` when docker compose services are down, there should be a fast exit
try (for example) `uv run pytest tests/functional` when only `docker compose database` is up, there should be a fast exit
try (for example) `uv run pytest tests/functional` when docker compose services are up, should work
try (for example) `uv run pytest tests/functional` when aws sso on your local is not refreshed or the target env file is not present, there should be fast exit

to test the seed db changes:
* `docker compose down -v` to remove cached volume data
* `docker compose up` to start local services
* `./seed_local_db.sh` to seed db (this should error if local services are not running)


etl-pipeline/seed_local_db.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -euo pipefail
+
+./docker-compose-healthcheck.sh
+
+source .venv/bin/activate
+python main.py -e local -p SeedLocalDataProcess -i daily
```
bantucaravan, line 7: @jdohan can you add a line here to seed GRIN postgres data needed for /chat? Wouldn't we also need to seed some S3 files?
Thoughts?

jdohan: Are you referring to the setup required for running the /chat API tests? In #1025, that's handled in a [new step](https://github.com/NYPL/digital-research-books/blob/2cbef268a07d79627aa5ca505b972d05c96c4188/.github/workflows/etl-pipeline-tests.yaml#L102-L109) added to the ETL Pipeline Tests workflow.

It might be worthwhile to assess the value of `SeedLocalDataProcess` as a setup step for VRA. While it ingests new records from HathiTrust and FRBRizes them into a local Postgres, it stops short of running them through the vector indexing pipeline. We could expand it to do that (perhaps using a dynamically created Turbopuffer namespace), but then we wouldn't know exactly which books end up in the vector database. Since happy-path API tests rely on specific prompts returning specific books, we really need the dataset to be a known quantity.

Because `SeedLocalDataProcess` isn't currently depended on for non-unit CI checks anyway (even for DRB), #1025 provides an alternate option for seeding local Postgres with _known_ books (see [c7d9766](https://github.com/NYPL/digital-research-books/pull/1025/changes/c7d9766803e06696446bbec166e1f665b0a56a29)).

Also, regarding S3: for the /chat tests specifically, I think we only need the Postgres metadata to map up with the vector DB, so I don't believe we need to seed physical S3 files for the tests to pass.

Hope that answers your question.

jdohan (approved):

alea12 (commented):
@bantucaravan @jdohan  I have a general question about this approach:

```sh
docker compose up --wait
```

won't this be sufficient? It also [implies detached mode](https://docs.docker.com/reference/cli/docker/compose/up/#options). I'm wondering if we could simply replace 

```yml
 - name: Start services
   run: docker compose up -d
 
 - name: Wait for services to be healthy
   run: ./docker-compose-healthcheck.sh
```

with: 

```yml
 - name: Start services
   run: docker compose up --wait
```

in `etl-pipeline-tests.yaml`.

etl-pipeline/seed_local_db.sh
jdohan: I overlooked the convenience of having the seed fixture run as part of the script for setting up local testing (as it’s handled in CI already in aforementioned PR). I'll add the commands after it's merged to main but just a heads up that if you're testing changes against a fully local setup before publishing, `SeedLocalDataProcess` can be skipped and the local Postgres can be seeded directly from the JSON fixture instead. Keep in mind that the current set of books in that file is limited; it will grow, but be mindful of the FRBRized data locally when verifying semantic search results for /chat.

jdohan:
I’m in favor of keeping the health check due to the increased logging in case of deadlocks and more aggressive fail-fast behavior by aborting immediately if a container crashes or something else goes awry, rather than just hanging until a timeout.

bantucaravan:
@alea12 @jdohan 
My concerns are more around the implied detach mode. For local dev there are cases we want the docker network up non-detached and elsewhere check that they are healthy. Specifically i plan to pull this check into the pytest setup for the local env where I want to be able to check that the docker project is up and its ussually running in blocking mode in a different terminal.

I will admit I wasn't aware of --wait when I was writing this.


That said an approach that could handle wanting to check health without detached mode and without this custom script could be to run `docker compose up --no-recreate --abort-on-container-failure` instead of yhe custom healthcheck. i *think* that would get at making sure everything is running without killing what is already running, tho i need to look deeper into what happens when you call docker compose up when things are already running.

For now the custom health check is fine IMO, but i should explore alternatives

alea12:
@jdohan @bantucaravan thanks for the context, sounds good, I'll be ready to run a final review & approve once the conflict is resolved

bantucaravan:
@jdohan  @alea12 so I did some experimenting
An exit code check on `docker compose up --no-recreate --wait` would pretty much replace the effect of the healthcheck. The main difference is that `docker compose up --no-recreate --wait` re-runs the idempotent DB init service. the healthcheck script on the other hand, parses the file to only check services that have defined healthchecks, which the DB init does not, thus ignoring the DB init. Also this will actually start the services if they are not running already which is a different ask than than just checking if they are running. This is the biggest difference. It is problematic because it makes the healthcheck in the pytest start up responsible for starting the services if they are not already running.

I am going remove the custom healthcheck script, and split the difference and add a 10 second timeout to the healthcheck in pytest start up so that if the services are starting from scratch it will timeout. This balances the benefit of removing custom code and not having pytest responsible for starting tests. Overall I don't feel strongly or confidently about this, so other suggestions for other configurations are welcome, and I can always revert.

alea12:
@bantucaravan thanks for confirming! I'm ok with removing the custom healthcheck too.

alea12 (approved):

jdohan:
> @jdohan @alea12 so I did some experimenting An exit code check on `docker compose up --no-recreate --wait` would pretty much replace the effect of the healthcheck. The main difference is that `docker compose up --no-recreate --wait` re-runs the idempotent DB init service. the healthcheck script on the other hand, parses the file to only check services that have defined healthchecks, which the DB init does not, thus ignoring the DB init. Also this will actually start the services if they are not running already which is a different ask than than just checking if they are running. This is the biggest difference. It is problematic because it makes the healthcheck in the pytest start up responsible for starting the services if they are not already running.
> 
> I am going remove the custom healthcheck script, and split the difference and add a 10 second timeout to the healthcheck in pytest start up so that if the services are starting from scratch it will timeout. This balances the benefit of removing custom code and not having pytest responsible for starting tests. Overall I don't feel strongly or confidently about this, so other suggestions for other configurations are welcome, and I can always revert.

The native command works for simplicity but perhaps keeping the health check script around as a diagnostic tool could be valuable for one-off purposes due to the clearer visibility provided by its logging.

bantucaravan:
@jdohan  i'll move it into dev_scripts/

PR #1040, opened 2026-04-06
NOREF: Optimize Playwright tests workflow

jdohan:
### Single dockerized build
**Web CI** now exports the dockerized app it builds to **Playwright Tests** to be used as the testing target, whereas previously the app would be built again (but natively) for running the Playwright tests against. This removes redundant build steps, reduces execution time, and ensures the Playwright tests run against the same type of build that'll eventually be deployed.

### Security enhancements
All sensitive values are now fetched from SSM and masked in workflow logs, plus additional protection applied to the API key.

#### Dummy Airtable key

Note that the Airtable key is given a dummy value to prevent risk of exposure of the actual key in the exported artifact since it's baked in upon building the Docker image. While a cleanup step exists following the test run, there is a window of opportunity during testing in which anyone can access the artifact since this is a public repository and the only settings for access restrictions for GitHub Actions artifacts deal with the time in which they're retained.

Therefore no sensitive values are exported in the tar file. Confirmed that no tests depend on having this value set properly, as none exist that actually submit the feedback form (and I don't think we'd want to pollute actual feedback with test data).

All other sensitive values (e.g. `API_KEY`) are passed to the `docker run` command and cannot be read by users without AWS privileges.

### Runner optimizations & cleanup
Prevents parallel runs on the self-hosted runner via a `concurrency` block,  which protects the self-hosted runner from conflicting jobs and saves compute time. Specific cleanup steps have been added that remove the Docker image artifact, app container, and a generated environment file to free up runner disk space.

### Enhanced debugging
Configures the workflow to dump application logs automatically upon completion of local test runs.

### Refactored execution logic
Refactored the conditional logic for running VRA tests and updated manual runs to target the QA application environment by default.

### Workflow renaming
The file `playwright.yml` has been renamed to `web-playwright-tests.yml`, aligning it with the naming convention used by other workflows.

## How to test
Confirm CI checks pass and note any improvement in execution time.

Prior to opening this PR, the following commands were run locally to simulate the workflows and confirm the app gets built  without failure and functions as expected:

Build the Docker image
```
docker build \
  --build-arg APP_ENV="development" \
  --build-arg airtable_api_key="dummy_airtable_api_key" \
  -t web:secure-ci-artifact .
```

Start the dockerized app
```
docker run --name web-secure-simulator -d \
  -e BASE_URL="http://local.nypl.org:3000/" \
  -e API_URL="https://drb-api-qa.nypl.org" \
  -e API_KEY=<qa-key> \
  -p 3000:3000 \
  web:secure-ci-artifact
```


jdohan:
@jackiequach note that I’ve removed about 30 lines of setup code from [test-utils.ts](https://github.com/NYPL/digital-research-books/blob/NOREF/optimize-playwright-tests-workflow/web/playwright/support/test-utils.ts) with [6e2f132](https://github.com/NYPL/digital-research-books/pull/1040/commits/6e2f1324ab922235d45e61cef39ae1f9d499df03). It was attempting to spin up a native Next.js production server for every integration test, which would [crash](https://github.com/NYPL/digital-research-books/actions/runs/24367146423/job/71162061803) the new Docker-centric approach this PR aims for.

I've confirmed that all tests pass whether targeting a native or dockerized app following the change.

jdohan:
Since both frontend and backend testing workflows use the self-hosted runner, they'll benefit from a shared concurrency group using a block similar to the one introduced to the frontend workflow by this PR. That will prevent them from running simultaneously and competing for resources, potentially causing build failures (`cancel-in-progress` should be `false` though to prevent one cancelling the other out).

I'm holding off on that update though to avoid merge conflicts with this PR since it involves a file renaming. I'll open a small follow-up PR for it once this merges (pending review).

That is, unless we believe the recent memory bump on the runner is sufficient enough to handle concurrent workflow runs.

cc: @bantucaravan

jackiequach:
The changes in this PR are reverting #1041.

jdohan:
> The changes in this PR are reverting #1041.

Good callout @jackiequach. There must've been a merge conflict mishap. Following a reset and force push, everything looks to be up to date with main now other than the files in the scope of these changes.

.github/workflows/web-ci.yml
```diff
@@ -121,22 +84,56 @@ jobs:
...
-          docker build -f Dockerfile --build-arg airtable_api_key=$AIRTABLE_KEY \
-          --build-arg APP_ENV=$APP_ENV .
+          docker build -f Dockerfile \
+            --build-arg APP_ENV=$APP_ENV \
+            --build-arg airtable_api_key=$AIRTABLE_KEY \
+            -t web:test .
+          docker save -o /tmp/web.tar web:test
+
+      - name: Upload Docker image as artifact  # For usage in Playwright tests
+        uses: actions/upload-artifact@v4
```
jackiequach, line 108: Is there a reason why we are not using the latest version of this action?

.github/workflows/web-ci.yml
```diff
@@ -121,22 +84,56 @@ jobs:
...
+  cleanup:
+    name: Clean up Docker image artifact
+    needs: [playwright_tests]
+    if: always()
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+    steps:
+      - name: Delete web-image artifact  # Done natively for security
+        uses: actions/github-script@v7
```
jackiequach, line 133: Same as above

jackiequach (commented):
looking good so far, just a few questions related to actions versions

.github/workflows/web-ci.yml
jdohan: I initially held off because `v5+` introduced a breaking requirement for a matching `download-artifact` step on self-hosted runners, however I just double-checked and confirmed our runner version (2.333.1) fully supports the latest action version.

Thanks for flagging! I've updated both the upload and download actions to `v7`.

.github/workflows/web-ci.yml
jdohan: Updated to latest version.

jdohan:
Added step **Upload error context (if any)** to the Playwright Tests workflow, which will attach an artifact that assists with debugging after noticing a message such as the following in a [recent failed run](https://github.com/NYPL/digital-research-books/actions/runs/24740805585/job/72380099350#step:14:143):
```
Error Context: test-results/research-assistant-page-re-4f144-g-with-non-zero-paging-text-chromium/error-context.md
```
Such a file cannot be accessed without a step that explicitly uploads it.

alea12 (approved):

PR #1043, opened 2026-04-08
Vector indexing end-to-end

kunalnabar:
Includes:
 - Add a process to main.py to run the vector indexing pipeline the same way we run etl-pipeline
 - Some minor fixes

Ultimately we'll be doing the actual orchestration for processing differently, but this allows us to at least generate the data for some books on a target index as needed. You can run this via: 

```
STAGE="development" LOG_LEVEL="debug" DRB_API_HOST=localhost uv run python main.py -p VectorIndexingProcess -e production <barcodes>
```

### DIFF
--- a/etl-pipeline/main.py
+++ b/etl-pipeline/main.py
@@ -51,7 +51,12 @@ def main(args):
         logger.exception(f"Failed to initialize process {process} in {environment}")
         return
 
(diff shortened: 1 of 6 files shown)


alea12 (approved):
Pushed minor fixes, good to go!

PR #1049, opened 2026-04-09
SCHOL 422: stoch process test updates

bantucaravan:
## Describe your changes
Agent behavior tests were not working because they were not updated based on updates to update_chat()
example: https://github.com/NYPL/digital-research-books/actions/runs/24257462801/job/70832911797

* updated the tests to be compatible with current update_chat()
* tweaked tests to be better, focusing on only the filter construction by the LLM and mocking/not accessing other things
* xfailed tests that are unstable but should not block our deployments

* inorder to clean up test conversations in DB, I refactored the creation of read and read/write SQLalchemy engines so that we can delete with a read/write engine while using a read only engine in most circumstances

## How to test
tests should pass or xfail with substantive errors


bantucaravan:
on main merge tests are currently failing... needs this PR to fix
example: https://github.com/NYPL/digital-research-books/actions/runs/24356308137/job/71124407321
@alea12

jdohan (approved):
Looks good to me. One thought for the future regarding the `xfail` tests is we could use conditional logic to mark them that way in CI or based on target environment so that they're run more strictly outside of PR checks (e.g. locally prior to publishing or after deploys to QA like you mentioned).

alea12 (approved):

PR #1066, opened 2026-04-20
NOREF: Update seeds and vector database namespace config

jdohan:
### Updates Postgres seeding
Adds FRBR data for [12 books](https://docs.google.com/spreadsheets/d/1MfPEGzL7_UDNQ9owon235tLJ3ZbMZLmpznm6NEvF-fk/edit?usp=sharing) to the `frbr_seed.json` fixture, which is used to seed a local Postgres db to further support testing in local builds (as /chat only returns hits in the vector db that are FRBRized in Postgres).

Additionally, the script that fetches the data from the prod Postgres db has been updated to merge new books into the JSON file gracefully instead of its prior behavior of overwriting everything. Note this script has been renamed to `generate_frbr_seed_json.py` for better clarity.

### Vector database namespace config
Introduces the envar `TURBOPUFFER_NAMESPACE` into all environment configs, where the testing namespace **`vra-test`** (which contains 14 books as of 4/20/26) is used in **local** and **docker-compose** environments and the default namespace containing 10k+ books **`vra-dev`** is used in **qa** and **production**.

The prior approach of simply declaring the envar in the local environment was not sufficient for local dockerized builds, thus it's been removed from the env block in the **ETL Pipeline Tests (Pull Request)** workflow.

A side effect of pointing to the currently-small testing namespace is that some books returned by catalog search are not as relevant to the query as when the larger dataset is used. For example, "Find books on the history of Missouri" returns the following three books with the small dataset:
- [The Missouri merchant one hundred years ago](https://drb-qa.nypl.org/item/9e6a33c8-eee0-498c-821c-f19e4ac70660?featured=15251797)
- [History of the first half century of the Sand Lake Baptist Church](https://drb-qa.nypl.org/item/e849981b-129b-4020-954c-c907153a0e3d?featured=15297541)
- [La flotte de combat](https://drb-qa.nypl.org/item/e415bab1-8a17-4039-8712-f23cf658a78c?featured=15314880)

This seems to be a better option than pointing to `vra-dev` while only a small set of books are FRBRized in Postgres, where most queries will return no results since the more-relevant vector search hits are not backed by the required Postgres data. I encourage reviewers to weigh in on which option is more desirable, and will adjust if needed. Note that the testing namespace will continue to grow.


bantucaravan:
@jdohan keeping the search index and DB in sync makes sense to me.

alea12 (approved):

PR #1071, opened 2026-04-22
SCHOL-175: Item page AQA

jackiequach:
[SCHOL-175](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-175)

## Describe your changes
- Resolves AQA items 1, 7, 9-11, 22 from [item page qa doc](https://docs.google.com/document/d/1eT_J0WxwGtYVLlyMwcNcoCRB93mliZH0zq2VMKS3Kfk/edit?usp=sharing)
- Upgrades web-reader to v5.0.2 for AQA changes. Had to downgrade emotion packages for peer dependency errors.

## How to test
Locally 

[SCHOL-175]: https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/src/components/ResearchAssistant/MessageBubble.tsx
```diff
@@ -84,7 +84,7 @@ const MessageBubble = memo(
                     ) : (
                       <Flex alignItems="center" justifyContent="space-between">
                         <AiGeneratedText />
-                        <FeedbackButtons />
+                        <FeedbackButtons label={`message ${index} feedback`} />
```
jackiequach: This label is pending recommendation from Clare since I don't think it is a good descriptor for screen readers.

jackiequach: Per Clare (in context to message bubble feedback button labels):

> They don't need to be unique. We'll have to rely on the surrounding content giving context to the buttons. The main reason to make button and link labels unique on the page is to support navigating by lists of links/buttons or by tabbing through, so a screen reader user can quickly tell the purpose of the button and go right to it or activate it without having to examine surrounding content. But these buttons are highly dependent on the surrounding content. You have to read what came before in order to decide which button to use (if at all). So let's leave them non-unique.

web/package.json
```diff
@@ -16,10 +16,10 @@
   },
   "dependencies": {
     "@chakra-ui/react": "2.5.4",
-    "@emotion/react": "11.14.0",
-    "@emotion/styled": "11.14.1",
+    "@emotion/react": "11.11.1",
+    "@emotion/styled": "11.11.0",
```
alea12, line 20: Question: Is there a blocker on the web reader side to use a more recent version?

web/src/components/ResultCard/ResultCard.tsx
```diff
@@ -155,7 +155,7 @@ export const ResultCard: React.FC<ResultCardProps> = ({
             __css={{ button: { padding: "xs" } }}
           >
             <AiGeneratedText />
-            <FeedbackButtons />
+            <FeedbackButtons label="Summary feedback" />
```
alea12: nits: should this start with a lower case?

web/package.json
jackiequach: Yes the peer dep is from the web-reader.

web/src/components/ResultCard/ResultCard.tsx
jackiequach: It should not matter the casing for this since screen readers will read it in the same way whether lowercase or uppercase, but I'll update this to be lower case to be consistent with the other labels.

alea12 (approved):

PR #1072, opened 2026-04-22
misc clarity fixes

bantucaravan:
## Describe your changes
* clarify param name in record ingestor
* code comments

### DIFF
--- a/etl-pipeline/logger.py
+++ b/etl-pipeline/logger.py
@@ -82,28 +82,37 @@ class LogContextVars:
         self._context_var.reset(self._token)
 
 
-def configure_loggers():
+def configure_loggers(log_level=None, stage=None):
     """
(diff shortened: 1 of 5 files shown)


etl-pipeline/logger.py
```diff
@@ -82,28 +82,37 @@ def __exit__(self, *_):
         self._context_var.reset(self._token)
 
 
-def configure_loggers():
+def configure_loggers(log_level=None, stage=None):
```
alea12, line 85: is it correct to assume these parameter passing would be used mainly to support `dev_scripts`?

bantucaravan: No, its just a general improvement in the api IMO, that I happen to use somewhere in dev scripts

alea12 (approved):

PR #1073, opened 2026-04-23
SCHOL 553: embedding recommendation

bantucaravan:
## Describe your changes
Changes needed to create new search indexes for different embedding models via CLI. This work is used for creating candidate indexes in the embedding model replacement worksteam.

**SageMaker model deployment**
* add deploy scripts for sagemaker and bedrock
* thru-put benchmarking script for sagemaker endpoints
* create embedders for sagemaker and bedrock

**Indexing Pipeline logic**
* do not continue to subsequent steps if the metadata embellishment step fails
* added default step configurations to the main Pipeline object
* create an index config to support pre-configuring index dimensions and embedding models for each TP index

**index books CLI**
* add batching to prevent memory overload processing 10k books in memory
* add barcodes input form grin 10k DB table.

NOTE: this branch is built on top of "NOREF/misc-clarity-fixes" (https://github.com/NYPL/digital-research-books/pull/1072),  so view "files changed" for commits after 3c1d7769c4 for changes unique to this PR

## How to test
one unit test for the "do not continue to subsequent steps if the metadata embellishment step fails" was added

The CLI is not tested


etl-pipeline/vector_indexing/scripts/index_books.py
bantucaravan, line 1: since there is already a scripts folder at the top level , I think we should bring `vector_indexing/scripts/` there rather than proliferate scripts/ folders

etl-pipeline/model/postgres/grin_public_domain_10k.py
```diff
@@ -0,0 +1,14 @@
...
+
+from .base import Base
+
+
+class GrinPublicDomain10k(Base):
+    __tablename__ = "grin_public_domain_10k"
+
+    barcode = Column(Text, primary_key=True)
+    record_id = Column(
+        Integer, unique=True, nullable=True
```
bantucaravan, line 13: @alea12  it would be better if it were not nullable but is was created as nullable and this reflects that. I could recreate the table, but it seems like not a big deal

bantucaravan: also we need to talk to kunal and decide on a DB schema versioning plan

etl-pipeline/vector_indexing/scripts/index_books.py
```diff
@@ -0,0 +1,586 @@
...
+DEFAULT_RESULTS_DIR = Path(__file__).resolve().parent / "indexing_results"
+JOB_METADATA_FILE = "job_metadata.json"
+JOB_STATE_FILE = "job_state.json"
+
+
+class MockEmbedder:
+    """Mock embedder for testing - returns random vectors."""
+
+    def __init__(self, dimensions: int | None = None):
+        self.dimensions = dimensions or get_config().embedding_dimensions
```
alea12, line 82: `get_config` seems to be missing from the imports

etl-pipeline/vector_indexing/components/embedders/bedrock.py
```diff
@@ -0,0 +1,92 @@
...
+        model_arn: str,
+        dimensions: int = None,  # QWEN3_EMBEDDING_8B_DIMS,
+        region: str = "us-east-1",
+        aws_profile: str | None = None,
+        assume_role: str | None = None,
+    ) -> None:
+        self._model_arn = model_arn
+        self._dimensions = dimensions
+        if aws_profile:
+            boto3.setup_default_session(profile_name=aws_profile)
```
alea12, line 52: this would modify process-wide default session, whereas `SageMakerEmbedder` creates a local `boto3.Session`. are there any reasons we are not using the same pattern?

etl-pipeline/vector_indexing/scripts/index_books.py
```diff
@@ -0,0 +1,586 @@
...
+    if not candidates:
+        return None
+    candidates.sort(key=lambda item: item[0])
+    return candidates[-1][1]
+
+
+##############################
+
+
+#### Barcode Discovery Helpers ####
```
alea12: these are nice! might worth moving this entire block to a separate file

etl-pipeline/vector_indexing/scripts/index_books.py
bantucaravan: @kunalnabar  any architectural considerations?

etl-pipeline/vector_indexing/components/embedders/bedrock.py
bantucaravan: No, really I just didn't update this. this file should probably be deleted since we foundout bedrock does not support custom embedding models. 
Looks like they have built in support for Cohere and AWS embedding models https://docs.aws.amazon.com/bedrock/latest/userguide/knowledge-base-supported.html#knowledge-base-supported-embeddings
but since those are AWS controlled deployments I think we have the same issue of not controlling deprecation cycles, so I think for now we will not be using bedrock. I would probably opt to delete

alea12:
@bantucaravan which AWS profile is recommended to test this implementation end to end? I'm assuming the sandbox account but I run into profile errors with the following:

```
$ python -m vector_indexing.scripts.index_books \
   --barcodes 33433001234567 \
   --index-name vra_test-10k-harrier_oss_v1_.6b \
   --batch-size 1 \
   --results-dir /tmp/test-run
```

bantucaravan:
> @bantucaravan which AWS profile is recommended to test this implementation end to end? I'm assuming the sandbox account but I run into profile errors with the following:
> 
> ```
> $ python -m vector_indexing.scripts.index_books \
>    --barcodes 33433001234567 \
>    --index-name vra_test-10k-harrier_oss_v1_.6b \
>    --batch-size 1 \
>    --results-dir /tmp/test-run
> ```

unfortunately, I just deleted the endpoints from the nypl-sandbox account which are configured in the config.py file. So realistically I need to update the config.py with the new vra-sandbox endpoint names before it will run. Or... I should probably add some kind of config override argument for the CLI.

etl-pipeline/vector_indexing/scripts/index_books.py
bantucaravan: fixed

etl-pipeline/vector_indexing/scripts/index_books.py
bantucaravan: done. in vector_indexing/utils/barcodes.py

bantucaravan:
> @bantucaravan which AWS profile is recommended to test this implementation end to end? I'm assuming the sandbox account but I run into profile errors with the following:
> 
> ```
> $ python -m vector_indexing.scripts.index_books \
>    --barcodes 33433001234567 \
>    --index-name vra_test-10k-harrier_oss_v1_.6b \
>    --batch-size 1 \
>    --results-dir /tmp/test-run
> ```

@alea12  I updated the config and added an override feature. This should work for nypl-vra-sandbox account now.

bantucaravan:
> > @bantucaravan which AWS profile is recommended to test this implementation end to end? I'm assuming the sandbox account but I run into profile errors with the following:
> > ```
> > $ python -m vector_indexing.scripts.index_books \
> >    --barcodes 33433001234567 \
> >    --index-name vra_test-10k-harrier_oss_v1_.6b \
> >    --batch-size 1 \
> >    --results-dir /tmp/test-run
> > ```
> 
> @alea12 I updated the config and added an override feature. This should work for nypl-vra-sandbox account now.

@alea12  I realized I configured the embedders with the profile name on my local. So unless you set your profile name to "vra-sandbox" you will have to set `--config-overrides {"embedder.params.aws_profile":"<your profile name>"}`

alea12 (approved):

PR #1075, opened 2026-04-24
NOREF: Fix pytest config

jdohan:
### Pytest config file name
Updates name and syntax of a pytest configuration file to align with standard convention, as `pytest.toml` was not being picked up automatically in test runs.

Confirmed the settings in the file that weren't being applied yield no side effects now that they're recognized, particularly with `asyncio_mode = "strict"` (which is the default setting anyway).

### New setting
Added `pythonpath = "."` to ensure import paths resolve without excess code when tests are run from etl-pipeline/ (like they are in CI). No code cleanup is needed.

### Workflow update
Added the file to the `paths` block in the **ETL Pipeline Tests (Pull Request)** workflow to ensure it runs after making test configuration changes.

## How to test
Ensure no import errors occur when running tests from etl-pipeline/.


bantucaravan:
just a note: there are async tests, just search `pytest.mark.asyncio` in the codebase

alea12 (approved):

jdohan:
> just a note: there are async tests, just search `pytest.mark.asyncio` in the codebase

@bantucaravan Ah, my bad. I recall doing a quick search to check but it must've been a bad search. Thanks for noting.

PR #1078, opened 2026-04-29
NOREF: Fix New Relic token tracking via custom instrumentation

jdohan:
### Overview
Implements custom instrumentation for [New Relic AI monitoring](https://newrelic.com/platform/ai-monitoring) by manually-defining LLM events and sending them to New Relic instead of relying on their OOTB instrumentation. 

This was needed in order to properly track token usage by our custom agent that utilizes a Gemini LLM. Although the New Relic Python agent has [instrumentation hooks](https://github.com/newrelic/newrelic-python-agent/tree/main/newrelic/hooks) that include support for a range of models, including Gemini (note [`mlmodel_gemini.py`](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/hooks/mlmodel_gemini.py)), token counts would still be zero when viewing the [AI monitoring dashboard](https://one.newrelic.com/nr1-core/mlops-llm/ai-details/MTIxMzM0fEFQTXxBUFBMSUNBVElPTnwxNDcxMTUwODU5?account=121334&begin=1777483995594&end=1777486811486&filters=%28name%20LIKE%20%27qa%27%20OR%20id%20%3D%20%27qa%27%20OR%20domainId%20%3D%20%27qa%27%29&state=da8cbc1e-c119-3826-648c-6c14537df126).

<br>
Screenshot 2026-05-01 at 2 26 22 PM
<br>

This was likely occurring because our application uses the standard `openai` SDK (`AsyncOpenAI`) to route requests to Google's GenAI proxy endpoint (`v1beta/openai`). New Relic's built-in `mlmodel_openai` hook gets intercepted but fails to extract token metrics because Google's endpoint payload doesn't perfectly match the strict sequence and OpenAI schema expectations that New Relic's hook requires for token extraction. By explicitly defining custom LLM events, this mapping limitation is circumvented.

### Feature flags

Feature flags have been added to toggle sending events to New Relic altogether and for sending user prompt text to New Relic (note it still sends the LLM response).

With this custom solution, the relevant settings for the OOTB feature added in #1060 have been disabled to avoid event duplication. 

See the AI monitoring dashboard for [DRB ETL Pipeline (DEV)](https://one.newrelic.com/nr1-core/mlops-llm/ai-details/MTIxMzM0fEFQTXxBUFBMSUNBVElPTnwxNDY0ODkzOTE0?account=121334&begin=1777485075149&end=1777485153876&filters=%28name%20LIKE%20%27qa%27%20OR%20id%20%3D%20%27qa%27%20OR%20domainId%20%3D%20%27qa%27%29&state=da8cbc1e-c119-3826-648c-6c14537df126) for the outcome of this custom instrumentation.

<br>
Screenshot 2026-05-01 at 2 28 52 PM
<br>

## How to test
Spin up a clean local dockerized API using [docker-compose.newrelic.yml](https://drive.google.com/file/d/1JLGRU90zObYVE7X47Cpr4uJnIMoq8v21/view?usp=drive_link) then send chat requests using it. 

From etl-pipeline/:
```
export NEW_RELIC_LICENSE_KEY="<key>"

docker compose down --volumes && \
docker compose build devsetup && \
docker compose run --rm --entrypoint python devsetup main.py \
  -p LocalDevelopmentSetupProcess \
  -e docker-compose && \
docker compose run --rm --entrypoint python devsetup \
  -m tests.integration.api.assistant.support.seed_frbr_data && \
docker compose stop && \
docker compose -f docker-compose.newrelic.yml build --no-cache api && \
LOG_LEVEL=debug docker compose -f docker-compose.newrelic.yml up --force-recreate
```

The requests should be present as conversations in the Responses table on the AI monitoring dashboard for the **DRB ETL Pipeline (DEV)** entity.

**Note**: Running the /chat API tests (e.g. `make integration-vra`) after the above commands complete will create a user `vra_integration_test_user` (whose password is stored as `/drb/local/vra/test_user_password` in Parameter Store). Note the API key to use for the `X-API-Key` header in the docker-compose config file.

While in the New Relic dashboard, run a NRQL query like the following to gain insights on each turn made by the agent during a /chat request:
```
SELECT eventType(), *
FROM LlmChatCompletionMessage, LlmChatCompletionSummary
WHERE appName = 'DRB ETL Pipeline (DEV)'
SINCE '2026-05-18 08:00:00' UNTIL '2026-05-18 09:00:00' WITH TIMEZONE 'America/New_York'
```
Each turn gets captured as an `LlmChatCompletionMessage` event that includes valuable information such as content, role, and sequence. A single `LlmChatCompletionSummary` event also gets sent that ties all respective `LlmChatCompletionMessage` events together and reports key metrics like duration and token count, as well as custom attributes for conversation type and edition ID. 

The results of a query can be exported in CSV or JSON format (e.g. 
[new-relic-llm-event-logging-sample.csv](https://github.com/user-attachments/files/27287505/new-relic-llm-event-logging-sample.csv) for the above query).


etl-pipeline/api/assistant/agent.py
```diff
@@ -268,10 +274,208 @@ class ContentSearchExecutionContext:
...
+                    if (
+                        role == "user"
+                        and os.getenv(
+                            "NEW_RELIC_CUSTOM_LLM_SEND_USER_PROMPT_TEXT_ENABLED"
+                        )
+                        != "true"
+                    ):
+                        pass
+                    else:
+                        msg_payload["content"] = str(content_val)
```
alea12: could we invert the condition so that `pass` would be unnecessary?

etl-pipeline/api/assistant/agent.py
```diff
@@ -290,12 +497,20 @@ async def on_llm_end(
...
-        elapsed_str = f"{elapsed:.3f}s" if elapsed is not None else "unknown"
 
-        output_types = [o.type for o in response.output]
+        self.num_llm_calls += 1
+        self.total_llm_elapsed += elapsed
+        self._last_response = response
+
+        if response.usage:
+            self.total_input_tokens = response.usage.input_tokens or 0
+            self.total_output_tokens = response.usage.output_tokens or 0
```
alea12: clarification: does `usage.(input|output)_tokens` contain the total usage? if not should we `+=` here?

etl-pipeline/api/assistant/agent.py
```diff
@@ -268,10 +274,208 @@ class ContentSearchExecutionContext:
 class LLMLoggingHooks(RunHooks):
```
alea12, line 272: general structural note: `agent.py` is getting quite long, and the New Relic instrumentation logic is pretty much self-contained. would it be worth extracting it into a dedicated module?

alea12 (changes requested):

etl-pipeline/api/assistant/agent.py
jdohan: Looked into this and finding that the usage token properties do **not** contain the total multi-turn usage, but just the token counts for a single LLM API request (i.e. the prompt history sent to the model for that turn plus the new text generated in that turn).

So yes, we should use += here to increment on each turn (like how the number of LLM calls and elapsed time are).

Thanks for noting this; big difference.

alea12 (approved):
thanks @jdohan!

PR #1087, opened 2026-05-05
Code to patch metadata in turbopuffer

kunalnabar:
Allows for patching by document, patching by filter, and adaptive batching for patch operations.


etl-pipeline/vector_indexing/components/backends/turbopuffer.py
```diff
@@ -513,25 +666,128 @@ def _flush_with_retry(
...
+            elif isinstance(value, list):
+                size += (
+                    sum(
+                        len(v.encode("utf-8")) if isinstance(v, str) else 8
+                        for v in value
+                    )
+                    + len(value) * 5
+                )
+            else:
+                size += 8  # Numbers, booleans, etc.
```
alea12, line 734: This could cause nested dicts to be calculated as 8 bytes which would likely cause an underestimate - should we run recursive calculation or do we consider it as an overkill

etl-pipeline/vector_indexing/components/backends/turbopuffer.py
```diff
@@ -202,9 +226,138 @@ def delete_document(self, doc_id: str) -> bool:
             return False
 
     def patch_document(self, doc_id: str, fields: dict) -> bool:
```
alea12, line 230: minor: these new methods are not declared in the base class, I'd like them to be added for consistency if other backends should implement this interface

etl-pipeline/vector_indexing/components/backends/turbopuffer.py
kunalnabar: I think it's overkill in that the patch dict is always going to be flat since all of our chunk attributes are primitives (strings, numbers, dates, lists of strings). Vector can't be patched only upserted. 
My initial alternative was just taking the size of `json.dumps()` on the object. But I didn't want to waste CPU time on a serialization on every single object when this is really just a heuristic to try to max out the bytes on the request. If we miscalculate, we'll split and adaptively make the batch smaller.

I'll add a comment to the API to make sure that anyone using this is aware if a nested dict object is ever added to the chunk document.

etl-pipeline/vector_indexing/components/backends/turbopuffer.py
kunalnabar: Will add!

etl-pipeline/vector_indexing/components/backends/turbopuffer.py
alea12: Sounds fair, thanks for adding those comments!

alea12 (approved):

PR #1093, opened 2026-05-07
SCHOL-452: Locust load testing setup

jdohan:
## Changes summary
Establishes a **Locust load test suite** in the codebase along with its first locustfile that simulates a full frontend user journey.

The test offers immediate value by enabling shifted-left performance testing. By running quick, low-concurrency tests against a local feature branch and its base, a before/after performance assessment can be made before publishing new code.

Note that the same locustfile is compatible with LoadForge, which has its own file management system for reading in external data or secrets (e.g. `authToken`). LoadForge uses a distributed cloud network for generating user traffic, which yields more reliable results for formal test runs (e.g. [Enhanced Search Load Testing](https://docs.google.com/document/d/1STgOAZQFIsmZc7gr-iL1W6gFA5qBempu-hfCVzRtHE8/edit?usp=sharing)).

Future work can be planned to integrate it (or smaller, more focused tests) into the CI pipeline to automate response time measurements and export them in a fashion that allows tracking performance trends across deployments over time. LoadForge can be used by leveraging their [REST API](https://docs.loadforge.com/api-reference/endpoint/run-start).

### Design decisions
- **Locust Playwright plugin**: The [locust_plugins.users.playwright](https://github.com/SvenskaSpel/locust-plugins/blob/master/examples/playwright_ex.py) package is used in the script to extend traditional Locust testing (i.e. sending HTTP requests) to the browser level to accurately-simulate user actions (e.g. chat panel interaction) and gain a full-stack performance assessment.
  - Future work includes load testing the /chat API endpoint directly.
- **Auth bypass**: Logging in via the UI is bypassed by setting `authToken` in local storage before any page navigation.
- **Multi-path design**: Different paths are taken via randomization to simulate a range of user behavior at key decision points, including:
  - Which user prompt to submit
  - Using either text input or a suggestion button on the landing page
  - Choosing whether to submit a follow-up prompt (note this step is commented-out pending [SCHOL-593](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-593))
  - Selecting a book to explore among the returned search results
- **Randomized think-time**: Used `asyncio.sleep` with randomized ranges between each step to simulate realistic user pacing.
- **Static event names**: Event names (e.g. "results page: chat response") are purposefully kept generic so results aggregate cleanly, which avoids sifting through numerous alike events when analyzing test results. Prompts are surfaced in logging and get embedded in exception messages for per-prompt attribution upon failure.
- **Edition ID extraction**: The edition ID of a chosen search result gets parsed from the DOM before navigating to the item page so it can be embedded in content search failure messages, enabling post-hoc correlation of failures to specific books.
- **Centralized test data**: All user prompts are stored in a dedicated JSON file `user_prompts.json` and read in at runtime. When run with LoadForge, prompts are read from the same JSON stored in their file management system.

## How to test
From etl-pipeline/:
```
VRA_USER_AUTH_TOKEN=<token> \
uv run \
  --with "locust" \
  --with "locust-plugins[playwright]" \
  locust \
    -f tests/locust/web_e2e_playwright.py \
    --host <localhost or qa> \
    --users 5 \
    --run-time 10m \
    --spawn-rate 0.0167
```
Navigate to `http://localhost:8089` to access the Locust web UI to start the run and see results in real-time. Add `--headless` to run immediately without the UI.

[SCHOL-593]: https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jdohan:
Should probably put in validation for the auth token.

jdohan:
I've recently been informed that LoadForge system host names always start with `loadforge-`.

So a conditional statement like the following can be used to eliminate the need for committing near-duplicate locustfiles for running LoadForge-compatible tests:

```python
import socket

if socket.gethostname().startswith("loadforge-"):
    # Read data from LoadForge file management system
else:
    # Read data from envars, external files, etc.
```

Note that locustfiles dedicated for LoadForge tests may still be needed on occasion (e.g. if a module is imported from the codebase) since their file management system only supports storing CSV, JSON or TXT formats.

etl-pipeline/tests/locust/web_e2e_playwright.py
```diff
@@ -0,0 +1,226 @@
...
+                    user_auth_token = f.readline().strip()
+            except Exception as e:
+                raise FileNotFoundError(
+                    f"Error loading file containing authToken: {type(e).__name__}: {e}"
+                ) from e
+        else:
+            user_auth_token = os.environ["VRA_USER_AUTH_TOKEN"]
+        await page.context.add_init_script(
+            f"localStorage.setItem('authToken', {json.dumps(user_auth_token)})"
+        )
```
alea12, line 121: Leaving a small note here that these would become unneeded once we remove basic auth

etl-pipeline/tests/locust/web_e2e_playwright.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # 1. Navigate to landing page
+        async with event(self, "landing page: load"):
+            await page.goto("/research-assistant-landing")
+            await page.get_by_role(
+                "heading", level=1, name=re.compile("NYPL Virtual Research Assistant")
```
alea12: Our landing page copies have been updated, could we use the new language here?

alea12 (changes requested):

etl-pipeline/tests/locust/web_e2e_playwright.py
jdohan: Updated locator to find "AI-enabled Enhanced Search".

etl-pipeline/tests/locust/web_e2e_playwright.py
jdohan: Good call. Added a comment for future work to remove after auth is taken down.

alea12 (approved):

PR #1094, opened 2026-05-07
SCHOL-553: add sagemaker inference tei image

bantucaravan:
## Describe your changes
add sagemaker inference tei image

The HF created images do not support the latest TEI version which is required for PPLX and Kalm models

I was not sure where to put this, it was originally in vra_experiments, it could be its own repo. Putting it here for now, but open to other ideas

## How to test
No automated tests

### DIFF
--- a/etl-pipeline/sagemaker_tei_image/serve.py
+++ b/etl-pipeline/sagemaker_tei_image/serve.py
@@ -1,0 +1,158 @@
+import asyncio
+import os
+import shlex
+import signal
+import subprocess
+import sys
+import time
+from contextlib import asynccontextmanager
+
+import httpx
+from fastapi import FastAPI, Request, Response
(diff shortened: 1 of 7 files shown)


alea12 (approved):

etl-pipeline/sagemaker_tei_image/serve.py
```diff
@@ -0,0 +1,158 @@
...
+    print(message, flush=True)
+
+
+def build_tei_command() -> list[str]:
+    cmd = [
+        "text-embeddings-router",
+        "--model-id",
+        HF_MODEL_ID,
+        "--hostname",
+        "0.0.0.0",
```
alea12, line 42: is this mismatch between `TEI_HOST` default value intentional?

alea12 (approved):

PR #1098, opened 2026-05-11
SCHOL-574: Display parsed markdown of responses

jackiequach:
[SCHOL-574](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-574)
### Frontend
- Adds `react-markdown` for parsing markdown responses and uses DS components for formatting.
    - For links, only edition links will be parsed, others will only render text.  
- Adds a mock for react-markdown to render text without parsing since Jest does not support ESM at this time.

### Backend
- Updates the system prompt to format response in markdown.

TODO: Add e2e tests for markdown parsing verification ([ticket](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-601))

### Screenshots
image

image

## How to test
Locally

[SCHOL-574]: https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


etl-pipeline/tests/stochastic_processes/test_agent_responses.py
bantucaravan: ideally we should update the test prompt to something that is actually likely to produce some of the undesired markdown. But really... I think we should delete the test unless it is aligned with some specific markdown the the parsing library does not support

web/src/util/MarkdownParser.tsx
```diff
@@ -0,0 +1,85 @@
...
+            {children}
+          </Heading>
+        ),
+
+        h3: ({ children }) => (
+          <Heading level="h3" size="heading8" fontWeight="bold" marginTop="xs">
+            {children}
+          </Heading>
+        ),
+
```
alea12, line 48: If we are supporting up to h3, it might be worth mentioning that in the system prompt. I understand it's less likely to render h4 with the current prompt and we're unsure how strictly it will be followed, but clarifying the limit could be beneficial

web/src/util/MarkdownParser.tsx
```diff
@@ -0,0 +1,85 @@
...
+export const renderMarkdownContent = (
+  markdownText: string,
+  onEditionClick: (editionId: string) => void
+): React.ReactNode => {
+  // Convert edition tags to markdown links: $1 = edition id, $2 = edition text
+  const processedText = markdownText.replace(
+    /<edition id="([^"]+)">([^<]+)<\/edition>/g,
+    "[$2](#edition-$1)"
+  );
```
alea12, line 14: Would it make more sense to simply have the agent respond with markdown links for edition links (& page links for item page in the future)?

jackiequach: I think in the future we can update the agent to respond with markdown links; the only concern I have is that if we need to update the link routing on the frontend, a change would need to be made in the backend. This may be less of a problem once we make the orchestration updates.

etl-pipeline/tests/stochastic_processes/test_agent_responses.py
jackiequach: Agreed, removed this test.

alea12 (approved):
Thank you @jackiequach (and @bantucaravan for reviewing!)

PR #1099, opened 2026-05-12
SCHOL 553: index books fixes

bantucaravan:
## Describe your changes

This branch is based on `SCHOL-553/embedding-recomendation`, the first commit in this branch is 435caf1d49
best way to view diff for just this branch is to `git diff` against `main`

* create document vs query embedding semantics for all embedders. This makes sure that that specific document vs query embedding configurations are never accidentally left out
* fix issue where on_progress was called twice for failues in step 5 in index_books() pipeline  
* fix issue where —resume-latest was not correctly restarting on first never succeeded barcodes across multiple runs  
* add barcode filter file for copy_namespace.py (for TP namespace copies with filters)  
* fix incorrectly named QwenEmbedder to OpenAI embedder

## How to test

There are some TODO tests for embed batch


alea12 (approved):
One thing I want to run by you @bantucaravan before merging: `OpenAIEmbedder` and `BedrockEmbedder` still only implement the old interface and haven't been updated to the new methods defined in base.py. 

I understand since these embedders are outside our exploration scope right now, I don't think it's a blocker, but it will raise `NotImplementedError` if used out of the context.
 
Could we either add a comment in each of those embedder files noting the missing methods, remove these embedders entirely for now (might worth moving elesewhere?), or open a follow-up ticket to track it? Just want to make sure it doesn't get lost if we revisit those models down the road.

bantucaravan:
> One thing I want to run by you @bantucaravan before merging: `OpenAIEmbedder` and `BedrockEmbedder` still only implement the old interface and haven't been updated to the new methods defined in base.py. 
> 
> I understand since these embedders are outside our exploration scope right now, I don't think it's a blocker, but it will raise `NotImplementedError` if used out of the context.
>  
> Could we either add a comment in each of those embedder files noting the missing methods, remove these embedders entirely for now (might worth moving elesewhere?), or open a follow-up ticket to track it? Just want to make sure it doesn't get lost if we revisit those models down the road.

@alea12 I left it that way because i think of notImplemented error as self documenting. The reason I didn't implement is that the document vs query stuff is particular to the model so not relevant yet, they might even need yo be subclassed for specific models if we ever use them. I will add a code comment. Re: removal, I think the bedrock one is the stronger case for removal because ATM  because custom embedding models are not supported, but I'm a bit of a digital hoarder so I'd opt yo keep both, tho happy to do otherwise if you direct

PR #1100, opened 2026-05-12
Use GRIN barcode over edition_id for loading book metadata based on TP query

kunalnabar:
Use the barcode to probe into postgres after the turbopuffer query rather than the edition_id. This eliminates the issue with edition_ids becoming stale after FRBR reclustering.

### DIFF
--- a/etl-pipeline/api/assistant/agent.py
+++ b/etl-pipeline/api/assistant/agent.py
@@ -277,14 +277,10 @@ class ContentSearchExecutionContext:
 
     backend: TurbopufferBackend
     embedder: GoogleEmbedder
-<<<<<<< HEAD
     session_id: str
     edition_id: int
-    conversation_type: str = "contentSearch"
-=======
-    edition_id: Optional[int] = None
     barcode: Optional[str] = None
->>>>>>> b1cdce594e (Switch to barcode rather than edition_id)
+    conversation_type: str = "contentSearch"
     search_results: Dict = field(default_factory=dict)
     frbr_fields: Dict = field(default_factory=dict)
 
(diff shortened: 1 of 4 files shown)


etl-pipeline/api/assistant/agent.py
```diff
@@ -254,11 +255,17 @@ class CatalogSearchExecutionContext:
...
+    book filter for Turbopuffer searches is applied on `barcode` (stable across
+    FRBR re-clustering); otherwise it falls back to filtering on `edition_id`.
+    """
 
     backend: TurbopufferBackend
     embedder: GoogleEmbedder
-    edition_id: int
+    edition_id: Optional[int] = None
+    barcode: Optional[str] = None
     search_results: Dict = field(default_factory=dict)
```
bantucaravan: are both necessary? this will all get scrapped in favor of future state item_id/source_id and we know that barcode is indexed for all TP documents. Seems cleaner/simpler to just support barcode

bantucaravan: I don't know if there were concerns from the FE about only requesting contenSearch with barcode

kunalnabar: I guess both aren't necessary I just wanted to leave the code in case we wanted some sort of backwards compatibility

alea12:
@kunalnabar it'd be great if you could look into the conflicts so that we could merge this!

kunalnabar:
@alea12 My bad I thought this was merged! I'll get this in later today

alea12 (approved):

alea12 (approved):

PR #1103, opened 2026-05-12
Added a decorator to retry transactions on an error when committing

kunalnabar:
Some fixes on the postgres operations. 
 - Retry on deadlock correctly during commit, delete, and bulk_save
 - Actually commit when we delete


alea12 (approved):
@kunalnabar looking good, thank you for improving the tests!

PR #1116, opened 2026-05-21
SCHOL-425: Tie bar ligature render fix

nprotheroNYPL:
[SCHOL-425](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-425)

## Describe your changes

- Fixed the issue where the tie bar ligature (U+FE20 + U+FE21) was not rendering correctly in Google Chrome.
- Created `TextNormalization.ts` util file that contains functions which replaces `U+FE20` and `U+FE21` with `U+0361`.

## How to test

- Test locally

## Long explanation about fix if interested:
_from my understanding this is the issue and working fix:_

### The Problem
**Linguistic and Academic Russian:** 
In academic and linguistic contexts where the Cyrillic alphabet is not used, tie bars are required in order to understand how a sound/word is pronounced. Tie bars are identifiers that span across two letters which tells the reader to pronounce them as a single continuous sound.

**How are these typed:** 
To type a tie bar, the characters are inputted in a sequence:

`Letter 1 + Left Half [U+FE20] + Letter 2 + Right Half [U+FE21]`

Ideally a text rendering engine would understand this sequence as a single unbreakable group, or an Extended Grapheme Cluster. The engine then queries the font’s positioning tables to understand the coordinate anchors, which tells it the location of where the combining half marks should go, above the glyphs.

**Why this does not work on Chrome (but works on Firefox and Safari):**
Google Chrome’s rendering engine. 

Firefox and Safari use Gecko and CoreText respectively, engines that have cluster level fallback. This means that text is segmented by the cluster before a font is assigned. If a font cannot render the half mark, the entire cluster switches to a fallback font. With the whole cluster in the same font, the anchor coordinates are intact and the tie bar can be rendered.

Chrome’s layout engine Blink has character by character fallback splitting. If a font cannot support the half marks, the text is split mid cluster and the fallback font is only applied to the half marks and not the base letters. Because the letters and half marks are now using different fonts, the anchor coordinates are broken and the tie bar does not know where to go, causing this bug.

### The Fix
**What did not work:**
I tried multiple fixes which all did not work. 
- Using different font families, ones which include support for half marks.
- Applying css properties which force chromes engine to prioritize accuracy and ligatures.
- Edit the text right when it was returned from API to use precomposed characters.

**What worked:**
I realized that because of the way Chrome's Blink engine worked, tie bars that rely on two separated half marks would not work, or would create headaches based on what font is being used. I found that the Unicode Foundation created the Combining Double Inverted Breve (U+0361) to solve this issue. Instead of typing two half-marks that would be combined with each other by the engine, you use a single tie bar in between both of the characters. I created a util function that does this replacement and used it in locations where text including this tie mark could appear.

[SCHOL-425]: https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/src/lib/api/SearchApi.ts
```diff
@@ -69,7 +77,9 @@ export const workFetcher = async (query: WorkQuery) => {
   const url = new URL(recordUrl + "/" + query.identifier);
   url.search = new URLSearchParams(workApiQuery).toString();
   const res = await fetch(url.toString());
-  const workResult: WorkResult = await res.json();
+  const workResult: WorkResult = normalizeCombiningHalfMarksDeep(
+    await res.json()
+  );
```
alea12: We do know this issue only happens with Chrome - would it make sense to run these normalization conditionally?

web/src/util/TextNormalization.ts
```diff
@@ -0,0 +1,31 @@
+const replaceCombiningHalfMarks = (text: string | undefined): string => {
```
alea12, line 15: We already have an implementation to parse markdown, we might want to implement this on the FE unless there is a solid reason to calculate this on the API side.

jackiequach:
I agree that this does not need to be done on the API side, but since the markdown parsing only occurs for messages, it does need to applied in other instances including the item page, result cards, etc. 

I agree that it shouldn't be applied across the whole API response. I recommend narrowing this down to target fields that will be rendered instead of normalizing the full payload, this can be done on the API side or within impacted components. What are your thoughts on this @alea12?

alea12:
Thanks both for working on this. As discussed, with @nprotheroNYPL's recent commits, I'll merge this PR! Just for the record, noting following points from our discussions:

Short-term, I agree that this PR is the right immediate fix to unblock the current layout issues.

Long-term, resolving this at the the absolute upstream ingestion layer is where this fix ultimately belongs. That said, since we can't completely rule out the LLM generating these problematic ligatures/strings dynamically, we might also want to consider implementing a lightweight filtering utility or middleware on the API side down the road to scrub target fields before they hit the payload, if this becomes problematic. Not a critical priority at this point, but something to note for the future.

alea12 (approved):

PR #1120, opened 2026-05-28
SCHOL-175: Item page AQA - chat accessibility 

jackiequach:
[SCHOL-175](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-175)
[QA Doc](https://docs.google.com/document/d/1eT_J0WxwGtYVLlyMwcNcoCRB93mliZH0zq2VMKS3Kfk/edit?usp=sharing)

## Describe your changes
Addresses follow-up on AQA item 23 and unrelated landing page VQA item 26. After some discussion with Clare, we decided to try out an announcer element that appends/removes child message nodes as the screen reader announces them. There may be future work to improve this.

- Adds `ChatAnnouncer` used to mount a single persistent aria-live element to the app, so re-renders and route changes cannot unmount or interfere with it. This can be reworked in the future if we want to use a similar pattern for other things.
    - Estimates reading time from word count and clears the announcer elements after the screen reader has had time to finish to prevent stale content from being visible to screen readers.
    - Adds unit tests to test the functionality.
- Adds `markdownToPlainText`  helper method to strip markdown and edition tags from messages before announcing. 
- Replaces `scrollIntoView` + `window.scroll` with `scrollIntoView({ block: "nearest" })` to prevent unnecessary mutations to the DOM on every message update.
- Updates the user's message to be a heading.
- Updates stroke of chat input field to 2 on the landing page.

## How to test
Locally with VoiceOver

[SCHOL-175]: https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


web/src/components/ResearchAssistant/ResearchAssistantWindow.tsx
nprotheroNYPL, line 1: Not sure if this is an ongoing issue or it relates to this change, but after testing using a screen reader, I was only able to get one of my responses by enhanced search to be read aloud. Majority of the time, it would read me my inputted message, then that Enhanced Search was thinking, then when a new message appeared, it would read that the text input was empty and to "Type your response". Skipping the returned response from being read aloud completely.

Was wondering if you guys were getting this too?

nprotheroNYPL (commented):
One issue with using a screen reader with enhanced search. The VQA change looks good.

nprotheroNYPL (commented):
After testing, voiceOver works for every message except for the first returned one from enhanced search after getting navigated from another page.

If you start through the landing page and ask a question, the first result is not read aloud. Additionally when you navigate into an item it does not read the first message that it auto generates: "I can help you find relevant content in this book. Ask me a question, or try the suggestions below."

jackiequach:
> If you start through the landing page and ask a question, the first result is not read aloud. Additionally when you navigate into an item it does not read the first message that it auto generates: "I can help you find relevant content in this book. Ask me a question, or try the suggestions below."


After some research, I believe this is the intended behavior of screen readers. Users should be able to traverse the page normally with the screen reader when navigated from another page. When transitioning from landing to the research assistant page, the first announcement for the chat is the thinking input and then the response from the assistant is announced. We can confirm this behavior with Clare and create a separate ticket if there is different behavior that should be implemented.

nprotheroNYPL (commented):
When testing locally my machine keeps failing to announce the first response that arrives after switching from landing to research assistant page.

Working this through with Jackie.

nprotheroNYPL (approved):
VoiceOver on Chrome seems to occasionally skip the first response when changing from landing page to results page for me. This however seems to be an issue with VoiceOver's compatibility with Chrome and not the fix's implementation, as when tested on Safari, I do not encounter this bug. 

Jackie has tested this extensively and spoken with Clare about the issues with screen readers. The fix's implementation itself looks good to me!

PR #1124, opened 2026-06-01
SCHOL-169: Results page VQA - chat formatting

jackiequach:
### Backend

- Updates the system prompt to follow the core prompting principles from the [Gemini documentation](https://ai.google.dev/gemini-api/docs/prompting-strategies#core_prompting_principles).
    - Updates to follow a narrative prose formatting with a detailed operation sections:
      - Core mandates (scope, grounding, response protocols)
      - Step-by-step generation protocol
      - Uncertainty handling rules
      - Output formatting specifications (paragraph-only prose, no markdown lists/headers/code)
      - Translation protocol for non-English texts
      - Few-shot examples
- Adds agent behavior integration tests to validate responses are formatted in narrative prose with markdown.
    - These tests do not truly run on PRs since there is no Turbopuffer namespace assigned in the test env, we should eventually update these tests to use data from the test namespace. 

*Note*: the translation rule is a bit flaky

### Frontend

- Updates input disabled text from "Assistant is thinking..."  to "Thinking..." to address VQA item 28 in the [doc](https://docs.google.com/document/d/1JUoeghWimL8CxeAU7eJh9_-bSEg3NlEcRsoAf_5oSy4/edit?usp=sharing)

### Screenshots
Query that would typically result in headings and lists:
image

Query that includes a broad high-value topic (climate change):
image

Translation:
image

## How to test
Locally
```
TURBOPUFFER_NAMESPACE=vra-dev python -m pytest tests/integration/api/assistant/test_agent_behavior.py --env qa
```


jackiequach:
`test_on_max_turns` is failing due to the changes in handling overly broad or ambiguous user prompts. Should it be more lenient and accept the query: "find me books about climate change"?

etl-pipeline/tests/stochastic_processes/test_agent_behavior.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+async def run_catalog_query(query: str, session_id: str):
+    return await update_chat(
+        query, conversation_type="catalogSearch", session_id=session_id
+    )
+
+
+@pytest.fixture(scope="module")
+def cached_catalog_query_result():
```
kunalnabar, line 30: I like this, but it's worth noting that if any test mutates the run_result then the cached copy is permanently altered for the next usage. 
I think you could potentially guard against it by doing a `return deepcopy(cache[query])` as long as a deepcopy can be safely made of run_result.
Or document that tests can't mutate the returned object.

etl-pipeline/tests/integration/api/assistant/test_agent_behavior.py
```diff
@@ -0,0 +1,121 @@
...
+    run_result = await cached_catalog_query_result("fall of the Roman Empire")
+    citations = _EDITION_TAG_RE.findall(run_result.final_output)
+    assert len(citations) == 3, (
+        "Catalog response must include exactly 3 citations. "
+        f"Found {len(citations)} citation(s): {citations}\n"
+        f"Response:\n{run_result.final_output}"
+    )
+
+
+async def test_one_paragraph_per_book(cached_catalog_query_result):
```
kunalnabar: Basing off of 

> Every single unique quote/passage chosen MUST have its own dedicated paragraph.** Never combine multiple quotes or passages into the same paragraph, even if they come from the exact same book or page. Because you are mandated to output exactly 3 quote-and-citation pairs, your final response narrative MUST contain exactly 3 distinct, single-quote paragraphs (plus an optional introductory sentence, provided it contains no quotes or citations).

This test will allow multiple different quotes from for a single paragraph as long as they're from the same book right?

etl-pipeline/tests/integration/api/assistant/test_agent_behavior.py
```diff
@@ -0,0 +1,121 @@
...
+
+async def test_catalog_results_use_edition_markup(cached_catalog_query_result):
+    run_result = await cached_catalog_query_result("fall of the Roman Empire")
+    assert _EDITION_TAG_RE.search(run_result.final_output)
+
+
+async def test_catalog_response_has_exactly_three_citations(
+    cached_catalog_query_result,
+):
+    run_result = await cached_catalog_query_result("fall of the Roman Empire")
```
kunalnabar: Do we ever hit the case where the LLM could view this as 'too broad'?

etl-pipeline/tests/integration/api/assistant/test_agent_behavior.py
```diff
@@ -0,0 +1,121 @@
...
+    citations = _EDITION_TAG_RE.findall(run_result.final_output)
+    assert len(citations) == 3, (
```
kunalnabar: I think counting the tags could allow for 3 of the same book right? Which would be a violation of 3 unique citations?

kunalnabar (commented):
Looks good overall. What's the pass rate of the stochastic tests before / after this change? I know a lot of them are marked xfail but I'm curious if anything changes.

etl-pipeline/tests/integration/api/assistant/test_agent_behavior.py
```diff
@@ -0,0 +1,121 @@
...
+    cache = {}
+
+    async def _run(query: str):
+        if query not in cache:
+            query_hash = hashlib.md5(query.encode("utf-8")).hexdigest()
+            session_id = f"assistant-behavior-{query_hash}"
+            cache[query] = await run_catalog_query(query, session_id)
+        return cache[query]
+
+    return _run
```
bantucaravan: also check test_session_id out, I have been using https://github.com/NYPL/digital-research-books/blob/main/etl-pipeline/tests/conftest.py#L654-L675 since we are running this against the prod DB and I didn't want to pollute the DB with test instances.

etl-pipeline/tests/stochastic_processes/test_agent_behavior.py
bantucaravan, line 1: The only key thing is that I think these belong in the "stochastic_processes" folder. not the "integration". As long as we do keep that distinction, the purpose is to distinguish tests of (a) non-deterministic outputs where we (b) test example cases that do not deterministically guarantee the same behavior for all other inputs in the same "category" (e.g. "too broad"). As, long as we are keeping that distinction these should go there (also the stochastic_processes tests only run on main merge to save run time for CI

bantucaravan (approved):

etl-pipeline/tests/stochastic_processes/test_agent_behavior.py
jackiequach: Good callout, added documentation that tests can't mutate the returned object since `deepcopy` doesn't work with `update_chat`.

etl-pipeline/tests/integration/api/assistant/test_agent_behavior.py
jackiequach: Yes, this checks for edition tags per paragraph and would allow multiple quotes in 1 paragraph.

etl-pipeline/tests/integration/api/assistant/test_agent_behavior.py
jackiequach: The LLM does view this case as too broad, but provides results with clarifying questions.

etl-pipeline/tests/integration/api/assistant/test_agent_behavior.py
jackiequach: From my understanding there could be multiple citations from the same book. I updated this test to be more specific in that it will check that a quote isn't repeated.

kunalnabar (approved):

PR #1127, opened 2026-06-03
SCHOL-504: Optimize PDF loading by serving via CloudFront instead of base64

alea12:
> [!IMPORTANT]
> This PR is to be reviewed / merged *after* the release of Web Reader v5.0.3.

## Describe your changes

Refactored the item read endpoint to return only the item barcode rather than a bulky base64-encoded PDF. The Next.js BFF now uses this barcode to generate a manifest file pointing directly to our CloudFront CDN.

Previously, passing base64 PDFs between BE and FE consumed excessive memory and negatively impacted performance. This new approach will improve overall PDF load times and provides a much smoother user experience, particularly when navigating between pages.

## How to test

Locally run the API backend and Next.JS server and access item detail page (sample: http://localhost:3000/item/1e21d527-35cf-45e3-8c51-f1e60e92fe5d?featured=15297538). Ensure PDF loads.


etl-pipeline/tests/functional/processes/grin/download_test.py
```diff
@@ -46,6 +46,7 @@ def grin_status(db_manager, downloadable_barcode):
     db_manager.commit_changes()
 
 
+@pytest.mark.xfail
```
jackiequach: We should probably make this test more robust in a separate ticket since it is an integral part of our ingest process.

web/src/components/ResearchAssistant/ResearchAssistantViewer.tsx
```diff
@@ -36,6 +36,21 @@ const ResearchAssistantViewer: React.FC<{
...
   const toggleFullScreen = () => setIsFullViewport((v) => !v);
+  const getContent = React.useCallback(async (href: string) => {
+    const response = await fetch(href, {
+      mode: "cors",
+      // TODO: Once signed cookies are implemented, uncomment the following line:
+      // credentials: 'include',
+    });
+
+    if (!response.ok) {
+      throw new Error(
```
jackiequach: There is not UI feedback connected with this thrown error.

web/src/pages/api/manifest/[itemId].ts
```diff
@@ -1,35 +1,38 @@
 import type { NextApiRequest, NextApiResponse } from "next";
 import { itemsReadFetcher } from "~/src/lib/api/ResearchAssistantApi";
 
+const CLOUDFRONT_DOMAIN = "d2dqbod0obw22i.cloudfront.net";
```
Comment (jackiequach) [inline web/src/pages/api/manifest/[itemId].ts:4]: Should this be an env var instead of being defined here?

jackiequach (commented):
This works well! There is an unrelated in the web reader where the scaling buttons cause the page location to jump significantly, we should address this pre-launch or let the next wave know of this issue.

jackiequach:
Another issue I noticed that might be related to the PDF generation is that some books have text placement issues:

image

same book on QA using per page generation:
image

I will be reviewing this during my pdf gen handler implementation.

web/src/components/ResearchAssistant/ResearchAssistantViewer.tsx
alea12: Good call, since `getContent` itself was redundant until we implement signed cookies, I removed the function itself.

etl-pipeline/tests/functional/processes/grin/download_test.py
alea12: Removed both `xfail` additions to be addressed in a future PR.

alea12: https://github.com/NYPL/digital-research-books/pull/1127/changes/0bd05cda16e160c036ba69113fd7a34cc4d1c260

alea12:
Thanks @jackiequach for the comments, I've addressed them minus the env part, which I plan to fix in the new repo. Once you've approved, I'll port these diffs to the new repo and close this PR.

jackiequach (approved):
Looks good with the exception of the e2e tests failing, which I assume is related to using the CDN for the PDF.

alea12:
@jackiequach thanks, will cherry-pick these changes in the new repo!

PR #1131, opened 2026-06-03
Skip HathiTrust integration tests when API is unavailable

alea12:
## Describe your changes

Add `try/except` around the HathiTrust API call in the integration test, calling `pytest.skip()` on `requests.exceptions.RequestException`. This prevents CI failures when the external API is unreachable, consistent with the pattern used in `test_oclc_catalog_manager.py`.

## How to test

Confirm the test is reported as `SKIPPED` rather than `FAILED`


jackiequach (approved):
Just an FYI, we use rights determination for GRIN ingest, so we may want to keep an eye out for an issues with the ingest process.

PR #1134, opened 2026-06-05
SCHOL 418: filter construction errors

bantucaravan:
## Describe your changes

* add tests for filter construction errors
* update the tool descriptions/prompts to pass the tests
* use custom healthcheck script in test setup: the previous approach ran docker build automatically when needed
before running the health check on test startup

## How to test
run `uv run pytest tests/stochastic_processes/test_filter_construction.py::TestCatalogSearchFilterConstruction::test_no_filter_construction_errors --env=qa `


etl-pipeline/tests/conftest.py
```diff
@@ -84,19 +84,11 @@ def setup_env(pytestconfig, request):
...
-                "--no-recreate",
-                "--wait",
-                "--wait-timeout",
-                "10",
-            ],
+            ["bash", "scripts/docker-compose-healthcheck.sh"],
             capture_output=True,
             text=True,
             cwd=str(Path(__file__).parent.parent),  # etl-pipeline/
+            env={**os.environ, "POLL_INTERVAL": "1", "TIMEOUT": "2"},
```
alea12, line 91: @bantucaravan could you clarify the thought process behind this 2 second timeout vs default of 180 seconds?

alea12 (approved):

etl-pipeline/tests/conftest.py
bantucaravan: just fail fast. There is no need to account for the time it takes docker to start up because the services should already be on, so it should check and then fail immediately. I basically wanted it to poll once and then exit but I wanted to give a little buffer for the healthcheck script set up code to ex. Honestly it probably makes more sense to set timeout to 1 sec too

PR #1136, opened 2026-06-10
NOREF: add /chat 500 fall back error handler

bantucaravan:
## Describe your changes
add  500 fall back error handler to /chat

Note: there is an indentation change that makes the github diff a bit confusing (vscode will be better)

## How to test
added a test that makes sure runtime error in chat() is handled appropriately

### DIFF
--- a/etl-pipeline/api/blueprints/chat.py
+++ b/etl-pipeline/api/blueprints/chat.py
@@ -29,8 +29,6 @@ logger = create_log(__name__)
 
 chat_blueprint = Blueprint("chat", __name__, url_prefix="/chat")
 
-RESPONSE_TYPE = "chat"
-
 
 def prepare_search_response(search_results) -> Tuple[str, Dict] | Tuple[None, None]:
     """
@@ -160,6 +158,8 @@ def prepare_search_response(search_results) -> Tuple[str, Dict] | Tuple[None, No
 @require_session_jwt
 @timer(logger)
 def chat(session_id):
+    response_type = "chat"
+
     conversation_type = request.json.get("conversationType")
     message = request.json.get("message")
     edition_id = request.json.get("editionId")
@@ -183,18 +183,18 @@ def chat(session_id):
 
             if not message:
                 return APIUtils.formatResponseObject(
-                    400, RESPONSE_TYPE, {"message": "message is required"}
+                    400, response_type, {"message": "message is required"}
                 )
 
             if not conversation_type:
                 return APIUtils.formatResponseObject(
-                    400, RESPONSE_TYPE, {"message": "conversationType is required"}
+                    400, response_type, {"message": "conversationType is required"}
                 )
 
             if conversation_type not in ["contentSearch", "catalogSearch"]:
                 return APIUtils.formatResponseObject(
                     400,
-                    RESPONSE_TYPE,
+                    response_type,
                     {
                         "message": "conversationType must be either 'contentSearch' or 'catalogSearch'"
                     },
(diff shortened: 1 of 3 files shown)


etl-pipeline/api/blueprints/chat.py
```diff
@@ -177,84 +177,97 @@ def chat(session_id):
...
+    # even being a global error handler with logger defined in a different module. \
+    # something like https://www.structlog.org/en/stable/contextvars.html
     with LogContextVars(get_app_logger(), context=log_context):
-        return _chat_handler(
-            session_id, conversation_type, message, edition_id, barcode
-        )
+        try:
+            logger.info(f"Chat request received: {message[:20]}...")
 
+            if not message:
```
kunalnabar, line 190: Might be worth breaking the validation on these lines out into a new function so it doesn't bloat the chat() function.

etl-pipeline/tests/unit/api/blueprints/test_chat.py
```diff
@@ -109,3 +112,43 @@ def test_catalog_search_snippets_sorted(self, mocker):
...
+    def test_unexpected_error_returns_500_and_logs(self, client, mocker):
+        mocker.patch("newrelic.agent.add_custom_attribute")
+        mocker.patch(
+            "api.blueprints.chat.update_chat",
+            side_effect=RuntimeError("something went wrong"),
+        )
+        mock_logger = mocker.patch("api.blueprints.chat.logger")
+        mocker.patch.dict(
+            os.environ,
+            {"VRA_API_KEY": "test-key"},  # pragma: allowlist secret
```
kunalnabar, line 138: nice use of pragma

etl-pipeline/tests/stochastic_processes/test_filter_construction.py
```diff
@@ -332,6 +332,11 @@ async def test_author_filter_construction(self, test_session_id):
         ), f"filters do not match expected criteria: {filters}"
 
 
+# TODO: test authors with multiple spellings get multiple spellings in search/ranking_query e.g. avicenna
```
kunalnabar, line 335: Feels unrelated to this PR but all good

etl-pipeline/api/blueprints/chat.py
```diff
@@ -177,84 +177,97 @@ def chat(session_id):
...
+            response_data = {
+                "messages": messages,
+                "result_type": result_type,
+                "result": formatted_search_result,
+                "session_id": session_id,
+            }
+            return APIUtils.formatResponseObject(200, response_type, response_data)
+
+        except Exception:
+            logger.exception("Unable to execute chat")
```
kunalnabar, line 270: It may be worth having a conversation longer term about how we want to handle error handling on the client side. We may sometimes want to send more information about a specific failure back to the client than just the 500 and "Unable to execute chat" for different degradation scenarios.

kunalnabar (approved):

etl-pipeline/api/blueprints/chat.py
bantucaravan: good point. when I switch to streaming I am going to switch to fastAPI rather than flask which has pydantic model based validation, so any additional changes I will wait till after that

etl-pipeline/api/blueprints/chat.py
bantucaravan: certainly, this is a catch all, we haven't implemented any of the specific handling for specific errors like context length or token per minute limits

etl-pipeline/tests/unit/api/blueprints/test_chat.py
bantucaravan: I have no choice.... won't commit without

PR #1137, opened 2026-06-10
SCHOL-567: message ids

bantucaravan:
## Describe your changes
add messages ids from DB to assistant messages returned from /chat API endpoint

/chat response is the same but each assistant message dict gets a new `db_id` key

```json
{
   "db_id":589,
   "id":"__fake_id__",
   "content":[
      {
         "annotations":[],
         "text":"I cannot directly answer questions. Instead, I can provide you with relevant sections from the research library catalog that may help you find the answer. I will search the library for information about the narrator of Moby Dick.",
         "type":"output_text",
         "logprobs":[]
      }
   ],
   "role":"assistant",
   "status":"completed",
   "type":"message"
}
```

Changed `agent_messages.message_data` from text to jsonb data type and added an index to more performantly handle message data look ups.

Also, created a new error class to better scope handling of book not found errors in content search

#### Hanging concurrent request bug fix
A frustrating error where the localhost dev app was intermittently hanging indefinitely on different tests when concurrent requests where being made to the dev app.
Root problem: The openai-agents SDK intentionally shares a single httpx.AsyncClient at the module level across all calls. That client gets bound to the event loop of the first request that triggers it. When a second concurrent request calls asyncio.run(), it creates a new event loop — but the shared client is still attached to the first (now closed) loop, causing async operations on it to deadlock silently and hang indefinitely.
Solution: create a single persistent event loop in a background thread at module startup, replace asyncio.run() with a function that executes awaitables in the shared single loop. This ensures the shared httpx.AsyncClient is always used from the same event loop it was created in — Flask request threads submit coroutines to that loop and block waiting for the result, just as before, but without spinning up a new loop per request.

## How to test
existing tests pass
tests added for:
- chat() use of Session and source of Runner.run() input
- match session assistant messages to existing .new_items 
- test the handling of book not found errors for content search


kunalnabar (commented):
I think I don't necessarily understand the motivation of doing this in a subclass of `SQLAlchemySession` I think this is brittle because it pins us to this version of the `agents` code. `_sessions`, `_messages`, `_serialize_item`, `_ensure_tables`, and `_session_factory` are all private attributes of the parent class. Any change in the parent class in future versions of the SDK (for example handling first-write races) would either need to be mirrored. Is there a reason why we're not just reading back directly from the DB after `Runner.run` with a `select id, message_data from agent_messages where session_id = <session_id>`? That would get us the same data without coupling to a given version of the SDK and it would work for SQLite test cases. I think that it'd be a negligible cost on each turn relative to any LLM call and I believe the index would be hot after `Runner.run` executes.

bantucaravan:
> I think I don't necessarily understand the motivation of doing this in a subclass of `SQLAlchemySession` I think this is brittle because it pins us to this version of the `agents` code. `_sessions`, `_messages`, `_serialize_item`, `_ensure_tables`, and `_session_factory` are all private attributes of the parent class. Any change in the parent class in future versions of the SDK (for example handling first-write races) would either need to be mirrored. Is there a reason why we're not just reading back directly from the DB after `Runner.run` with a `select id, message_data from agent_messages where session_id = <session_id>`? That would get us the same data without coupling to a given version of the SDK and it would work for SQLite test cases. I think that it'd be a negligible cost on each turn relative to any LLM call and I believe the index would be hot after `Runner.run` executes.

@kunalnabar thanks, this very much needs conceptual review. In fact, I should have really asked for a review on an architecture plan before fully implementing. 

I agree I don't like this custom code. I didn't think about reading the data for session_id after the run... I imagine the main reason it didn't come to mind is that the session_messages could be many greater than even .inserted_items (in this agent turn) if its a long conversation. But that isn't necessarilly a problem, I have no sense of what I think the latency on that would be... I guess also to reduce the latency since we only reallly need that agent messages (not tool calls or tool call outputs), I could just do
```
SELECT
    id,
    message_data::json
FROM agent_messages
WHERE session_id = '498390b8-1b8a-43ad-b511-1a79a8f5b982'
  AND message_data::json->>'role' = 'assistant'
  AND message_data::json->>'type' = 'message';
  ```
because that would filter the returned data to much less and make pulling the entire convo maybe less latency intensive (especially not pulling the very large search results output).

How does that sound to you? What do you think about the latency vs custom code tradeoff?

**Comment by kunalnabar** [comment]
So I think there might be a way to limit the "pull whole conversation" issue.
Before `Runner.run` call and save: 
```
select coalesce(max(id), 0) as max_turn_id
from agent_messages
where session_id = <session_id>
```
After `Runner.run` call and save:
```
select id, message_data::json
from agent_messages
where 1=1
 and session_id = <session_id>
 and id > <max_id>
```
That returns all only the rows that were persisted by `Runner.run`. A thing to call out about your other two filters:
```
  AND message_data::json->>'role' = 'assistant'
  AND message_data::json->>'type' = 'message';
```
if there are many rows that are inserted by `Runner.run`, this can be costly since they'll have to cast message_data to json for each row and then do a probe/filter. I think you _should_ keep these filters, but we should change message_data from a `text` field to a `jsonb` field and create a GIN index over it.

But if you don't want to make that change, you could just do the filter on `max_id` you would have very good performance since there will be an index built over the primary key (I would guess around 1ms for each) which is very negligible compared to an LLM call that's also made during a turn.

File: etl-pipeline/api/assistant/session.py [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+        # MAYBE: just import our models from postgres/agent.py
+        self._messages = Table(
+            self._messages.name,
+            self._metadata,
+            Column("message_data", JSONB, nullable=False),
+            extend_existing=True,
+        )
+
+    async def _serialize_item(self, item: TResponseInputItem) -> dict:
+        return json.loads(json.dumps(item))
```
kunalnabar: I'm very confused by this.. I would figure this would be a no-op? What was the error you were getting before this when you didn't have a subclass and changed the `message_data` type to `jsonb`

etl-pipeline/api/assistant/agent.py
```diff
@@ -665,16 +654,88 @@ async def update_chat(
...
+    result = []
+    for item in run_result.new_items:
+        # calling .to_input_item() on ToolApprovalItem raises (agents/items.py).
+        # And they are not persisted agents/run_internal/session_persistence.py, line 243.
+        if isinstance(item, ToolApprovalItem):
+            continue
+        item_dict = item.to_input_item()
+        matched_db_id = None
+        for i, (db_id, msg_data) in enumerate(pool):
+            if item_dict == msg_data:
```
kunalnabar, line 759: I think using a regular dict equality here is a little risky. The two elements here are coming from different places with different serialization code paths. It's hard to be sure that they'll always be built the same way. If the SDK adds some sort of additional metadata at some point this could break the equality. We would silently return empty db_id for all messages and the FE wouldn't be able to correctly send thumbs up/down messages.

I think that maybe it makes sense to either (1) log the ones that fail to match or (2) count the failures and log the number at the end so if we ever start seeing results returned with empty db_id's we can investigate.

etl-pipeline/api/blueprints/chat.py
```diff
@@ -223,35 +237,36 @@ def chat(session_id):
...
             ## Build API response
 
-            # Extract new messages
-            messages = [item.to_input_item() for item in run_result.new_items]
-            logger.info(
-                f"Agent generated {len(run_result.new_items)} new message items"
-            )
+            # Note: Concurrent writes to *this* session are theoretically possible but unlikely
+            # in practice; get_new_items_with_ids() guards against them by only returning
+            # items that also appear in RunResult.new_items. MAYBE: just use session_message_items without filtering?
```
kunalnabar: I think you need the filtering. Sure concurrent writers to the same session are a little unlikely but you could have like retries based on network glitches or something and then you could process the same thing twice. So I think this filtering is actually doing some required logical work.

etl-pipeline/api/assistant/session.py
bantucaravan: thanks... yes this is a no-op, honestly inserted by the LLM that I left in, the reason for it in theory was to make sure if the `TResponseInputItem` allowed some pydantic model with non-json serializeable values this would normalize them or raise, but it is pointless because `TResponseInputItem` are all typed dicts. I'll remove this.

The pre override versions of these serialization funcs are here: https://github.com/openai/openai-agents-python/blob/main/src/agents/extensions/memory/sqlalchemy_session.py#L246-L252

The reason it is important to override the serialization funcs is that its was converting dicts into strs which is incompatible with writing to a JSONB field with asyncpg (it was throwing an error before I overrode the serialization functions)

etl-pipeline/api/assistant/agent.py
bantucaravan: I'm down to log the number that fail. I'm not super concerned about this because an "Item" is a defined type for the open AI API so there is little reason to think that the SQLAlchemySession would modify the "item" written to the DB but not modify the  item itself (Session.new_items is built from the same list that persists to DB, so it would be very weird for only the SQLAlchemySession implementation to insert some change to the data before it is persisted when the data is supposed to represent the Item types sent from the Open AI API)

kunalnabar (approved):
looks good! Good implementation of the max_id structure

kunalnabar (commented):
Will test_filter_construction#test_text_filter fail right now since it's using the update_chat API that takes in test_session_id instead of the session itself?
Also now that the functions in `TestCatalogSearchFilterUsage` don't have any async operations can we drop the async from all of them. I think right now this spins up an event loop for each test only to run nothing on it?

bantucaravan:
> Will test_filter_construction#test_text_filter fail right now since it's using the update_chat API that takes in test_session_id instead of the session itself?
> Also now that the functions in `TestCatalogSearchFilterUsage` don't have any async operations can we drop the async from all of them. I think right now this spins up an event loop for each test only to run nothing on it?

@kunalnabar good point, I'll fix the async marks, and update the filter construction tests. 

The async marks aren't causing the tests not to work because the tests are async def and sync def functions work fine inside async def functions, but the async handling is no longer necessary

PR #1141, opened 2026-06-12
SCHOL-418: filter construction errors v2

bantucaravan:
## Describe your changes

Finally fix filter construction errors (for good I think).
The core issue was in the `type` of the `filters` parameter in the function tool definition. The `type` was previously `array`. This was a problem because there was no formal definition of the full supported schema of turbopuffer filters just generic array. The type cannot be more specific because Gemini does not support full JSON schema definitions for parameter `type`s (unlike OpenAI).

The new approach lists the `filters` type as a str, but the description instructs that the str must be json encoded according to a strict json schema for the filters.

* openai function tool schema spec: https://developers.openai.com/api/docs/guides/function-calling#defining-functions
* gemeni function tool schema spec (more restrictive): https://ai.google.dev/gemini-api/docs/function-calling#function-declarations

The json schema provides strong prevention of filter construction errors.

Informative errors for effective filter construction error recovery by the LLM is provided by a pydantic model for the filters. 

## How to test
New Tests:
- The Filter pydantic model

Updated tests: filter usage and filter syntax


kunalnabar (approved):
lgtm. Do you know much latency the retry-cycling on filter-construction errors would adds to the agent time?

bantucaravan:
@kunalnabar this change should prevent errors from occurring, making error recovery rare

PR #1149, opened 2026-06-23
NOREF: centralize data seed into single script and update docs

bantucaravan:
## Describe your changes

`seed_frbr_data.py` relies on the local docker postgress instance to be running. The docs previously ensured this with a command overrode the entry-point + command of the "devsetup" docker service to use that services dependency conditions to guarantee the postgres service was running. Also the docs instructed to run one "OR" another of 2 different seeding commands, which was confusing.

I centralize data seed into single script and update docs to reflect  this. I also use a more explicit mechanism to ensure the postgres services is running (a direct `docker compose up --wait` command)

Q for reviewer: should the seed_local_data.sh shutdown all docker services after it runs?

## How to test
run seed locally

### DIFF
--- a/etl-pipeline/README.md
+++ b/etl-pipeline/README.md
@@ -60,28 +60,15 @@ This guide provides step-by-step instructions to set up local development and st
 #### (Option A) Run in docker compose
 
 
-1. Seed data in the local dockerized DB instance (one-time only):
+1. Seed data for DRB and VRA in the local dockerized Postgres DB instance (one-time only):
 
    ```bash
    # Run the database seeding process
-   docker compose -f docker-compose.setup.yml up --abort-on-container-exit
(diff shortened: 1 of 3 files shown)


jdohan:
I like consolidation for a more simplified setup.

To clarify on why there were two options that existed, one was the legacy seeding method of ingesting daily records from Hathitrust but it was no longer of value since we need to know what we're seeding (and ensure the records exist in the target vector DB namespace) in order to get hits from /chat. The newer method of seeding from the fixture file was added but the former was never completely cleaned up. This looks to achieve that.

jdohan:
Regarding whether all services should shut down after completion of seeding, I'd say no since it's typically run right before development/testing. Leaving it up is the least-surprising behavior.

alea12 (approved):

PR #1151, opened 2026-06-23
NOREF: Add chat cleared SR announcement

jackiequach:
## Describe your changes

- Adds "Chat cleared." announcement when the "Start over" button is activated

## How to test
VoiceOver locally


web/src/components/ResearchAssistant/ResearchAssistantPanel.tsx
```diff
@@ -39,6 +42,11 @@ const ResearchAssistantPanel: React.FC = () => {
     }
   }, [showChat]);
 
+  const handleStartOver = () => {
+    clearHistory(page, true);
+    announce("Chat cleared.");
+  };
```
alea12, line 55: This gets called even if the chat history is empty. Is this intended, or should we return early if no chat has been sent from patron side?

jackiequach: Good callout, updated to add a check for non-empty messages.

alea12 (approved):

PR #1154, opened 2026-06-24
SCHOL-664: Mobile/Tablet of Results Page

nprotheroNYPL:
[SCHOL-664](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-664)

## Describe your changes
- Added mobile/tablet responsiveness to results page
- Enhanced search panel now becomes a drawer at tablet breakpoint
- Added responsive directions to ArrowIcon component
- Moves 'Help and feedback' button above drawer when on tablet/mobile (Not shown in Figma but I talked to Apoorva about this).

## How to test
- Test on local machine using devtools

[SCHOL-664]: https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jackiequach:
Nice work on this implementation! I am going to split this review into 2 parts. This comment will be for visual/UX issues I found that you may have already confirmed with the designers, but I'll flag here in case you haven't.

The public domain badge in tablet and mobile view expand to full width (likely due to DS styles):
image

The hamburger menu styles do not match the designs:
image
From the designs - no border, active (and I assume hover) state bg color:
image
image

Could you confirm if the results paging text should be sticky or not on tablet/mobile view?

When clicking the start over button in tablet/mobile, I think the behavior should be to the top of the start searching panel if the patron had scrolled down in the results prior.

web/src/components/ResearchAssistant/icons/ArrowIcon.tsx
```diff
@@ -11,18 +20,34 @@ const DIRECTION_TO_ROTATION: Record<ArrowDirection, IconRotations> = {
...
+  
+  if (isLargerThanXLarge && responsiveDirection["2xl"]) {
+    resolvedDirection = responsiveDirection["2xl"];
+  } else if (isLargerThanLarge && responsiveDirection.xl) {
+    resolvedDirection = responsiveDirection.xl;
+  } else if (isLargerThanLarge && responsiveDirection.lg) {
+    resolvedDirection = responsiveDirection.lg;
+  } else if (isLargerThanMedium && responsiveDirection.md) {
+    resolvedDirection = responsiveDirection.md;
+  }
```
jackiequach, line 44: While these aren't issues in the use cases here, they can be issues in the future:
- `sm` is not being checked here, so a `sm` override in this prop would be ignored. 
- `xl` and `lg` are both gated on `isLargerThanLarge`, with `xl` being checked first, `lg` is never applied when `xl` is set, i.e `{ base: "right", lg: "down", xl: "right" }` resolves to `"right"` at any width where `isLargerThanLarge` is true. 

Also, according to the [DS breakpoints](https://github.com/NYPL/nypl-design-system/blob/development/src/theme/foundations/breakpoints.ts#L17), we should avoid using `2xl`.

web/src/components/ResearchAssistant/ResearchAssistant.tsx
```diff
@@ -14,6 +20,8 @@ import CatalogResultsSkeleton from "./CatalogResults/CatalogResultsSkeleton";
 import ResearchAssistantPanel from "./ResearchAssistantPanel";
 import ResultsBanner from "./ResultsBanner";
 
+const MIN_RESIZABLE_PANEL_HEIGHT = 512;
```
jackiequach: This can be moved into the `constants/researchAssistant` file. Also, there are instances of 512 that can be replaced with either this constant or another defined constant i.e. `DEFAULT_MOBILE_PANEL_HEIGHT`

web/src/components/ResearchAssistant/ResearchAssistant.tsx
```diff
@@ -37,9 +74,111 @@ const ResearchAssistant: React.FC = () => {
...
+      const deltaY = resizeStateRef.current.startY - event.clientY;
+      const rawHeight = resizeStateRef.current.startHeight + deltaY;
+
+      if (rawHeight < MIN_RESIZABLE_PANEL_HEIGHT - 30) {
+        resizeStateRef.current.shouldHide = true;
+      } else {
+        resizeStateRef.current.shouldHide = false;
+        if (rafIdRef.current !== null) {
+          cancelAnimationFrame(rafIdRef.current);
+        }
```
jackiequach: `cancelAnimationFrame` is only occurring in the `else` branch, but if a patron were to drag into the `shouldHide` zone, any rAF that was scheduled from a previous `pointermove` event is never cancelled. You would need to add a similar cancel in the above `if` branch.

Generally this would not be an issue, but sometimes if a pending height update isn't canceled when the drag enters the hide zone, then the panel can visibly snap to a stale height briefly before it closes.

web/src/components/ResearchAssistant/ResearchAssistant.tsx
jackiequach, line 1: I recommend moving the drag and resize behavior into a dedicated custom hook, for example `useResizablePanel(minHeight, defaultHeight, ...)`. Extracting this logic will improve readability and make the component easier to maintain.

A hook also makes this behavior much easier to test in isolation. Add tests for at least:
- Height clamping to min and max bounds
- Drag interactions updating height correctly
You can add any additional test cases you think are valuable.

web/src/components/ResearchAssistant/ResearchAssistantPanel.tsx
```diff
@@ -59,18 +71,42 @@ const ResearchAssistantPanel: React.FC = () => {
...
+            justifyContent="center"
+            width="100%"
+            position="absolute"
+            top="0"
+            left="0"
+            right="0"
+            height="18px"
+            zIndex="10000"
+            onPointerDown={onResizeStart}
+            role="separator"
```
jackiequach, line 117: According to [WCAG pattern for separators](https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/?__cf_chl_f_tk=sboOwHUSbU2EBf28dPDSbRiy.FluyAKKL1RtEivdvbo-1783029091-1.0.1.1-uHXFMFfCTVPyvEmLQNZMc9Qanu8mOYyQwimgrQxf6bg), keyboard support (at least arrow keys) would be needed for this. You can confirm with Clare about this behavior since it may be a bit different for a chat interface.

nprotheroNYPL:
Talked to Clare and she said that we should add keyboard support and follow WCAG patterns.

Apoorva and Bray said that the results paging text shouldn't be sticky to avoid cluttering the screen for mobile users. They also mentioned that clicking a link in a response bubble should hide the chat on mobile/tablet to allow users to view the result it is associated with.

web/src/components/ResearchAssistant/ResearchAssistantPanel.tsx
```diff
@@ -40,9 +61,7 @@ const ResearchAssistantPanel: React.FC = () => {
       isFirstRender.current = false;
       return;
     }
-    if (showChat) {
-      hideChatButtonRef.current?.focus();
-    } else {
```
jackiequach, line 45: What is the reasoning for removing the focus here? Curious since this would remove focus movement for keyboard/screenreader users.

web/src/hooks/useResizablePanel.tsx
```diff
@@ -0,0 +1,197 @@
...
+  );
+
+  useEffect(() => {
+    if (typeof window === "undefined") return;
+
+    const initializeHeight = () => {
+      const viewportHeight = window.innerHeight;
+      const defaultHeight = Math.round(viewportHeight * 0.8);
+      setMobilePanelHeight((previousHeight) => {
+        if (previousHeight === 640) {
```
jackiequach: Could we use a constant here to avoid a hardcoded magic number?

web/src/hooks/useResizablePanel.tsx
```diff
@@ -0,0 +1,197 @@
...
+    },
+    [clampPanelHeight, showChat]
+  );
+
+  useEffect(() => {
+    if (showChat) {
+      setMobilePanelHeight(512);
+    } else {
+      resizeStateRef.current.isResizing = false;
+      setMobilePanelHeight(512);
```
jackiequach: use `DEFAULT_MOBILE_PANEL_HEIGHT` here

web/src/components/ResearchAssistant/MessageBubble.tsx
```diff
@@ -25,6 +31,8 @@ interface MessageBubbleProps {
 
 const MessageBubble = memo(
   ({ message, index, isLoading, messageResults }: MessageBubbleProps) => {
+    const { isLargerThanMedium } = useNYPLBreakpoints();
+    const { showChat, toggleChat } = useResearchAssistant();
```
jackiequach: Since MessageBubble is memoized and the ResearchAssistantContext values are not, every context update will re-render every message bubble. 

I recommend that `showChat` and `toggleChat` should be passed down as props from the ResearchAssistantWindow that already reads context. This preserves memoization benefits.

One way to check this is by using React DevTools and enabling "Highlight updates when components render":

Image

jackiequach (commented):
Looking good! 

I also noticed a slight UI issue when on desktop view the "Show chat" button is wrapped now.

Image

web/src/components/ResearchAssistant/ResearchAssistantPanel.tsx
nprotheroNYPL: I changed it so that when a user opens the chat they are automatically focused onto the chat input box. However this could be disorienting, I can revert the change.

nprotheroNYPL:
passing in `showChat` and `toggleChat` as props from the ResearchAssistantWindow still re-rendered the components so I tried moving the check outside of the bubbles and used refs, which if I am using the devTool correctly seemed to work.

I also noticed that the implementation I used for resizing the mobile window triggered a re-render at every pixel change, so I switched to a css approach that updates only when you let go from resizing.

nprotheroNYPL:
Noticed that the style changes on the AssistantPanel made a visual bug on the item page, put a temporary fix for now but will fix this in the item page responsive PR.

web/src/components/ResearchAssistant/ResearchAssistantPanel.tsx
```diff
@@ -43,14 +64,24 @@ const ResearchAssistantPanel: React.FC = () => {
     if (showChat) {
       hideChatButtonRef.current?.focus();
     } else {
-      showChatButtonRef.current?.focus();
+      if (!showChat) {
```
jackiequach: This condition check is redundant.

web/src/hooks/useResizablePanel.tsx
```diff
@@ -0,0 +1,229 @@
...
+      setMobilePanelHeight((previousHeight) => {
+        let nextHeight = previousHeight;
+        if (previousHeight === DEFAULT_MOBILE_PANEL_HEIGHT) {
+          nextHeight = clampPanelHeight(defaultHeight);
+        } else {
+          nextHeight = clampPanelHeight(previousHeight);
+        }
+        currentHeightRef.current = nextHeight;
+        applyPanelHeightCssVar(nextHeight);
+        return nextHeight;
```
jackiequach: this can be simplified by using the `setPanelHeight` helper.

web/src/hooks/useResizablePanel.tsx
```diff
@@ -0,0 +1,229 @@
...
+        : KEYBOARD_RESIZE_STEP;
+
+      switch (event.key) {
+        case "ArrowUp":
+          event.preventDefault();
+          setMobilePanelHeight((currentHeight) => {
+            const nextHeight = clampPanelHeight(currentHeight + step);
+            currentHeightRef.current = nextHeight;
+            applyPanelHeightCssVar(nextHeight);
+            return nextHeight;
```
jackiequach: this and the ArrowDown case can use the `setPanelHeight` helper.

jackiequach (commented):
Thank you for working through all the comments and catching the mobile re-render issue. 

A few more minor comments. 

The active border state for the hamburger menu button does not have enough contrast with the background. You can change it to the same white outline as the other buttons on the header bar.

Image

Also, I noticed there is a scrollbar for the chat in mobile and was curious if we should expose the scrollbar for desktop view. This can be a separate ticket if so.

jackiequach (approved):
Looks good! Resolve the merge conflict and ensure the CI passes.

PR #1155, opened 2026-06-24
SCHOL-651 - switch /chat to gemini-embedding-2

bantucaravan:
## Describe your changes
Switch /chat endpoint to use gemini-embedding-2 for queries and indexed documents.

The change converts to the new "get_index_config" paradigm where a configuration exists for each index that specified an embedding model and index schema used both for index document embedding creation and query embedding creation.

Tests were also updated to accomodate the refactor

No default index is hardcoded into agent.py, the index used by /chat is entirely determined by the env file.

## How to test
all existing tests should pass.


alea12 (approved):
@bantucaravan looking good!

PR #1158, opened 2026-06-26
SCHOL-684: Add retries to known flaky tests

jdohan:
## Changes

Add retry once then skip logic tests in the "services/source/" folder, triggered by a limited set of errors deriving from the source data being unavailable.

Adds the `pytest-retry` package to backend dev dependencies to implement retries on some known flaky tests in `tests/integration/services/sources`.

1 retry is carried out when at least a network connection issue is encountered (the list of exceptions is found in the new conftest.py file) after a 2 min delay for simplicity and realtime feedback is printed to the terminal to give info on retry attempts, such as the test being retried and the time remaining until it's run again.

While `pytest-rerunfailures` already exists and is more tried-and-true than `pytest-retry`, the latter is more idempotent and works better with concurrent test runs since the retry logic runs before the result is reported back to the xdist controller. Since both libraries recognize `pytest.mark.flaky` and automatically rerun/retry, `pytest-rerunfailures` was removed to avoid the two colliding.

## How to test
To test locally, temporarily update a test in `tests/integration/services/sources` to throw an exception then run the tests using `make integration-drb`. For example, update `test_gutenberg_service.py` to:
```
import requests

from services.sources.gutenberg_service import GutenbergService


def test_get_records(monkeypatch):
    # TEMP: force timeout on every call to verify pytest-retry behavior
    monkeypatch.setattr(requests, "post", lambda *a, **kw: (_ for _ in ()).throw(requests.exceptions.Timeout()))

    gutenberg_service = GutenbergService()

    records = gutenberg_service.get_records(limit=100)

    for record in records:
        assert record is not None

```
The known flaky tests often pass after a rerun in CI but if the service is down for longer than a couple of minutes, the retry will fail. Can adjust going forward if desired.

### DIFF
--- a/etl-pipeline/tests/integration/services/sources/test_hathi_trust_service.py
+++ b/etl-pipeline/tests/integration/services/sources/test_hathi_trust_service.py
@@ -1,19 +1,13 @@
 from datetime import datetime, timedelta
 
-import pytest
-import requests
 from services import HathiTrustService
 
 
 def test_get_records():
     hathi_trust_service = HathiTrustService()
+    records = hathi_trust_service.get_records(
+        limit=5, start_timestamp=datetime.now() - timedelta(days=7)
+    )
 
-    try:
-        records = hathi_trust_service.get_records(
-            limit=5, start_timestamp=datetime.now() - timedelta(days=7)
-        )
-
-        for record in records:
-            assert record is not None
-    except requests.exceptions.RequestException as e:
-        pytest.skip(f"HathiTrust API unavailable: {e}")
+    for record in records:
+        assert record is not None
--- a/.github/workflows/etl-pipeline-ci.yaml
+++ b/.github/workflows/etl-pipeline-ci.yaml
@@ -75,17 +75,17 @@ jobs:
       - name: Run API integration tests against QA env
         working-directory: ./etl-pipeline
         run: |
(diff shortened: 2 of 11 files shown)


etl-pipeline/tests/integration/services/sources/test_hathi_trust_service.py
```diff
@@ -1,19 +1,13 @@
...
+    )
 
-    try:
-        records = hathi_trust_service.get_records(
-            limit=5, start_timestamp=datetime.now() - timedelta(days=7)
-        )
-
-        for record in records:
-            assert record is not None
-    except requests.exceptions.RequestException as e:
```
bantucaravan, line 18: @jdohan Shouldn't we add `requests.exceptions.RequestException` to RETRY_ON_EXCEPTION list?

bantucaravan (changes requested):

etl-pipeline/tests/integration/services/sources/test_hathi_trust_service.py
jdohan: That one can be left out IMO since it's a catch-all that could potentially conceal legitimate bugs. I think a selection of its subclasses like what's being done currently achieves what we're going for here re: retrying and skipping known issues.

bantucaravan (approved):

bantucaravan (approved):

PR #1159, opened 2026-06-26
SCHOL-470: why am i seeing this

bantucaravan:
## Describe your changes
* add /result-reason endpoint that implements the "why am I seeing this" UI content. see documentation: https://newyorkpubliclibrary.atlassian.net/wiki/spaces/DIGTL/pages/1766719489/result-reason+endpoint
* fixed bug where map_editions_and_records() recently started using `grin_statuses` table but the local env seed data did not include that data leading localhost to always return empty search results
* clarified that /chat response API for the cases: search executed with results, search executed with zero results, and no search executed
* refactored the writing and parsing of xml tag delimited search result (chunk text) data, passed by the search tools to the LLM to read

## How to test
Everything is tested
- llm behavioral tests on the result reason content
- API tests
- unit tests


etl-pipeline/api/assistant/agent.py
```diff
@@ -70,6 +73,10 @@
 
 logger = create_log(__name__)
 
+DEFAULT_LLM = "gemini-3.5-flash"
+
+TOOL_ERROR_PREFIX = "An error occurred while running the tool"
+
```
alea12, line 79: Possibly future work, we might want to refactor these constants to a separate file for code visibility. We can do that as part of https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-605

etl-pipeline/api/assistant/agent.py
```diff
@@ -1162,22 +1183,109 @@ def format_frbr_fields(orm_work, orm_edition):
...
+    tag matching one of _KNOWN_ELEMENT_NAMES (e.g. "</chunk>"), the whole
+    value is XML-escaped so it can't be mistaken for real structural markup.
+    Otherwise value is returned unchanged (as a str).
+    """
+    value = str(value)
+    collides = any(
+        f"<{name}>" in value or f"</{name}>" in value
+        for name in _SEARCH_RESULT_SCHEMA_ELEMENT_NAMES
+    )
+    return (xml_escape(value), True) if collides else (value, False)
```
alea12, line 1252: Could we just escape every string that gets injected to the XML to play safe (and make the code simpler)? Or is it more troublesome in the downstream to unescape these characters before presenting them to the UI / language models?

etl-pipeline/api/assistant/agent.py
```diff
@@ -1162,22 +1183,109 @@ def format_frbr_fields(orm_work, orm_edition):
...
+
+_SEARCH_RESULT_SCHEMA_ELEMENT_NAMES = _SEARCH_RESULTS_XSD_TREE.xpath(
+    "//xs:element/@name", namespaces={"xs": "http://www.w3.org/2001/XMLSchema"}
+)
+
+
+def _escape_if_collides_with_known_tags(value: Any) -> Tuple[str, bool]:
+    """
+    Return (value, was_escaped). If value contains a literal open or close
+    tag matching one of _KNOWN_ELEMENT_NAMES (e.g. "</chunk>"), the whole
```
alea12: minor: I think this comment is referring to `_SEARCH_RESULT_SCHEMA_ELEMENT_NAMES`, not `_KNOWN_ELEMENT_NAMES`

etl-pipeline/api/blueprints/chat.py
```diff
@@ -32,141 +33,224 @@
...
+chat_blueprint = Blueprint("chat", __name__)
 
+# --- types for prepare_search_response() ---
 
-def prepare_search_response(search_results) -> Tuple[str, Dict] | Tuple[None, None]:
+
+class SearchResponse(NamedTuple):
+    result_type: str
+    formatted_search_result: Dict
+    tool_call_id: str
```
alea12, line 49: Also could be another ticket: these type definitions can have its own file, something similar to `etl-pipeline/api/assistant/types.py`

etl-pipeline/api/blueprints/result_reason.py
```diff
@@ -0,0 +1,258 @@
...
+from ..assistant.snippets import format_conversation_history
+
+
+logger = create_log(__name__)
+
+result_reason_blueprint = Blueprint("result_reason", __name__)
+
+FALLBACK_RESULT_REASON = (
+    "This result appears in your search based on its relevance to your query. "
+    "The catalog's search algorithm identified this item as a potential match for your research. "
```
alea12: Related to Lara's feedback, we could just use what we already have as a fallback hardcoded in FE (web/src/components/ResultCard/ResultCard.tsx):

> You're seeing this result because this book covers topics relevant to your request. The following sections You're seeing this result because this book covers topics relevant to your request. The following sections were identified as matching your query.

IMO we might generally want to avoid technical terms such as "search algorithm" and "search engine", but this can be discussed with product / design.

etl-pipeline/api/blueprints/result_reason.py
```diff
@@ -0,0 +1,258 @@
...
+    "It may share themes, subjects, or content related to your inquiry."
+)
+
+RESULT_REASON_SYSTEM_PROMPT_TEMPLATE = """\
```
alea12: Another codebase question: when do we use `etl-pipeline/api/assistant/prompts` vs inline prompt template definition? We can create something like `etl-pipeline/api/assistant/prompts/result_reason` and move these there.

alea12 (commented):
Review 50% done, leaving first-pass comments, but will come back to this PR later!

etl-pipeline/api/blueprints/result_reason.py
```diff
@@ -0,0 +1,258 @@
...
+                    },
+                )
+
+            # Truncate conversation history to end at (and including) this tool call's
+            # arguments, excluding the tool call output.
+            messages = messages[:function_call_idx]
+
+            explanation, is_ai_generated = get_result_reason(messages, edition_result)
+            logger.info(
+                f"Result reason generated. Hardcoded fallback used?: {is_ai_generated}"
```
alea12: Log message might be inverted?

etl-pipeline/api/blueprints/result_reason.py
```diff
@@ -0,0 +1,258 @@
...
+RESULT_REASON_SYSTEM_PROMPT_TEMPLATE = """\
+You are a research assistant at a library helping users understand why specific \
+search results appear for their queries. Given the conversation history and the \
+search query that was executed, explain in 3-4 sentences (~450 characters) why the \
+specified book appears as a result. Be specific about what connects the book's \
+subject matter, themes, or content to the user's research interest. Write clearly \
+for a general audience.\
+
+Conversation History:
+{conversation_history}
```
alea12: Question: does this include tool calls? if so, would it make sense to filter with `type == "message"`? (could save a lot of input tokens)

etl-pipeline/api/assistant/agent.py
bantucaravan: I wanted to have the chunk text unmodified as possible for being shown to the LLM (also the form in which it is shown to the LLM is the form in which it is saved in the DB convo history). I did this with use cases like pulling exact quotations for snippets or cases where the chunk text itself was XML/HTML (like a table) and we want it to be clearly interpretable by the LLM. 

This particular point in the code has 2 complications. one is the the escaping only known elements, which is actually not that complicated. The second is the tracking of whether anything has been escaped in order to conditionally display the note at the end of the search result here https://github.com/NYPL/digital-research-books/pull/1159/changes/BASE..8169c2fb56c2c816cbf29af6eecaeb69277efe67#diff-33d474a72ebdb22b1d8df3e915a2310149b77ccf83450e0da0e6bba36cca3d32R1419-R1424 

The tracking of whether anything has been escaped is why emit has to be defined in the scope of format_book() and the nonlocal is used among other things. so simply always adding the note would reduce a bunch of the complication.

The question of whether the element text of the XML formatted search result should be universally xml escaped is less about code complication in my mind, but about the downstream uses, and whether we are comfortable showing the LLM xml escaped text for our downstream uses. I leaned toward not treating the search result as strict pure XML and more towards using xml tags are informal delimiters, but I can see other arguments. I don't think the code complication itself is a strong enough reason not to keep it as "psuedo xml"

etl-pipeline/api/blueprints/result_reason.py
bantucaravan: yes, I actually meant to replace that with the hardcoded copy from the front end, but forgot. this was a placeholder fallback.

etl-pipeline/api/blueprints/result_reason.py
bantucaravan: solid point, I was not being consistent about this

etl-pipeline/api/blueprints/result_reason.py
bantucaravan: no it does not. speaking of overcomplicated. the formatting function is here: https://github.com/NYPL/digital-research-books/pull/1159/changes/BASE..8169c2fb56c2c816cbf29af6eecaeb69277efe67#diff-5c362d6fb492edb9068fe2ce02334194065412d2d07afd1f450fc8cf74da499bR379-R393

(most of the complication is is deleting all but the last of consecutive tool calls)

alea12:
@bantucaravan thanks for additional pushes, I think we're getting close! Two things I'd like to flag:

## Barcode <> Editiion ID mismatch

Right now, the following curl request fails with `barcode is required` which isn't compatible with the doc.

```sh
curl -X POST http://127.0.0.1:5050/result-reason \
  -H "Content-Type: application/json" \
  -H "X-API-KEY: XXX" \
  -b "vra_session=XXX" \ 
  -d '{
    "call_id": "cwgt2gfg",
    "edition_id": 15297538
  }'
```

This succeeds:

```sh
{
    "call_id": "cwgt2gfg",
    "barcode": 33433062509165
}
```

## `/chat` response shape

This PR changes the `/chat` response from top level `result_type` / `result` to `search_result: {result_type, results, tool_call_id}` (which I agree is the right direction), but current FE still expects `chatData.result` / `chatData.result_type`. Right now this breaks the FE and search results are not shown on the item list.

Could we either:

* Keep `result_type` / `result` at the top level and just add `tool_call_id` (or `last_tool_call_id`). This doesn't break the FE, still satisfies input needed for `/result-reason`, and avoids merge-order coordination
* And/Or split the reshape into its own PR and coordinate merge order with the FE PR

bantucaravan:
> @bantucaravan thanks for additional pushes, I think we're getting close! Two things I'd like to flag:
> 
> ## Barcode <> Editiion ID mismatch
> Right now, the following curl request fails with `barcode is required` which isn't compatible with the doc.
> 
> ```shell
> curl -X POST http://127.0.0.1:5050/result-reason \
>   -H "Content-Type: application/json" \
>   -H "X-API-KEY: XXX" \
>   -b "vra_session=XXX" \ 
>   -d '{
>     "call_id": "cwgt2gfg",
>     "edition_id": 15297538
>   }'
> ```
> 
> This succeeds:
> 
> ```shell
> {
>     "call_id": "cwgt2gfg",
>     "barcode": 33433062509165
> }
> ```
> 
> ## `/chat` response shape
> This PR changes the `/chat` response from top level `result_type` / `result` to `search_result: {result_type, results, tool_call_id}` (which I agree is the right direction), but current FE still expects `chatData.result` / `chatData.result_type`. Right now this breaks the FE and search results are not shown on the item list.
> 
> Could we either:
> 
> * Keep `result_type` / `result` at the top level and just add `tool_call_id` (or `last_tool_call_id`). This doesn't break the FE, still satisfies input needed for `/result-reason`, and avoids merge-order coordination
> * And/Or split the reshape into its own PR and coordinate merge order with the FE PR

@alea12 

1) Yes, I quickly changed the API from accepting barcode to accepting edition_id before the demo yesterday and did not update the tests. its not the curl that fails (I was running edition_id calls in the demo), the "barcode is required" is a stale test that was testing the input payload keys. I'll  update the tests for the API change

2) I made the changes to the /chat API in consultation with Jackie so she is aware. Regarding how to merge, I would actually propose, we close this PR, and the FE starts a branch based on this branch with a new PR since the FE/BE must be aligned in order for E2E tests to pass however we make that change.

alea12 (approved):

alea12:
As per @bantucaravan's suggestion, we'll close this PR for now and wait for FE implementation by @nprotheroNYPL on the same branch!

PR #1160, opened 2026-06-29
SCHOL-732: Map chunk hits to items by barcode instead of book_id

alea12:
## Describe your changes
Enable using barcodes as the key for mapping chunk hits to item IDs. 

## How to test
Updated fixtures - tests should pass.


etl-pipeline/api/assistant/agent.py
```diff
@@ -381,6 +381,10 @@ def map_editions_and_records(record_ids=None, edition_ids=None):
         ids = edition_ids
         source_col = "edition_id"
         target_col = "record_id"
+    elif barcode_ids:
```
bantucaravan, line 388: we should add a similar assertion as for edition_ids and update the edition_id assertion for clarity. The function should only get one of the 3 options non-None

etl-pipeline/api/assistant/agent.py
```diff
@@ -720,28 +739,26 @@ def results_to_chunk_hits(results: list[ScoredHit]) -> Iterator[dict[str, Any]]:
     # NOTE: future: the item_id will be directly indexed in the chunk hit \
     # making this function unnecessary, ScoredHit can be used instead.
 
-    missing_item_ids = []
+    missing_barcodes = []
     try:
-        # map record_id->item_id to add to chunk metadata
-        # book_id = record_id
-        record_ids = set(cd.book_id for cd, _ in results)
```
bantucaravan, line 727: The root problem is identified by asking "Why was book_id returning barcodes not record_ids"? I believe the root cause was the switch from the gemini-embedding-001 index to the gemini-embedding-2 index, the newer index was built using a new indexing pipeline. In the original pipeline book_id was set to record_id and in the new one it was set to barcode. 

the relavant source code at the root cause is here: https://github.com/NYPL/digital-research-books/blob/SCHOL-732%2Ffix-barcode-mapping/etl-pipeline/vector_indexing/core/types.py#L32-L35

bantucaravan: as we move to the new schema. we should remove book_id from the ChunkDocument, it was a ill-defined term from an earlier period in our development (I added it a couple of weeks after I joined). I checked it is not used any where substantial, after this refactor.

I suggest removing it from the chunkDocument (and thus repo) in the PR if you have time for it, I could also push a commit for that.

bantucaravan: @kunalnabar make sure to kill book_id from ChunkDocument in the ETL for the new schema not needed for backwards compatibility, was originally supposed to be record_id, was poorly named by me just when I started.

bantucaravan (approved):

etl-pipeline/api/assistant/agent.py
alea12: thanks that's a good call. [added a slightly updated version of params validation](https://github.com/NYPL/digital-research-books/pull/1160/changes/46314e4ca1a31f6e295ea9f0646c1b066dfcf96b).

PR #1167, opened 2026-07-01
SCHOL-741: content search tool use bug

bantucaravan:
#### Problem
`search_book` (contentSearch's tool) was failing on every invocation.

#### Cause
commit a03a8ad2bd7 ("SCHOL 553: retrieval evals #1107"), introduced a bug where ctx.context.search_results[ctx.tool_call_id]["edition_data"] is accesses before it is written in search_book() the search tool used in content/item search. Also the commit added a modification of the same key in the except block that raises the key error in the except block, masking other errors that appear in the function.

#### Changes
- Reordered `search_book` to build the result locally, format it, *then* store it in `search_results[call_id]` — no more read-before-write.
- Changed `del ctx.context.search_results[ctx.tool_call_id]` to `ctx.context.search_results.pop(ctx.tool_call_id, None)` in both `search_book` and `search_catalog`, so a failed cleanup can no longer mask the original exception.

## How to test
added two unit tests that call `search_book.on_invoke_tool()` / `search_catalog.on_invoke_tool()` **directly** while mocking external data dependencies. This exercises the tool's real internal logic on every run, so any future bugs on the internal code will fail immediately and deterministically.


etl-pipeline/tests/unit/api/assistant/agent/test_agent.py
```diff
@@ -56,14 +68,93 @@ def test_update_chat_catalog_search(self, mocker):
...
+
+    @pytest.mark.asyncio
+    async def test_search_catalog_on_invoke_tool_returns_results(
+        self, mocker, mock_search_backend, chunk_doc
+    ):
+        mock_search_backend([chunk_doc])
+
+        context = CatalogSearchExecutionContext(
+            backend=mocker.MagicMock(),
+            embedder=mocker.MagicMock(),
```
alea12, line 132: backend & embedder here bypass `get_index_config` mock in `mock_search_backend`, is this an intentional design?

alea12 (approved):
Just left one non-blocking question. Looking good!

etl-pipeline/tests/unit/api/assistant/agent/test_agent.py
bantucaravan: @alea12  the `get_index_config()` mock is for using `mock_search_backend` when testing the full `update_chat()` or /chat API. `get_index_config` isn't actually called in the search tools, so this mock doesn't really bypass `get_index_config`, its just not part of the tested code.

I think the motivation for instantiating the backend and embedder outside of the tool function scope was to avoid instantiating them on each tool invocation incase there were multiple per agent turn. That said, looking at the __init__s the are very lightweight and it might be neater to just pass the index name to the search tool and instantiate the backend and embedder on each search tool invocation. Let me know if you have a strong feeling on this.

PR #1168, opened 2026-07-02
SCHOL-706: new test index

bantucaravan:
## Describe your changes
replace gemini-embedding-001 test index ("vra-test") with gemini-embedding-2 test index ("vra-test-gemini_embedding_2"). Same barcodes, same chunks, different vectors.

used copy_namespace.py to do the copy filtering on the barcodes in "vra-test"

Added config for new test index

Added env support for tp-scan.py

## How to test
N/A


etl-pipeline/vector_indexing/scripts/copy_namespace.py
```diff
@@ -144,7 +144,7 @@ def copy_namespace(
     if dry_run:
         total_scanned = 0
         total_bytes = 0
-        for chunk in src_backend.scan(filters=filters):
+        for chunk, _ in src_backend.scan(filters=filters):
```
alea12, line 147: good catches!

alea12 (approved):

PR #1169, opened 2026-07-02
SCHOL-676: Add DataLayer for Landing Page Query

nprotheroNYPL:
[SCHOL-676](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-676)

## Describe your changes
- Added a temp. flag in session `queryFromLandingPage` to keep record that query came from landing page.
- ResearchAssistant now pushes to dataLayer when new results come in (only from landing page for now).

## How to test
- Test on local machine using dev tools
- Once this review is passed, Data team will test if data is streaming to GA4.

[SCHOL-676]: https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/web/src/components/ResearchAssistant/ResearchAssistant.tsx
+++ b/web/src/components/ResearchAssistant/ResearchAssistant.tsx
@@ -1,11 +1,12 @@
 import { Box, Flex } from "@nypl/design-system-react-components";
-import React, { useEffect, useMemo } from "react";
+import React, { useEffect, useMemo, useRef } from "react";
 import {
   HEADER_HEIGHT,
   MARGIN_BLEED,
   PADDING_COUNTER,
 } from "~/src/constants/researchAssistant";
 import { useResearchAssistant } from "~/src/context/ResearchAssistantContext";
+import { trackEvent } from "~/src/lib/gtag/Analytics";
 import { isCatalogResults } from "~/src/util/ResearchAssistantUtils";
 import BackToResultsButton from "../BackToResultsButton/BackToResultsButton";
 import EmptySearchPrompt from "../EmptySearchPrompt/EmptySearchPrompt";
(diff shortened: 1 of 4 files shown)


web/src/components/ResearchAssistant/ResearchAssistant.tsx
```diff
@@ -58,6 +58,23 @@ const ResearchAssistant: React.FC = () => {
     return null;
   }, [messages.length, results]);
 
+  useEffect(() => {
+    const queryFromLandingPage = sessionStorage.getItem("queryFromLandingPage");
```
jackiequach: Instead of creating a new `sessionStorage` key for this flow, we can track whether a landing query is sent with a local ref in this component and reuse the existing initial message effect.

web/src/components/ResearchAssistant/ResearchAssistant.tsx
```diff
@@ -58,6 +58,23 @@ const ResearchAssistant: React.FC = () => {
...
+      let resultsCount = 0;
+      if (isCatalogResults(latestResults)) {
+        resultsCount = latestResults.paging?.totalRecords || 0;
+      }
+      window.dataLayer = window.dataLayer || [];
+      window.dataLayer.push({
+        event: "view_query_results",
+        query_type: "landing_page",
+        results_count: resultsCount,
+      });
```
jackiequach: `trackEvent` can be used for this pattern. You'll need to add a new type for this event.  Add a test for the new event shape in Analytics.test.ts

jackiequach (commented):
Could you add a unit test for the new event data type?

jackiequach (approved):
looks good

PR #1170, opened 2026-07-02
SCHOL-748: run stochastic processes tests on PR push

bantucaravan:
## Describe your changes
see  https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-748

Changed the PR push triggered GHA workflow to include the stochastic processes tests. This change included a refactor to run each test type as a concurrent job (separate  GHA runners).

The stochastic processes tests use the qa env services (sql DB, vector DB) in order to get vectors and metadata for the full 10k books. It would probaly be more natural to locate the stochastic process tests in the same workflow file as other tests that use QA env for DRY purposes and just change the triggers to conditionally include PR pushes for the stochastic processes tests, but for this PR I have kept the PR push triggered tests in a single file

In the future, we can refactor stochastic processes test to run against the local/dev env by either increasing the available data in that env or tweaking the tests to accommodate the more constrained data availability currently in the local env.

Q for reviewers: I can scope each test time to changes in particular paths/files in the repo using https://github.com/dorny/paths-filter. Any recommended? I left all tests running for all paths for now.

## How to test
Do the tests pass!?

### DIFF
--- a/.github/actions/aws-auth-for-gha-runner/action.yaml
+++ b/.github/actions/aws-auth-for-gha-runner/action.yaml
@@ -1,0 +1,17 @@
+name: 'AWS Auth for GHA-Hosted Runner'
+description: 'Authenticate to AWS from a GitHub-hosted runner via OIDC by assuming the GithubActionsDeployerRole'
+
+inputs:
+  aws-region:
+    description: 'AWS region'
+    required: false
+    default: 'us-east-1'
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        role-to-assume: arn:aws:iam::946183545209:role/GithubActionsDeployerRole
+        aws-region: ${{ inputs.aws-region }}
--- a/.github/actions/aws-auth-for-gha-runner/action.yaml
+++ b/.github/actions/aws-auth-for-gha-runner/action.yaml
@@ -1,6 +1,10 @@
 name: 'AWS Auth for GHA-Hosted Runner'
 description: 'Authenticate to AWS from a GitHub-hosted runner via OIDC by assuming the GithubActionsDeployerRole'
 
+# NOTE: the calling workflow/job must declare `permissions: id-token: write` — composite
+# actions cannot grant this themselves. See:
+# https://docs.github.com/en/actions/how-tos/secure-your-work/security-harden-deployments/oidc-in-aws#adding-permissions-settings
+
 inputs:
   aws-region:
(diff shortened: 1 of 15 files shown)


jdohan:
This looks great. I really like the parallel execution on separate legs and isolation of the AWS auth. Running certain tests by identifying paths ain't a bad idea either, especially since the stochastic processes tests run against the QA data (we can filter database changes with it perhaps).

The latest commit (only adding a YAML comment) resulting in a failed check whereas the prior one passed indicates there's still some flakiness to iron out that could become troublesome if merged to main unaddressed. Note the 3 xpassed tests as well. Could we loosen up the LLM-as-judge's decision making?

alea12:
One thing to note: making self-hosted runner reachable from pull_request trigger could be somewhat risky since it allows any external contributor's PR to run on our AWS infrastructure. Is there a way to gate this?

bantucaravan:
> One thing to note: making self-hosted runner reachable from pull_request trigger could be somewhat risky since it allows any external contributor's PR to run on our AWS infrastructure. Is there a way to gate this?

@alea12 this is a very good point. The main problem is external/untrusted actors executing code simply by creating a PR.  Overall, since the point of the self-hosted runner is to provide authentication into NYPL AWS accounts, it is important that we can trigger things that test our actual infra on PRs (even if just for data access as in this case). The new enhanced-search repo is "internal" vs this repo being "public", so the simplest solution IMO is just to port this change over to the new repo and apply it there where only internal (thus trusted we hope) actors can trigger a PR.  Additionally we can restruct PR creation to collaborators, which if it isn't done already should be done on both repos (see: https://github.blog/changelog/2026-02-13-new-repository-settings-for-configuring-pull-request-access/#restrict-pull-requests-to-collaborators).  Additionally this repo should have the setting for GHA workflows from any external contributor to require approval (if this is not already done), see instructions here: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories.

My concluding thought is that if both these settings are configured, it is safe to add this self-hosted repo access on this repo.

ONce this is done these settings should be noted in our confluence documentation

alea12:
@bantucaravan 

> simplest solution IMO is just to port this change over to the new repo and apply it there 

I like this approach! I'll try to cherry-pick your work from this repo to the new one (so that the git authorship gets transferred). Thanks for the notes re: repo settings as well. I confirmed both repos limit PRs to collaborators. I didn't have the access to GHA workflows setup, which I'll check back later.

alea12 (approved):

PR #1172, opened 2026-07-13
NOREF: docker healthcheck fix

bantucaravan:
## Describe your changes

**BUG**: previous "docker-compose-healthcheck.sh" did not guaranteed all services listed in docker-compose.yaml were healthy (if healthcheck available). Instead it checked if running services were healthy. 

**FIX**:
(a) I updated "docker-compose-healthcheck.sh" to check all services listed in the docker-compose.yaml of the CWD and ensure that those specific services are healthy (if healthcheck is available)
(b) because the changes increased the complexity of "docker-compose-healthcheck.sh", and @alea12  had mentioned in prevoious discussion of this issue that he preferred less complexity. I opted to refactor conftext.py to check a hard coded list of required docker containers rather than use "docker-compose-healthcheck.sh"  to solve the more generic problem of identifying arbitrary services from the docker-compose.yaml file then checking them. I left "docker-compose-healthcheck.sh" around so it can be used if desired, but it is no longer used any where and can be deleted if desired.

## How to test
starting  a local env test run on a non unit test should exit if the required docker services are not running


alea12 (approved):
Let's remove `docker-compose-healthcheck.sh` if it's not used anymore, we could always trace back git history if that becomes necessary. I'd generally prefer removing any unreachable code in these cases.

Otherwise looking solid (+thanks for locking the `openai` version), good to go!

